PR #23, opened 2015-04-02
Adds layout/chrome and footers (SUL & global) plus fixes for rubocop and scss-lint

### LINKED ISSUE #5: Add site layout/chrome
screen shot 2015-04-01 at 2 08 34 pm

### LINKED ISSUE #7: Add global footer & SUL footer
Similar to SearchWorks

screen shot 2015-04-01 at 2 09 40 pm

jchristo4:
Closes #5 
Closes #7 

image


app/controllers/home_controller.rb
```diff
@@ -1,6 +1,8 @@
 # Controller to handle root HTTP
 # connections and display the home page
 class HomeController < ApplicationController
+  layout 'sul-requests'
```
jkeck: Should we add this to the `ApplicationController` instead so it applies globally to all of our controllers?

jkeck: (alternatively, should we just make this the `application.html.erb` layout file and let rails make it the default)

app/views/shared/_top_navbar.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <header id="topnav" class="header-logo" role="banner">
+    <%= link_to 'http://library.stanford.edu' do %>
+      <%= content_tag :span, "Stanford University Libraries", class: "sr-only" %>
+      <%= image_tag "sul-logo.png", class: "hidden-xs", alt: "", height: 25 %>
+      <%= image_tag "sul-logo-small.png", class: "visible-xs", alt: "", height: 25 %>
+    <% end %>
+
+    <div class="header-links">
+      <%= link_to "My Account", "http://library.stanford.edu/myaccount", class: 'first' %>
+      <%= link_to "Feedback", "#"  %>
```
jkeck: @jvine should we just link this to http://searchworks.stanford.edu/feedback for now (or permanently)?

app/controllers/home_controller.rb
jchristo4: @jkeck moved `sul-requests` contents to default `application` layout

app/views/shared/_top_navbar.html.erb
jvine: @jkeck Yeah, that's fine for now, I'll think about permanently.

PR #41, opened 2015-04-10
Add authorization controller to handle webauth login and logout

### LINKED ISSUE #40: Add controller to handle login/logout
With appropriate redirects and flash messages.

Login URL: `/webauth/login`
Logout URL: `/webauth/logout`

jkeck:
Closes #40

(WIP while I try to figure out some LDAP issues)


config/deploy.rb
```diff
@@ -37,6 +43,14 @@
 # set :keep_releases, 5
 
 namespace :deploy do
+  desc 'Restart application'
+  task :restart do
+    on roles(:app), in: :sequence, wait: 5 do
+      execute :touch, release_path.join('tmp/restart.txt')
```
cbeer: Have you seen https://github.com/capistrano/passenger? Although I see it's doing the passenger 5-style restarts (maybe we should figure out how to do those in our environment..)

jkeck: Yeah, I did see that.  The Passenger-5 style restarts was what kept me away from it (although that was also before we had passenger 5 running on the server).

cbeer: :ok_hand: Maybe when we rebuild exhibits, I'll get a chance to try out passenger 5 and see how it goes.

jkeck:
@jchristo4 I think this is ready for :eyes:

app/controllers/application_controller.rb
```diff
@@ -6,13 +6,32 @@ class ApplicationController < ActionController::Base
...
 
   private
 
   def user_id
-    ENV['REMOTE_USER']
+    request.env['REMOTE_USER'] || ENV['REMOTE_USER']
+  end
+
+  def request_ldap
+    @request_ldap ||= begin
```
cbeer: Are you sure this construct works? At the very least, a `case` statement might be more clear.

jkeck: Yeah, I had just realized that, went out for a walk before I fixed it.

jkeck: And none of the LDAP stuff works so it's always just falling back on the configured groups until we yank it out. :/

PR #67, opened 2015-04-21
Adds style for input focus

jchristo4:
image
image

### DIFF
--- a/app/assets/stylesheets/sul-variables.scss
+++ b/app/assets/stylesheets/sul-variables.scss
@@ -7,6 +7,7 @@ $gray: #3f3c30;
 $gray-ccc: #ccc;
 $gray-41-percent: #696969;
 $pantone-401: #b6b1a9;
+$mostly-pure-orange: #eaab00;
 $white: #fff;
 $dark-cyan: #009b76;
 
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/sul-variables.scss
```diff
@@ -15,6 +16,7 @@ $sul-footer-bg-color: $beige-10-percent;
 $sul-h1-font-color: $gray;
 $sul-h2-border-color: $beige-30-percent;
 $sul-h2-font-color: $cardinal-red;
+$sul-form-input-focus: $mostly-pure-orange;
```
jkeck: It doesn't look like the `$sul-form-input-focus` is used anywhere (I guess `scss-lint` doesn't catch that).

app/assets/stylesheets/modules/forms.scss
```diff
@@ -1,4 +1,13 @@
 .form-group {
+  a.btn:focus,
```
jkeck: Should these rules apply outside of the `.form-group` class or should they be global to any input/button regardless of where it shows up?

app/assets/stylesheets/sul-variables.scss
jchristo4: @jkeck Fixed

app/assets/stylesheets/modules/forms.scss
jchristo4: @jkeck Added scss-mixin for using the outline style outside `.form-group` elements

jchristo4:
Had to add `!important` in mixin because bootstrap resets global outline styles

jkeck:
@jchristo4 it looks like Bootstrap stores the border focus color for inputs in the variable `$input-border-focus`.  Perhaps it's possible we can override that variable w/ our color variable, and then add the mixin (or straight css) to the link focus.  Let me know if going that route doesn't work and we can :ship: the `!important`.

PR #68, opened 2015-04-21
Updates rubocop gem version

jchristo4:


Gemfile
```diff
@@ -23,7 +23,7 @@ gem 'jquery-rails'
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem 'turbolinks'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
-gem 'jbuilder', '~> 2.0'
+gem 'jbuilder', '~> 2.2'
```
jkeck: I'm not sure we need to be this specific.  With the current specification we're already locked at [2.2.13](https://github.com/sul-dlss/sul-requests/blob/master/Gemfile.lock#L116) which is the [latest version](https://rubygems.org/gems/jbuilder/versions/2.2.13).

I believe the Gemnasium badge is still yellow due to two things.
1. Gemnasium cashes the status, and needs to be manually refreshed (by @sul-devops-team since they own that account).
2. Rubocop released [a new version](https://rubygems.org/gems/rubocop/versions/0.30.1) which we'll need to update to (would you like to submit a PR updating to that, and once we do we'll ask @sul-devops-team to manually refresh our Gemnasium status).
https://github.com/sul-dlss/sul-requests/pull/68/files#

PR #69, opened 2015-04-23
First pass at mediation admin UI

### LINKED ISSUE #29: Create mediation interface for admins
- [ ] Admins should get tables for every library that they can mediate (via LDAP workgroups?)
- [ ] What should admins that can mediate multiple libraries see?
- [ ] What should admins that can mediate ALL libraries see?

mediation

jchristo4:
Partially closes #29 

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,8 +28,10 @@ gem 'jbuilder', '~> 2.0'
 gem 'cancancan', '~> 1.10'
 # bundle exec rake doc:rails generates the API under doc/api.
 gem 'sdoc', '~> 0.4.0', group: :doc
-# Use faraday for making HTTP requests 
+# Use faraday for making HTTP requests
 gem 'faraday'
+# Use kaminari for pagination
+gem 'kaminari'
 # Use ActiveModel has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
 
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,0 +1,10 @@
+# Controller to handle mediations for admins
+class AdminController < ApplicationController
+  before_filter do
+    authorize! :manage, Request.new
+  end
+
+  def index
+    @requests = Request.order(:origin).group_by(&:origin)
+  end
+end
--- a/app/views/admin/index.html.erb
+++ b/app/views/admin/index.html.erb
@@ -1,0 +1,31 @@
+<h1>Requests</h1>
+
+<div class="admin-requests">
+  <% @requests.each do |origin, collection| %>
+    <% uniq_locations = collection.map { |h| h['origin_location'] }.uniq %>
+
+    <h3><%= LibraryLocation.library_name_by_code(origin) %></h3>
+    <h4>
+      Library=<%= origin %>
+      <% if uniq_locations.one? %>
+        &amp; location=<%= uniq_locations.first %>
+      <% end %>
+    </h4>
+
+    <table class="table table-striped">
+      <%= render 'table_header' %>
+      <tbody>
+        <% collection.each do |request| %>
+          <tr>
+            <td><%= request.type %></td>
+            <td><%= request.origin_location %></td>
+            <td><%= request.destination %></td>
+            <td><%= request.item_id %></td>
+            <td><%= format_date(request.created_at) %></td>
+          </tr>
+        <% end %>
+      </tbody>
+    </table>
+
+  <% end %>
+</div>
--- a/spec/controllers/admin_controller_spec.rb
+++ b/spec/controllers/admin_controller_spec.rb
@@ -1,0 +1,42 @@
+require 'rails_helper'
+
+describe AdminController do
+  before do
+    allow(controller).to receive_messages(current_user: user)
+  end
+
+  describe '/admin' do
+    describe 'for super admin' do
+      let(:user) { create(:superadmin_user) }
+      it 'should be accessible' do
+        get :index
+        expect(response).to be_successful
+      end
+    end
+    describe 'for site admin' do
+      let(:user) { create(:site_admin_user) }
+      it 'should be accessible' do
+        get :index
+        expect(response).to be_successful
+      end
+    end
+    describe 'for webauth user' do
+      let(:user) { create(:webauth_user) }
+      it 'should not be accessible' do
+        expect(-> { get :index }).to raise_error(CanCan::AccessDenied)
+      end
+    end
+    describe 'for non-webauth user' do
+      let(:user) { create(:non_webauth_user) }
+      it 'should not be accessible' do
+        expect(-> { get :index }).to raise_error(CanCan::AccessDenied)
+      end
+    end
+    describe 'for anon user' do
+      let(:user) { create(:anon_user) }
+      it 'should not be accessible' do
+        expect(-> { get :index }).to raise_error(CanCan::AccessDenied)
+      end
+    end
+  end
+end
--- a/spec/features/admin_requests_spec.rb
+++ b/spec/features/admin_requests_spec.rb
@@ -1,0 +1,25 @@
+require 'rails_helper'
+
+describe 'Viewing all requests' do
+  describe 'by a superadmin user' do
+    before do
+      stub_current_user(create(:superadmin_user))
+      create(:page, item_id: 2345)
+      create(:page, item_id: 2346, origin: 'SAL-NEWARK')
+    end
+    it 'should list data in tables' do
+      visit admin_index_path
+
+      expect(page).to have_css('h3', text: /Green Library/)
+      expect(page).to have_css('h4', text: /Library=GREEN & location=STACKS/)
+      expect(page).to have_css('td', text: '2345')
+
+      expect(page).to have_css('h3', text: /SAL Newark/)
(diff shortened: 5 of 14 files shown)


app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,6 @@
+# Controller to handle mediations for admins
+class AdminController < ApplicationController
+  def index
+    @requests = Request.order(:origin)
```
jkeck: Check out https://github.com/ryanb/cancan/wiki/authorizing-controller-actions#index-action

app/views/admin/index.html.erb
```diff
@@ -0,0 +1,42 @@
+<% origin = '' %>
+
+<h1>Requests</h1>
+<div class="admin-requests">
+
+<% for request in @requests %>
+
+  <% if origin != request.origin %>
+    <% if origin != '' %>
+      </tbody></table>
```
jkeck: I don't quite understand what's happening here.  Is this grouping the requests by origin into separate tables?

app/views/admin/index.html.erb
```diff
@@ -0,0 +1,42 @@
+<% origin = '' %>
+
+<h1>Requests</h1>
+<div class="admin-requests">
+
+<% for request in @requests %>
```
jkeck: For loops are okay, but typically the syntax used for this kind of think in ruby would be

```ruby
@requests.each do |request|
   ...
end

File: app/controllers/admin_controller.rb [hunk 4]
```diff
@@ -0,0 +1,6 @@
+# Controller to handle mediations for admins
+class AdminController < ApplicationController
+  def index
+    @requests = Request.all.group_by(&:origin).sort
```
jkeck: You'll probably want to use [#order](http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-order) for sorting active record queries, which will play nicer with Kaminari.

app/views/admin/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<h1>Requests</h1>
+
+<div class="admin-requests">
+  <% @requests.each do |origin, collection| %>
+    <% uniq_locations = collection.map { |h| h['origin_location'] }.uniq %>
+
+    <h3><%= SULRequests::Application.config.libraries[origin] %></h3>
+    <h4>
+      Library=<%= origin %>
+      <% if uniq_locations.size == 1 %>
```
jkeck: In rails you can use `uniq_location.one?` as a slightly shorter and more readable alternative.

spec/features/admin_requests_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+describe 'Viewing all requests' do
+  describe 'by a webauth user' do
+    before do
+      stub_current_user
+    end
+    it 'should list requests in tables' do
+      visit new_page_path(item_id: '1234', origin: 'GREEN', origin_location: 'STACKS')
+      click_button 'Send request'
+      visit new_page_path(item_id: '1235', origin: 'SAL-NEWARK', origin_location: 'STACKS')
+      click_button 'Send request'
```
jkeck: You can effectively do the same thing by calling
```ruby
Page.create!(item_id: '1234', origin: 'GREEN', origin_location: 'STACKS')
Page.create!(item_id: '1235', origin: 'SAL-NEWARK', origin_location: 'STACKS')
```
at the beginning of your test instead of needing the test to take 4 additional requests to complete.

On a side-note, it's probably about time we begin using `factory_girl` for creating the request objects so we don't have too much pain when the objects get more complex.  I'll submit an issue about doing that.

app/views/admin/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<h1>Requests</h1>
+
+<div class="admin-requests">
+  <% @requests.each do |origin, collection| %>
+    <% uniq_locations = collection.map { |h| h['origin_location'] }.uniq %>
+
+    <h3><%= SULRequests::Application.config.libraries[origin] %></h3>
```
jkeck: There is now a method on the `LibraryLocation` class we can use instead of the config directly (this wasn't in master when you submitted this PR initially).

https://github.com/sul-dlss/sul-requests/blob/master/app/models/library_location.rb#L22-L24

```ruby
LibraryLocation.library_name_by_code(origin)
```

spec/features/admin_requests_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+describe 'Viewing all requests' do
+  describe 'by a webauth user' do
+    let(:current_user) { create(:superadmin_user) }
+    before do
+      Page.create!(item_id: '1234', origin: 'GREEN', origin_location: 'STACKS')
+      Page.create!(item_id: '1235', origin: 'SAL-NEWARK', origin_location: 'STACKS')
```
jkeck: You can now use the `factory_girl` syntax to create these fixtures:

```ruby
create(:page)
create(:page, origin: 'SAL-NEWARK')
```

spec/features/admin_requests_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+describe 'Viewing all requests' do
+  describe 'by a webauth user' do
+    let(:current_user) { create(:superadmin_user) }
```
jkeck: In the context of a controller (controller/feature specs) you'll need to stub the current_user method on the controller.  I've added a helper method to do this `stub_current_user`.

By default it will create an anonymous user, but takes an optional parameter for a user object itself.

In your before block you can call 
```ruby
stub_current_user(create(:superadmin_user))
```
and this will create a `superadmin` user and stub the `current_user` method on the controller with that user object.

jkeck: (and you can get rid of the `let(:current_user)` business all-together)

spec/controllers/admin_controller_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+describe AdminController do
+  describe '#index' do
+    it 'should be successful' do
+      get :index
+      expect(response).to be_successful
```
jkeck: You aren't setting any user here, so you're testing that an anonymous user is going to get to the index action successfully (which they won't because we're calling `authorize!` in the index action).  This should be a test, but we should be asserting that CanCan is throwing an access denied error for anonymous users. (see other places where we test `CanCan::AccessDenied` since asserting errors thrown is a little tricky).

You can also have another test (or multiple for site admins, super admins, normal webauth users) where you stub the `current_user` w/ a created user of that type.

```ruby
stub_current_user(create(:superadmin_user))
stub_current_user(create(:site_admin_user))
stub_current_user(create(:webauth_user))
```

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,8 @@
+# Controller to handle mediations for admins
+class AdminController < ApplicationController
+
```
jkeck: One the specs pass RuboCop is going to complain about this empty space.  Just FYI.

Gemfile
```diff
@@ -28,6 +28,8 @@ gem 'jbuilder', '~> 2.0'
 gem 'cancancan', '~> 1.10'
 # bundle exec rake doc:rails generates the API under doc/api.
 gem 'sdoc', '~> 0.4.0', group: :doc
+# Use kaminari for pagination
+gem 'kaminari'
```
jkeck: Since you didn't do anything with Kaminari in this PR do you want to add a new ticket about adding kaminari pagination?

app/views/admin/index.html.erb
```diff
@@ -0,0 +1,35 @@
+<h1>Requests</h1>
+
+<% if can? :manage, Request %>
```
jkeck: I think we should remove this authorization check and just count on the `authorize!` call in the index action.

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,8 @@
...
+  def index
+    @requests = Request.all.group_by(&:origin).sort
+    authorize! :index, @requests
```
jkeck: Would something like `authorize! :manage, Request` work here?  If it does we can actually do the authorization before we hit the database (since we won't need the `@requests` variable).

spec/features/admin_requests_spec.rb
jchristo4: :heavy_check_mark:  done

spec/features/admin_requests_spec.rb
jchristo4: :heavy_check_mark: done

spec/controllers/admin_controller_spec.rb
jchristo4: Added tests for each type of user in `admin_controller_spec.rb` with `CanCan::AccessDenied` error assertions

Gemfile
jchristo4: :heavy_check_mark: done

app/controllers/admin_controller.rb
jchristo4: @jkeck Tried `authorize! :manage, Request` before `@requests` assignment. Didn't work. Ended up leaving the line as it is

jchristo4:
@jkeck Updated changes from https://github.com/sul-dlss/sul-requests/commit/285fa9828590908368b42097e0ac2d9f63f79b66

PR #73, opened 2015-04-24
Add model to get item data (title) from searchworks api and associated t...

jgreben:
...ests. Will review with Jessie on Monday as to any needed changes and what to do about testing Faraday response from searchworks.

### DIFF
--- a/spec/models/searchworks_item_spec.rb
+++ b/spec/models/searchworks_item_spec.rb
@@ -1,0 +1,92 @@
+require 'rails_helper'
+
+describe SearchworksItem do
+  let(:subject) { SearchworksItem.new('123') }
+
+  describe '#uri' do
+    it 'should return the base uri from the settings.yml file' do
+      expect(subject.send(:uri)).to eq('http://searchworks.stanford.edu')
+    end
+  end
+  describe '#url' do
+    it 'should return a url for the searchworks api' do
+      expect(subject.send(:url)).to eq('http://searchworks.stanford.edu/view/123.mobile?covers=false')
+    end
+  end
+  describe '#xml' do
+    it 'should return xml as the body of the response object' do
+      expect(subject.send(:xml)).to start_with('<?xml version="1.0" encoding="utf-8"?>')
+    end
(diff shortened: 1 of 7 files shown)


spec/models/searchworks_item_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+
+describe SearchworksItem do
+  let(:item_id) { SearchworksItem.new }
```
jkeck: If you want to use subject you can write:

```ruby
let(:subject) { SearchworksItem.new(item_id) }
```

You can then use subject in your assertions.  (e.g. `expect(subject.searchworks_uri).to eq('http://searchworks.stanford.edu')`)

jkeck: RSpec will automatically instantiate the class w/o any arguments as an implicit subject.  You can change that by defining your own subject.

jkeck: I should also mention that you don't HAVE to use subject, it's totally optional (but sometimes useful IMO).

jkeck: But if you don't use `subject` you'll still want to send the `has_title?` message to the `@item` variable instead of sending the `has_title?` message to the `@item.success?` boolean.

jgreben: So instead of `item` I am using `subject` now...
let(:subject) { SearchworksItem.new('123') }

This works:
`expect(subject.title?).to be true`
But not this:
`expect(subject).to have_title?`

jkeck: Is the name of your method now `#title?` instead of `#has_title?`?

jgreben: yes

jkeck:
:+1: looks good.  Just needs a squash and we're ready to :shipit:

jkeck:
I think this is due to https://www.traviscistatus.com/incidents/h91j8ft0qs2l

Stay :tv:

PR #97, opened 2015-04-29
Adds title to admin UI

### LINKED ISSUE #95: Update admin UI requests list with titles from database

jchristo4:
Closes #95 

image

### DIFF
--- a/app/views/admin/index.html.erb
+++ b/app/views/admin/index.html.erb
@@ -21,7 +21,8 @@
             <td><%= request.origin_location %></td>
             <td><%= request.destination %></td>
             <td><%= request.item_id %></td>
-            <td><%= request.item_title %></td>
+            <td><%= searchworks_link(request.item_id, request.item_title) %></td>
+            <td><%= requester_info(request.user) %></td>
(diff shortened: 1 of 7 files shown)


app/views/admin/index.html.erb
```diff
@@ -21,6 +21,7 @@
             <td><%= request.origin_location %></td>
             <td><%= request.destination %></td>
             <td><%= request.item_id %></td>
+            <td><%= request.item_title %></td>
```
jkeck: According to the mockups (page 37) this should be a link to the SearchWorks record.

sw-title

jkeck: Also, do you want to add the Requester while you're in here?  There's already a method on the user model that will give you the string you want for SUNet or Name+Email users (we don't handle library ID yet, so idk what we would do there).

https://github.com/sul-dlss/sul-requests/blob/master/app/models/user.rb#L11-L17

jchristo4: * Added SearchWorks link to title
* Added requester info and link

image

PR #107, opened 2015-05-01
Adds Go back link to the page and mediated page request forms when no-sunet-id button is clicked

jgreben:
Pull request to merge changes to add a back to login toggle link on no-sunet-id page and mediated_page requests.

screen shot 2015-05-01 at 10 04 03 am
screen shot 2015-05-01 at 10 04 20 am


app/views/shared/_back_to_login.html.erb
```diff
@@ -0,0 +1,3 @@
+<div data-go-back-link='true' style='display:none;line-height:3'>
+  <%= link_to "< Go back (show the login option)", 'javascript:;', class: "#{content_column_class} #{label_column_offset_class}", data: { no_sunet: true, target: '[data-no-sunet-content]', content_target: '[data-no-sunet-target]', link: '[data-go-back-link]' } %>
```
jvine: Better choice for the < character:

<span aria-hidden="true">&lsaquo; </span>

(creates a visible small arrow, but hides it from screen-reader that would try to read "less than")

app/assets/javascripts/toggle_name_and_email_input.js
```diff
@@ -0,0 +1,33 @@
...
+        content.append(link.show()).show();
+
+        if ($('#new_page').length) {
+          content.appendTo('#new_page');
+          store.appendTo('#new_page').hide();
+        }
+        else if ($('#new_mediated_page').length) {
+          content.appendTo('#new_mediated_page');
+          store.appendTo('#new_mediated_page').hide();
+        }
```
jkeck: Can we use something other than the ID of these elements here? I would be nice if, when we created a new kind of request form we didn't have to come into this plugin and ad another `else if` conditional here for it to work.

app/views/shared/_back_to_login.html.erb
```diff
@@ -0,0 +1,3 @@
+<div data-go-back-link='true' style='display:none;line-height:3'>
```
jkeck: I think the `display:none` is okay in a style tag when we want to be absolutely sure it's hidden immediately even before any css is loaded, but we should avoid styling things like `line-height` directly in the markup like this.

app/views/shared/_back_to_login.html.erb
```diff
@@ -0,0 +1,3 @@
...
+  <%= link_to "‹ Go back (show the login option)", 'javascript:;', class: "#{content_column_class} #{label_column_offset_class}", data: { no_sunet: true, target: '[data-no-sunet-content]', content_target: '[data-no-sunet-target]', link: '[data-go-back-link]' } %>
```
jkeck: Just in case it got lost in the fray, see https://github.com/sul-dlss/sul-requests/pull/107#discussion_r29518915

jkeck:
@jgreben anyway I can get you to put something more descriptive than "85 login link" as the subject of this PR?

app/views/shared/_back_to_login.html.erb
jgreben: Is there a particular css file that this should go in? Is it possible to style on the `data-go-back-link=‘true’` or should I add an id for this?

jkeck: I would add a class (styling data attributes is probably a bad idea in general).

Here are the list of css modules we load. https://github.com/sul-dlss/sul-requests/tree/master/app/assets/stylesheets/modules

I would say that `forms.scss` or `buttons.scss` are logical places for this to live.

app/views/shared/_back_to_login.html.erb
```diff
@@ -0,0 +1,3 @@
+<div data-go-back-link='true' class='go-back-link' style='display:none;'>
+  <%= link_to "‹ Go back (show the login option)", 'javascript:;', class: "#{content_column_class} #{label_column_offset_class}", data: { no_sunet: true, target: '[data-no-sunet-content]', content_target: '[data-no-sunet-target]', link: '[data-go-back-link]' } %>
```
jkeck: @jgreben just FYI, there was another comment on this file that was removed because the line was updated, but there is still one change to make.

The `<` in the link text is not correct.   See @jvine's comment at https://github.com/sul-dlss/sul-requests/pull/107#discussion_r29518915

jgreben: This is done.

app/views/shared/_back_to_login.html.erb
```diff
@@ -0,0 +1,5 @@
+<div data-go-back-link='true' class='go-back-link' style='display:none;'>
+  <%= link_to 'javascript:;', class: "#{content_column_class} #{label_column_offset_class}", data: { no_sunet: true, target: '[data-no-sunet-content]', content_target: '[data-no-sunet-target]', link: '[data-go-back-link]' } do %>
```
jkeck: I don't think you want the `content_column_class` here.  We really only need the offset and don't necessarily want the link element to take up 9 columns itself.  This makes the underline really far beneath the link and extend the full 9 columns.

bootstrap-column

app/views/shared/_back_to_login.html.erb
```diff
@@ -0,0 +1,5 @@
...
+    <span aria-hidden=\"true\">&lsaquo; </span>Go back (show the login option)
```
jkeck: Please remove the explicit quote escaping here.  It makes the actual attribute value different:

quote-escape

jkeck:
:+1: @jgreben.

Just squash these commits it it's ready to go.

jgreben:
Hound is taking a very long time to review this pull request.

jgreben:
It seems that Hound is taking a very very long time to review this PR. Is something wrong or is this normal?

- Josh

jkeck:
@jgreben not sure if it's related, but one issue I notice is that there is a commit that is already in master in you branch ab9a767b44d9e0e6ade534a65cc78f28521dfc72.

commits

It looks like if you run rebase on current master (remember to pull master first) and push that up it will resolve that issue.  Your PR should only have commits that you (or others that collaborated on the PR) contributed.

Hound may be responding slowly, and we can :shipit: if it won't complete.

jgreben:
I'm not sure what I did to make that SearchworksItem commit appear as part of the squashed commit, but I think rebasing fixed it.

PR #131, opened 2015-05-14
[WIP] Initial version of date-picker for mediated pages

### LINKED ISSUE #90: Implement Date Picker Widget
Use the date picker identified by #14

jchristo4:
Closes #90 

image

image


app/helpers/requests_helper.rb
```diff
@@ -16,6 +16,10 @@ def label_for_pickup_libraries_dropdown(pickup_libraries)
     end
   end
 
+  def label_for_needed_date()
+    'Planned date of use'
```
jkeck: I have started putting these labels into i18n keys.  See https://github.com/sul-dlss/sul-requests/blob/59-confirmation-emails/config/locales/en.yml#L43-L44 and https://github.com/sul-dlss/sul-requests/blob/59-confirmation-emails/app/models/request.rb#L103-L104

This will allow us to have a static label now, but we can add dynamic (per request type) labels if/when we need it for hold/recalls.

PR #147, opened 2015-05-19
Restrict scan requests to locations that include scannable item types.

### LINKED ISSUE #65: Update definition of what is scannable
Depends on #57 and possible #16 

We currently define scannable by the origin being "SAL3" and the origin_location being "STACKS".

This needs to be refined to limit things that have a resource_type of either "Book" or "Journal/Periodical".

At the item level, we'll restrict based on circulating item types. (For SAL3: STKS, STKS-MONO, and STKS-PERI).

jkeck:
Closes #65 

After this PR when requests are made from scannable library/locations they will also be checked to see if the requested location has any scannable holdings (based on item-type).

### DIFF
--- a/app/models/library_location.rb
+++ b/app/models/library_location.rb
@@ -2,12 +2,13 @@
 #  Class to handle configuration and logic around library codes and labels
 ###
 class LibraryLocation
-  attr_reader :library, :location
+  attr_reader :request, :library, :location
 
   include Scannable
   include Mediateable
 
   def initialize(request)
+    @request = request
     @library = request.origin
     @location = request.origin_location
   end
--- a/app/models/concerns/scannable.rb
+++ b/app/models/concerns/scannable.rb
@@ -3,6 +3,24 @@
 ###
 module Scannable
   def scannable?
-    @library == 'SAL3' && @location == 'STACKS'
+    scannable_library? &&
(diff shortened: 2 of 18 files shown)


app/models/library_location.rb
```diff
@@ -2,12 +2,13 @@
...
 ###
 class LibraryLocation
-  attr_reader :library, :location
+  attr_reader :request, :library, :location
 
   include Scannable
   include Mediateable
 
   def initialize(request)
+    @request = request
```
cbeer: I wouldn't hold up merging the PR over this, but I'm a little bothered by coupling a request and a location like this. Maybe a future refactor could make `Request` support `scannable_item?`, and a `LibraryLocation` support `scannable_location?` (or whatever the word for origin library + origin location actually is..).

cbeer:
Generally :+1:, but I don't feel confident enough to merge it myself.

app/models/library_location.rb
jkeck: Just for clarity we decided that we would discuss the request coupling during future work and identify a consistent patter (so this shouldn't hold up this PR).

cbeer:
Seems like others have had enough time to chime in if anything was amiss...

Issue #121, opened 2015-05-08
Add MHLD data for SAL-NEWARK requests


jkeck:
This should be below the free-text "Item(s) requested" textarea. Add this textarea for SAL-NEWARK requests if not completed in another PR.

- [ ] Public note?
- [ ] Library has?
- [ ] Latest received?

PR #149, opened 2015-05-20
Add status page for requests

### LINKED ISSUE #60: Create request status page
This should be viewable by the user that submitted the request.

Add a link to the confirmation email for non-webauth users that will allow them to view this page w/o authenticating. (possibly hash the requester's email with the request id and created_at date)

request-status

jkeck:
Closes #60 

This PR consolidates the request information information that is on all the requests success and status pages into one shared partial.

status-single-item
---
status-multi-item


config/routes.rb
```diff
@@ -21,10 +21,16 @@
     end
   end
 
+  concern :statusable do
```
cbeer: I wonder if `status` could actually be `show`? That said, maybe that's best left to a future refactor.

app/views/shared/_request_status_information.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      <%= LibraryLocation.library_name_by_code(request.destination) %>
+    </dd>
+  <% end %>
+
+  <% if request.commentable? && request.data['comments'].present? %>
+    <dt><%= t("forms.labels.comments") %></dt>
+    <dd><%= request.data['comments'] %></dd>
+  <% end %>
+
+  <% if request.data.present? %>
```
cbeer: Should these data elements get pulled up to the request (or some other model-like object, I suppose)?

app/views/shared/_status.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="<%= dialog_column_class %>">
```
cbeer: Does this really belong in `shared`, or in `requests` (where Rails templating could automatically discover it..)?

app/views/shared/_status.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <h1>Status of your request</h1>
+  <hr />
+  <dl class='dl-horizontal dl-invert'>
+    <dt>Requested on</dt>
+    <dd><%= request.created_at.to_formatted_s(:long_ordinal) %></dd>
```
cbeer: I don't know what `long_ordinal` is, but could this be e.g. `<%= l request.created_at %>`? [1]


[1] http://guides.rubyonrails.org/i18n.html#adding-date-time-formats

app/views/shared/_status.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  </dl>
+  <hr />
+  <%= render partial: 'shared/searchworks_item_information', locals: { request: request } %>
```
cbeer: I'm not a fan of passing the request around as a local. I guess we're doing this because we're dealing with a bunch of different resources with various instance variable names? Maybe we should consider loading them under the same name, e.g.:

```ruby
load_and_authorize_resource through: :exhibit, instance_name: 'request'
 ```

to make it easier to share common partials without `locals`?

(and this might be part of a larger refactor, not intended to hold up this PR)

cbeer:
Couple, hopefully minor, comments. https://github.com/sul-dlss/sul-requests/pull/149#discussion_r30710374 is the only one I'd like to see addressed here (if appropriate).

config/routes.rb
jkeck: I had always conceptually thought of `show` as something that an admin may do from the admin/mediation pages (although there isn't anything in the mocks about doing that) so when we said we want a status page I thought of it as something different (and easily controllable in this way to not have status pages for particular kinds of requests).  I could honestly go either way at this point though.

app/views/shared/_request_status_information.html.erb
jkeck: I honestly don't think they should be pulled into the request (but perhaps that's just my preference).  Different request objects may want to provide different pieces of data and this allows us to stay flexible in that respect without having to to lots of database and data migrations.

app/views/shared/_status.html.erb
jkeck: Yeah, TBH I've seen a few date formats around the app and think we should strictly identify how we want to format dates on the particular screens.  This was a built-in rails way that at least looked something like what was in the mock.  I'll create a ticket about validating the different date formats around the app and we can throw them into i18n for referencing around the app.

app/views/shared/_request_status_information.html.erb
cbeer: I didn't mean pull them up to columns in the database, just providing model-level accessors.

app/views/shared/_status.html.erb
cbeer: :+1:

app/views/shared/_status.html.erb
jkeck: Hmm, I feel like I've ran into nothing but trouble using CanCan and inheritance, but it may be worth a try.

I was also thinking of a `current_request` pattern that could be set in each controller.

cbeer: Hm. I stole that example from Spotlight, so we've been able to make it work before. `#current_request` would also work for me.

app/views/shared/_status.html.erb
jkeck: Let's talk about this a bit more at/after standup.  I agree with you, but I want to think of the best way to handle this.

cbeer: Sure. I wouldn't hold up the PR over it. It's easy enough to move around later.

app/views/shared/_request_status_information.html.erb
jkeck: Ah, okay.  Since there's a few things flying around wrt the kinds of data we're storing and what we call it (see #121) maybe we can wait until the dust settles on that and then add all the appropriate accessors and tests.  Would that be reasonable?

cbeer: Of course.

app/views/shared/_status.html.erb
jkeck: Yeah, Spotlight is where I have had trouble with it.  I know it works as we've implemented it but I found some rough edges working with it.

I'm working on implementing a `current_request` pattern.

jkeck:
Okay, I think we may be ready to :shipit:

app/views/shared/_status.html.erb
cbeer: Want to talk about this before :rocket: ?

jkeck: :+1:

PR #151, opened 2015-05-20
Added an ILLiad redirect to sulils-test and supporting tests

jgreben:
This includes technical debt for properly stubbing the illiad_query url in the tests and getting a callback response from sul-illiad-test to the successful scans page.

### DIFF
--- a/spec/features/create_scan_spec.rb
+++ b/spec/features/create_scan_spec.rb
@@ -1,10 +1,15 @@
 require 'rails_helper'
 
 describe 'Create Scan Request' do
+  before do
+    allow_any_instance_of(ScansController).to receive(:illiad_query).and_return('http://illiad.ill')
+  end
   describe 'by a webauth user' do
-    before { stub_current_user(create(:webauth_user)) }
+    before do
+      stub_current_user(create(:webauth_user))
+    end
     it 'should be possible' do
-      visit new_scan_path(item_id: '1234', origin: 'SAL3', origin_location: 'STACKS')
+      visit new_scan_path(item_id: '12345', origin: 'SAL3', origin_location: 'STACKS')
 
       page_range = '1-3, 7, 19-29'
       fill_in 'Page range', with: page_range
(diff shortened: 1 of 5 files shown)


spec/features/create_scan_spec.rb
```diff
@@ -55,7 +58,11 @@
 
       click_button 'Send request'
 
-      expect(page).to have_css('h1#dialogTitle', text: 'Request complete')
+      # expect(page).to have_css('h1#dialogTitle', text: 'Request complete')
```
jkeck: Cruft?

spec/features/create_scan_spec.rb
```diff
@@ -55,7 +58,11 @@
...
+
+      s = ScansController.new
```
jkeck: Cruft?

spec/features/create_scan_spec.rb
jgreben: In case we need it again for when the illiad callback starts working?

jkeck: I meant the commented `expect` in your test that matches the one above it character for character.  Do we need that commented out `expect`?  I don't quite understand what the H1 has to do w/ the Illiad callback.

jgreben: Cruft. I will remove it and re-commit, etc.

spec/features/create_scan_spec.rb
```diff
@@ -55,7 +57,9 @@
 
       click_button 'Send request'
 
-      expect(page).to have_css('h1#dialogTitle', text: 'Request complete')
+      s = ScansController.new
+
+      expect(s.illiad_query(Scan.last)).to eq('something')
```
jkeck: I feel like this particular assertion doesn't really belong here.  This is a unit test on a method of the ScansController, but the unit test is being executed in the context of a feature/integration test (not in the context of the ScansController spec).

A possible suggestion would be to change `something` in line 49 to be a url (e.g. http://test.com/ or http://example.com/) and then you can assert that the `current_url` (after clicking the `Send request` button) is what you set as the return value of the `illiad_query` method (effectively validating that we redirect to the result of that method).

It would then be nice to have unit test(s) in the scans_controller_spec that invokes the `illiad_query` method and actually asserts what it returns.  As of right now you there are no tests that validate anything about how the attributes of the scan object get translated into the URL string we send to Illiad.

PR #154, opened 2015-05-26
Make first form field have focus for pages and mediated pages

jgreben:
When you click 'I don't have a SUNet ID' the first form field should have focus. Also added tests for this in spec/features/.

### DIFF
--- a/app/assets/javascripts/toggle_name_and_email_input.js
+++ b/app/assets/javascripts/toggle_name_and_email_input.js
@@ -1,21 +1,27 @@
 (function($) {
   $.fn.toggleNameAndEmailInputs = function() {
     this.each(function(){
-      $(this).on('click', function(){
+      var contentForm = $(this).closest('form');
+
+      $(this).on('click', function() {
         var target = $($(this).data('target'));
         var content = $($(this).data('content-target'));
         var link = $($(this).data('link'));
         var targetScope = $(this).data('target');
         var store = target.detach();
-        var form = $('form').last();
 
         content.append(link.show()).show();
-        content.appendTo(form);
-        store.appendTo(form).hide();
+        content.appendTo(contentForm);
+        store.appendTo(contentForm).hide();
 
         if ( targetScope === '[data-no-sunet-content]' )
         {
           link.hide();
+        }
+        else if ( targetScope === '[data-no-sunet-target]' )
+        {
+          $(contentForm).find('input[type=text]')
+            .filter(':visible:first').focus();
         }
       });
     });
--- a/spec/features/create_mediated_page_request_spec.rb
+++ b/spec/features/create_mediated_page_request_spec.rb
@@ -7,9 +7,12 @@ describe 'Creating a mediated page request' do
       visit new_mediated_page_path(item_id: '1234', origin: 'SPEC-COLL', origin_location: 'STACKS')
       click_link "I don't have a SUNet ID"
 
+      expect(page).to have_field('Library ID', type: 'text')
       expect(page).to have_field('Name', type: 'text')
       expect(page).to have_field('Email', type: 'email')
       expect(page).to have_css('a', text: '‹ Go back (show the login option)')
+      expect(page).to have_css('input#mediated_page_user_attributes_library_id')
+      expect(page.evaluate_script('document.activeElement.id')).to eq 'mediated_page_user_attributes_library_id'
 
       click_link '‹ Go back (show the login option)'
       expect(page).to have_css('a', text: "I don't have a SUNet ID")
@@ -29,6 +32,9 @@ describe 'Creating a mediated page request' do
     it 'should be possible if a library id is filled out', js: true do
       visit new_mediated_page_path(item_id: '1234', origin: 'SPEC-COLL', origin_location: 'STACKS')
       click_link "I don't have a SUNet ID"
+
+      expect(page).to have_css('input#mediated_page_user_attributes_library_id')
+      expect(page.evaluate_script('document.activeElement.id')).to eq 'mediated_page_user_attributes_library_id'
 
       fill_in 'Library ID', with: '123456'
 
--- a/spec/features/create_page_request_spec.rb
+++ b/spec/features/create_page_request_spec.rb
@@ -12,6 +12,9 @@ describe 'Creating a page request' do
       visit new_page_path(item_id: '1234', origin: 'GREEN', origin_location: 'STACKS')
       click_link "I don't have a SUNet ID"
 
+      expect(page).to have_css('input#page_user_attributes_library_id')
+      expect(page.evaluate_script('document.activeElement.id')).to eq 'page_user_attributes_library_id'
+
       fill_in 'Name', with: 'Jane Stanford'
       fill_in 'Email', with: 'jstanford@stanford.edu'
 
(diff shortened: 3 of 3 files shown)


app/assets/javascripts/toggle_name_and_email_input.js
```diff
@@ -17,6 +17,10 @@
         {
           link.hide();
         }
+        else if ( targetScope === '[data-no-sunet-target]' )
+        {
+          $('form').find('input[type=text]').filter(':visible:first').focus();
```
jkeck: I feel like this is scoped too broadly.  If we were to add another form anywhere above this form in the DOM for any reason, this would unexpectedly attach focus to the first visible text input in **that** form.

Also, on line 10 there is a form variable (although I don't particularly like that either, because again, if we add a form after this one in the DOM we'll be scoped to the wrong form).

Here's a suggestion:

1. Change the variable in line 10 to use [`closest()`](https://api.jquery.com/closest/) to find the parent form.
2. Change this line to be scoped to finding inputs within that form.

spec/features/create_mediated_page_request_spec.rb
```diff
@@ -7,9 +7,11 @@
       visit new_mediated_page_path(item_id: '1234', origin: 'SPEC-COLL', origin_location: 'STACKS')
       click_link "I don't have a SUNet ID"
 
+      expect(page).to have_field('Library ID', type: 'text')
       expect(page).to have_field('Name', type: 'text')
       expect(page).to have_field('Email', type: 'email')
       expect(page).to have_css('a', text: '‹ Go back (show the login option)')
+      expect(page.execute_script("$('input#mediated_page_user_attributes_library_id').trigger('focus')"))
```
jkeck: I don't think this is doing what you're trying to accomplish.  RSpec's [`expect`](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers) requires that you assert something.  Here, you just have `expect(...)` where you would need `expect(...).to` **or** `expect(...).not_to`.  Simply calling `expect(...)` doesn't actually do anything (besides executing the code passed in).

Also, the `execute_script` here is not what you want.  That is causing the javascript you pass to be executed on the page, effectively causing the referenced input to have the focus event triggered by the test itself. See [`trigger()`](https://api.jquery.com/trigger/).

CSS already has a [`:focus`](http://www.w3schools.com/cssref/sel_focus.asp) selector that we can take advantage of here so we should be able to make an assertion like `expect(page).to have_css('input#mediated_page_user_attributes_library_id:focus')`

app/assets/javascripts/toggle_name_and_email_input.js
jgreben: > On May 26, 2015, at 4:57 PM, Jessie Keck <notifications@github.com> wrote:
> 
> In app/assets/javascripts/toggle_name_and_email_input.js <https://github.com/sul-dlss/sul-requests/pull/154#discussion_r31092053>:
> 
> > @@ -17,6 +17,10 @@
> >          {
> >            link.hide();
> >          }
> > +        else if ( targetScope === '[data-no-sunet-target]' )
> > +        {
> > +          $('form').find('input[type=text]').filter(':visible:first').focus();
> I feel like this is scoped too broadly. If we were to add another form anywhere above this form in the DOM for any reason, this would unexpectedly attach focus to the first visible text input in that form.
> 
> Also, on line 10 there is a form variable (although I don't particularly like that either, because again, if we add a form after this one in the DOM we'll be scoped to the wrong form).
> 
Isn’t this form scoped by calling this function?

$(document).on('ready page:load', function(){
  $('[data-no-sunet="true"]').toggleNameAndEmailInputs();
});

So the last form would always be the sibling next to the toggled content?
> Here's a suggestion:
> 
> Change the variable in line 10 to use closest() <https://api.jquery.com/closest/> to find the parent form.
closest() broke it I think because it is looking up the DOM and the form is the last sibling.

spec/features/create_mediated_page_request_spec.rb
```diff
@@ -7,9 +7,11 @@
       visit new_mediated_page_path(item_id: '1234', origin: 'SPEC-COLL', origin_location: 'STACKS')
       click_link "I don't have a SUNet ID"
 
+      expect(page).to have_field('Library ID', type: 'text')
       expect(page).to have_field('Name', type: 'text')
       expect(page).to have_field('Email', type: 'email')
       expect(page).to have_css('a', text: '‹ Go back (show the login option)')
+      expect(page).to have_css('input#mediated_page_user_attributes_library_id', :focus)
```
jkeck: Try commenting our the javascript that applies the focus and see if this test still passes.  I have a feeling it will pass regardless because `:focus` is not an appropriate filter for `have_css`.

app/assets/javascripts/toggle_name_and_email_input.js
```diff
@@ -7,16 +7,20 @@
         var link = $($(this).data('link'));
         var targetScope = $(this).data('target');
         var store = target.detach();
-        var form = $('form').last();
+        var contentForm = $('form:last');
```
jkeck: This is absolutely leaking outside of the context of this plugin.  The jQuery selector will select anything on the page that it is called (regardless of how the plugin was instantiated).  That's just how jQuery works.

I'm not sure why `closest()` didn't work however that doesn't mean we should allow our plugin's behavior to leak outside of our scope.

jkeck: WRT `closest()`, you may have had a scoping issue.  Try moving the `contentForm` variable assignment outside of the click event.

```javascript
this.each(function(){
  var contentForm = $(this).closest('form');
  $(this).on('click', function(){
  ...
```

spec/features/create_mediated_page_request_spec.rb
jgreben: > On May 26, 2015, at 11:29 PM, Jessie Keck <notifications@github.com> wrote:
> 
> In spec/features/create_mediated_page_request_spec.rb <https://github.com/sul-dlss/sul-requests/pull/154#discussion_r31106379>:
> 
> >        expect(page).to have_field('Name', type: 'text')
> >        expect(page).to have_field('Email', type: 'email')
> >        expect(page).to have_css('a', text: '‹ Go back (show the login option)')
> > +      expect(page).to have_css('input#mediated_page_user_attributes_library_id', :focus)
> Try commenting our the javascript that applies the focus and see if this test still passes. I have a feeling it will pass regardless because :focus is not an appropriate filter for have_css.
> 

Yes, that does pass the test regardless.  

I tried `have_css('input#mediated_page_user_attributes_library_id:focus’)` but there were no matches. 

I also tried `expect(page.evaluate_script('$("input#mediated_page_user_attributes_library_id").is(":focus")')).to be_true` but the script evaluates to false. 

I cannot seem to find any test that will detect whether the field is focused, which makes me wonder if the test is actually doing or registering the field focus.

jkeck: Okay, so this is a bit hacky but I believe should work.  Replace that expectation (and the others like it) with:

```ruby
expect(page).to have_css('input#mediated_page_user_attributes_library_id')
expect(page.evaluate_script('document.activeElement.id')).to eq 'mediated_page_user_attributes_library_id'
```

PR #155, opened 2015-05-26
Set checked state on item selector to handle Back-Forward Cache.

### LINKED ISSUE #152: Scan request can exceed item selection limit on page refresh
For scans (and also perhaps mediated pages) when you have a item checked and you refresh the page, you are able to select another `n` items according to the defined limits (a turbolinks issue?)

jkeck:
Closes #152 

I don't believe that page-refresh is actually a problem (as it clears out the checkboxes) but clicking navigating forward/backward in browser history was the actual issue.  @jgreben perhaps you can confirm since you unearthed the bug.

### Before
item-selector-before

### After
item-selector-after

### DIFF
--- a/app/assets/javascripts/limit_selected_items.js
+++ b/app/assets/javascripts/limit_selected_items.js
@@ -6,8 +6,14 @@
       var counter = $(itemSelector.data('counter-target'));
       var checkboxSelector = 'input[type="checkbox"]';
       var checkboxes = $(checkboxSelector, itemSelector);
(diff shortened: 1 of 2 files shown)


jgreben:
I actually saw this bug on the scans request page, not the mediated request, but it appears to be fixed now. Would you like me to merge this PR?

jkeck:
The same javascript is executed regardless of what kind of request it is.

:+1: yes, please merge if it looks good!

PR #159, opened 2015-05-28
Added illiad_params to illiad_query

jgreben:
Added a couple of new params so that the OpenURL request can be directly imported into ILLiad. Also added javascript to relevant illiad pages to process the OpenURL requests automatically and then send the user back to the successful scans page.

https://gist.github.com/jgreben/8262ba57425e39fbc1b3
https://gist.github.com/jgreben/e65237a065b8da544ec7

### DIFF
--- a/app/controllers/scans_controller.rb
+++ b/app/controllers/scans_controller.rb
@@ -35,7 +35,7 @@ class ScansController < RequestsController
 
   def illiad_query(scan)
     illiad_params = {
-      'rft.genre': 'scananddeliverArticle',
+      'Action': '10', 'Form': '30', 'rft.genre': 'scananddeliverArticle',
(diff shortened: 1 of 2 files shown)


app/controllers/scans_controller.rb
```diff
@@ -35,7 +35,7 @@ def current_request
 
   def illiad_query(scan)
     illiad_params = {
-      'rft.genre': 'scananddeliverArticle',
+      'Action': '10', 'Form': '30', 'rft.genre': 'scananddeliverArticle',
```
jkeck: Perhaps add these to the the test @ https://github.com/sul-dlss/sul-requests/blob/master/spec/controllers/scans_controller_spec.rb#L91-L97 so we can make sure we assert they're important.

PR #165, opened 2015-06-03
Display a copyright restrictions notice on the scan form

cbeer:
screen shot 2015-06-03 at 10 45 02 am
screen shot 2015-06-03 at 10 45 07 am

### DIFF
--- a/app/assets/stylesheets/modules/scan-or-deliver.scss
+++ b/app/assets/stylesheets/modules/scan-or-deliver.scss
@@ -41,3 +41,40 @@
     }
   }
 }
+
+.panel-danger {
+  .panel-heading {
+    border-bottom: 0;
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/modules/scan-or-deliver.scss
```diff
@@ -41,3 +41,40 @@
...
+
+    a {
+      border-bottom: 0;
+    }
+
+    .glyphicon {
+      line-height: 1.3em;
+    }
+
+    .glyphicon-chevron-down {
```
cbeer: I'm not thrilled by this, but couldn't find it in me to find the bootstrap-collapse-twiddling code we've done elsewhere.. at least until we decide whether we **can** collapse it, or if we're showing this note at all.

PR #166, opened 2015-06-04
Add Hold/Recall Form

### LINKED ISSUE #142: Add Hold/Recall Form

jkeck:
Closes #142

Example record:
`/requests/new?barcode=36105212925395&item_id=10785714&origin=GREEN&origin_location=STACKS`

hold-recall-form

### DIFF
--- a/app/controllers/mediated_pages_controller.rb
+++ b/app/controllers/mediated_pages_controller.rb
@@ -31,7 +31,7 @@ class MediatedPagesController < RequestsController
     if !current_user.webauth_user? && create_via_post?
       redirect_to login_path(
         referrer: create_mediated_pages_path(
-          page: params[:mediated_page].except(:user_attributes)
+          mediated_page: local_object_param.except(:user_attributes)
         )
       )
     else
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -5,7 +5,9 @@
 class Request < ActiveRecord::Base
   include Requestable
 
-  delegate :scannable?, :mediateable?, :pageable?, to: :library_location
+  attr_accessor :requested_barcode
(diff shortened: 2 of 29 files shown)


app/models/request.rb
```diff
@@ -80,8 +85,9 @@ def find_existing_user
   end
 
   def holdings
-    if persisted?
-      @holdings ||= searchworks_item.requested_holdings.by_barcodes(barcodes)
+    if persisted? || requested_barcode.present?
```
jkeck: This is just temporary, I would like to find a better way of doing this in general.

app/views/hold_recalls/_header.html.erb
```diff
@@ -0,0 +1 @@
+<h1 id='dialogTitle'>Request [checked-out | on-order] item</h1>
```
jkeck: I need to update this.  We don't have the tooling to know if the item is one or the other, so I need to build that in.

app/controllers/mediated_pages_controller.rb
```diff
@@ -27,11 +27,12 @@ def validate_mediated_pageable
     fail UnmediateableItemError unless @mediated_page.mediateable?
   end
 
+  # this should not be page param (needs test)
```
cbeer: Is this a TODO, or a ticket that should be made?

jkeck: Good catch. Thanks! This was actually addressed in 215035ebbfb39c8af2052cfdc520cef2ee1a2417 but I neglected to remove the comment.  Just took that out and pushed the branch back up.

cbeer:
:+1:

PR #173, opened 2015-06-09
Add proxy access lookup model

### LINKED ISSUE #160: Add proxy request integration
These are params we can send to symphony when we make the request.

We will need to make a call out to a proxy script on symphony to know if the person is in the appropriate symphony group to have proxy requests.

proxy

cbeer:
Fixes #160 

screen shot 2015-06-09 at 4 14 24 pm

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -70,7 +70,7 @@ class RequestsController < ApplicationController
 
   def rescue_can_can(exception)
     if !current_user.webauth_user? && create_via_post? && current_request.new_record?
-      request_params = local_object_param.except(:user_attributes)
+      request_params = params[:request].except(:user_attributes)
       redirect_to login_path(referrer: polymorphic_path([:create, current_request], request: request_params))
     else
       super
@@ -78,7 +78,9 @@ class RequestsController < ApplicationController
   end
 
   def new_params
-    validate_new_params
+    params.require(:origin)
+    params.require(:item_id)
+    params.require(:origin_location)
 
     params.permit(:origin, :item_id, :origin_location, :barcode)
   end
@@ -98,10 +100,6 @@ class RequestsController < ApplicationController
     params.require(:request).permit(:needed_date)
   end
 
-  def local_object_param
-    params[:request]
-  end
-
   def check_if_proxy_sponsor
     return unless current_request.user.sponsor? && params[:request][:proxy].nil?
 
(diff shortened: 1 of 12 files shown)


app/controllers/requests_controller.rb
```diff
@@ -23,6 +24,12 @@ def current_request
 
   protected
 
+  def check_if_proxy_sponsor
+    return if params[:proxy].present? || !current_user.sponsor?
```
jkeck: I'm receiving an error that traces back to this line when submitting a page request with a library ID.  (I guess that means we've been missing an important integration test about submitting w/ Library ID only).

In the case when a request is submitted w/ a Library ID only (or name+email) the `current_user` is still anonymous.  The user object that *I think* we're interested in is accessible through the `current_request`.

cbeer: Can you help me replicate this? I tried a simple page with just a library id and just name / email and it worked for me.

app/views/requests/sponsor_request.html.erb
```diff
@@ -0,0 +1,17 @@
+<h1 id="dialogTitle">Share with your proxy group?</h1>
```
jkeck: There is a div that wraps the other action's views that will show up in a dialog (while the home page and the admin pages get full-width).

It's possible that we need to do something more clever with layouts, but in the meantime this page looks vastly different from the other pages.  With the addition of some wrapper divs and making the buttons 12 column we can get a more consistent display.

### Current
sponsor-form-before

### Modified
sponsor-form-after


```ruby
<div class='<%= dialog_column_class %>'>
  <h1 id="dialogTitle">Share with your proxy group?</h1>

  <%= render 'form_flashes' %>
  <%= render 'searchworks_item_information' %>
  <div class="<%= "#{label_column_offset_class} #{content_column_class}" %>">
    <%= bootstrap_form_for(current_request) do |f| %>
      <%= hash_to_hidden_fields(request_params) %>

      <h2><%= render_user_information %></h2>
      <%= f.form_group help: 'Anyone in your group can check the status of this request and pick it up when it’s available.' do %>
        <%= f.button 'Yes, share with my group.', class: 'btn btn-default col-md-12', value: 'true', name: 'proxy' %>
      <% end %>

      <%= f.form_group help: 'Only you can see and pick up this request.' do %>
        <%= f.button 'No, just me.', class: 'btn btn-cardinal col-md-12', value: 'false', name: 'proxy' %>
      <% end %>
    <% end %>
  </div>

</div>

```

cbeer: Thanks, I've incorporated these changes.

app/views/requests/sponsor_request.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <%= render 'form_flashes' %>
+  <%= render 'searchworks_item_information' %>
+
+  <div class="<%= "#{label_column_offset_class} #{content_column_class}" %>">
+    <%= bootstrap_form_for(current_request) do |f| %>
+      <%= hash_to_hidden_fields(request_params) %>
+
+      <h2><%= render_user_information %></h2>
+      <%= f.form_group help: 'Anyone in your group can check the status of this request and pick it up when it’s available.' do %>
+        <%= f.button 'Yes, share with my group.', class: 'btn btn-default', value: 'true', name: 'request[proxy]' %>
```
jkeck: If you add `col-xs-12` to the button class these will behave like the similar buttons on the form.

jkeck:
:+1: one minor comment, not a blocker.

cbeer:
Sure, makes sense to me (although maybe we should refactor that into the CSS some time?)

app/views/requests/sponsor_request.html.erb
cbeer: Fixed in 92769b3.

jkeck:
:+1: I think we currently have 3 button types.  Normal buttons, form buttons, and scan/page chooser buttons.  Maybe wait until they all shake out to figure out how to organize their styles.

PR #174, opened 2015-06-09
Allow materials in Newark w/o barcodes to be commentable.

### LINKED ISSUE #121: Add MHLD data for SAL-NEWARK requests

jkeck:
Closes #121 

### Form
sal-newark-form

### Success Page
success-page

### DIFF
--- a/app/models/searchworks_item.rb
+++ b/app/models/searchworks_item.rb
@@ -64,6 +64,8 @@ class SearchworksItem
   #  holdings to just what was requested by the user
   ###
   class RequestedHoldings
+    delegate :mhld, to: :location
+
     def initialize(searchworks_item)
       @searchworks_item = searchworks_item
     end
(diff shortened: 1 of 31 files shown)


app/models/searchworks_item.rb
```diff
@@ -81,6 +83,12 @@ def all
       location.items
     end
 
+    def barcoded_holdings
+      @barcoded_holdings ||= all.select do |item|
+        item.barcode.starts_with?('36105')
```
cbeer: 36105 seems pretty magic. Can we extract this to a method and/or configuration?

jkeck: :+1: pushed the pattern into a method.

PR #177, opened 2015-06-10
Proxy refactor

cbeer:
This extracts some of the clean-up work leading up to #173


spec/helpers/application_helper_spec.rb
```diff
@@ -46,4 +46,30 @@
...
+    before do
+      helper.extend(Module.new do
+        def current_user; end
+      end)
+
+      allow(helper).to receive(:current_user).and_return(user)
+    end
+
+    let(:user) do
+      User.new name: 'Some Body', webauth: 'xyz'
```
jkeck: We could use `build(:webauth_user, name: 'Some Body')` to take advantage of `FactoryGirl`.

cbeer: :+1:

cbeer:
@jkeck thanks, I've addressed those two items and rebased them as appropriate.

PR #179, opened 2015-06-11
Changed ILLiad OpenURL params to correctly place item information in ILLiad

jgreben:
Changed ILLiad OpenURL params to correctly place item information in illiad client and to comply with rubocop's demands.

### DIFF
--- a/app/controllers/scans_controller.rb
+++ b/app/controllers/scans_controller.rb
@@ -33,17 +33,30 @@ class ScansController < RequestsController
 
   protected
 
+  def illiad_page
+    @illiad_page = { 'Action': '10', 'Form': '30', 'rft.genre': 'scananddeliverArticle' }
+  end
+
+  def illiad_item_data(scan)
+    @illiad_item_data = { 'rft.location': scan.origin,
+                          'rft.callnum': scan.holdings.first.callnumber,
+                          'rft.item': scan.holdings.first.barcode }
+  end
+
+  def illiad_citation(scan)
+    @illiad_params = { 'rft.jtitle': scan.item_title,
+                       'rft.au': scan.data[:authors],
+                       'rft.pages': scan.data[:page_range],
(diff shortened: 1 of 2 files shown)


app/controllers/scans_controller.rb
```diff
@@ -33,17 +33,30 @@ def current_request
 
   protected
 
+  def illiad_page
+    @illiad_page = { 'Action': '10', 'Form': '30', 'rft.genre': 'scananddeliverArticle' }
```
jkeck: In the context of the controller you typically set an instance variable (`@variable`) to make something accessible to the view layer.  Since you're calling these method directly you can simply use the return value of the method and don't need to set any instance variable.

app/controllers/scans_controller.rb
```diff
@@ -33,17 +33,30 @@ def current_request
...
+                       'rft.au': scan.data[:authors],
+                       'rft.pages': scan.data[:page_range],
+                       'rft.atitle': scan.data[:section_title] }
+  end
+
+  def illiad_callback(scan)
+    @illiad_callback = { 'scan_referrer': successful_scan_url(scan) }
+  end
+
   def illiad_query(scan)
```
jkeck: This seems to be getting a bit larger than the context of the controller.  It's possible that it may be time to refactor all these methods out of the controller class and into its own class or module.

jgreben: Already started doing this on Chris’ suggestion. I noticed after trying rake that I also need to create a new spec file too.

jkeck: :+1: I wasn't aware you guys talked it through.

jgreben: We didn’t talk anything through. He just mentioned it at the standup, so I started giving it a try, but it looks like master has now changed, and my commits will now cause many conflicts, so back to square 1.

jgreben:
Starting over.

PR #183, opened 2015-06-12
Add js module for the item selector and two separate modules for…

jkeck:
…updating the item count and enforcing selection limit.


app/assets/javascripts/item_selector/item_selector_limit.js
```diff
@@ -0,0 +1,64 @@
+var itemSelectorLimit = (function() {
+
+  var selectorLimit = function(_this) {
+    return parseInt(
+      _this.selectorElement()
+           .find('[data-limit-selected-items]')
+           .data('limit-selected-items'), 10);
+  };
+
+  return $.extend(itemSelector, {
```
jkeck: ಠ_ಠ Yes it is

app/assets/javascripts/item_selector/item_selector_incrementor.js
```diff
@@ -0,0 +1,51 @@
+var itemSelectorIncrementor = (function() {
+  return $.extend(itemSelector, {
```
jkeck: ಠ_ಠ Yes it is

app/assets/javascripts/item_selector/item_selector_incrementor.js
```diff
@@ -0,0 +1,51 @@
...
+
+        $(this).on('item-selector:deselected', function() {
+          _this.decreaseSelectedItemCount();
+        });
+      });
+    },
+
+    increaseSelectedItemCount: function() {
+      this.selectedItemCounter().text(
+        parseInt(this.selectedItemCounter().text(), 10) + 1
```
cbeer: You were unable to use `numberOfSelectedCheckboxes` here?

jkeck: I was not.  It didn't seem like it was correctly set in order that it would have needed to be.  (maybe would have needed a promise or something to work like that, but that seemed like a lot of overhead).

PR #208, opened 2015-07-02
Add paging schedule estimates

### LINKED ISSUE #15: Allow for paging schedules to be defined
Just a sketch of a possible DSL.

```ruby
when_paging from: "SAL3", to: "GREEN", before: "12:00pm" do
  will_arrive "11:00am"
  after_days 2
end

when_paging from: "SAL3", to: "GREEN", after: "12:00pm" do
  will_arrive "11:00am"
  after_days 3
end
```

jkeck:
Closes #15 

## Admin Listing
paging-schedule-admin

## Single Destination
single-lib-destination

## Multiple Destinations
paging-estimate-change

### DIFF
--- a/app/assets/javascripts/paging_schedule_updater.js
+++ b/app/assets/javascripts/paging_schedule_updater.js
@@ -1,0 +1,83 @@
+var pagingScheduleUpdater = (function() {
+
+  var defaultOptions = {
+    selector: '[data-paging-schedule-updater="true"]',
+    urlSelector: '[data-scheduler-lookup-url]',
+    textSelector: '[data-scheduler-text]',
+    singleLibrarySelector: '[data-single-library-value]'
+  };
+
+  return {
+    init: function(opts) {
+      var _this = this;
+      _this.options = $.extend(defaultOptions, opts);
+      $(document).on('ready page:load', function(){
+        _this.requestPagingScheduleData(); // Set initial schedule state
+        _this.addSelectChangeBehavior();
+      });
+    },
+
+    options: {},
+
+    addSelectChangeBehavior: function() {
+      var _this = this;
+      _this.selectElement().on('change', function() {
+        _this.requestPagingScheduleData();
+      });
+    },
+
+    requestPagingScheduleData: function() {
+      var _this = this;
+      var originalValue = _this.schedulerValue();
+      if ( originalValue && _this.schedulerText().length > 0 ) {
+        $.ajax({
+          url: _this.schedulerUrl(_this.schedulerValue())
+        }).success(function(data) {
+          if ( originalValue == _this.schedulerValue()) {
+            _this.updateSchedulerText(data);
+          }
+        }).fail(function() {
+          _this.updateSchedulerText({'text': 'No estimate available'});
+        });
+      }
+    },
+
+    schedulerValue: function() {
+      if ( this.selectElement().val() ) {
+        return this.selectElement().val();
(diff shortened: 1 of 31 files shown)


app/assets/javascripts/paging_schedule_updater.js
```diff
@@ -0,0 +1,80 @@
...
+      var _this = this;
+      _this.selectElement().on('change', function() {
+        _this.requestPagingScheduleData();
+      });
+    },
+
+    requestPagingScheduleData: function() {
+      var _this = this;
+      if ( _this.schedulerValue() && _this.schedulerText().length > 0 ) {
+        $.ajax({
```
cbeer: Do we need some kind of spinner when we're loading the data (in case the request takes some time..)?

app/assets/javascripts/paging_schedule_updater.js
```diff
@@ -0,0 +1,80 @@
...
+        return this.selectElement().val();
+      }else if ( this.singleLibraryValue() ) {
+        return this.singleLibraryValue();
+      }
+    },
+
+    updateSchedulerText: function(data) {
+      var _this = this;
+      if(_this.schedulerText().text() != data.text) {
+        _this.schedulerText().text(data.text);
```
cbeer: Should we check if the data returned is still useful (e.g. the user hasn't changed the destination while we were loading the data)?

app/helpers/requests_helper.rb
```diff
@@ -28,15 +28,20 @@ def select_for_multiple_libraries(form, pickup_libraries)
     return unless pickup_libraries.keys.length > 1
     form.select :destination,
                 pickup_libraries_array(pickup_libraries),
-                label: label_for_pickup_libraries_dropdown(pickup_libraries),
-                selected: 'GREEN'
+                { label: label_for_pickup_libraries_dropdown(pickup_libraries), selected: 'GREEN' },
```
cbeer: I know this wasn't a new addition, but:

`GREEN` is here because it's the default library? Should we move it to a helper or configuration or something?

app/helpers/requests_helper.rb
```diff
@@ -28,15 +28,20 @@ def select_for_multiple_libraries(form, pickup_libraries)
...
+                'aria-controls' => 'scheduler-text',
+                'data-paging-schedule-updater' => 'true'
```
cbeer: Both aria and data attributes could be written e.g.:

```
aria: { controls: 'scheduler-text' }
```

Since it's only a single value, i don't know if it matters much.

config/application.rb
```diff
@@ -19,6 +19,9 @@ class Application < Rails::Application
     # Load all request types automatically
     config.autoload_paths += %W( #{config.root}/app/models/requests )
 
+    require 'library_hours'
```
cbeer: Are these requires necessary, or can autoloading do it for us?

config/paging_schedule.rb
```diff
@@ -0,0 +1,105 @@
+PagingSchedule.configure do
```
cbeer: I'm not clear why `config/paging_schedule` and `config/initializers/paging_schedule` are both needed. I'd naively expect this content in the initializer.

cbeer:
:+1: good enough for me, just some comments + questions that could be ignored or slated for the future.

I fix the `term-ansicolor` build failure in #207 (c57bd2c)

PR #240, opened 2015-07-21
Limit scan and deliver eligibility

cbeer:
The flash message isn't in the same position as the design doc, but I'm hoping we can fix that globally..

screen shot 2015-07-21 at 10 11 57 am

### DIFF
--- a/app/controllers/scans_controller.rb
+++ b/app/controllers/scans_controller.rb
@@ -4,6 +4,28 @@
 class ScansController < RequestsController
   protected
 
+  def rescue_can_can(exception)
+    if action_name == 'create' && current_user && current_user.webauth_user? && !can?(:create, Scan)
+      # if they are logged in, but not eligible, send the user to the appropriate page for
+      redirect_to delegated_new_request_path(@request, new_request_params), flash: {
+        html_safe: true,
+        error: render_to_string(partial: 'ineligible_exception')
+      }
+    else
+      super
+    end
+  end
+
+  def new_request_params
+    {
+      origin: @request.origin,
+      origin_location: @request.origin_location,
+      item_id: @request.item_id,
+      barcodes: @request.barcodes,
+      modal: params[:modal]
+    }
+  end
+
   def redirect_to_success_with_token
     redirect_to illiad_query(current_request)
   end
--- a/app/views/scans/_ineligible_exception.html.erb
+++ b/app/views/scans/_ineligible_exception.html.erb
@@ -1,0 +1,5 @@
+<strong class="h4">Scan-to-PDF not available.</strong>
+
+<p>
+This pilot service is currently available only to faculty, grad students, postdocs, academic staff, fellows, and visiting scholars. Your feedback will help us refine and expand the service!
+</p>
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -16,7 +16,7 @@ describe ApplicationController do
     end
     it 'should return the ldap groups as an array' do
       allow(controller).to receive_messages(user_id: 'some-user')
-      allow(controller).to receive_messages(request_ldap: 'ldap:group1|ldap:group2')
+      allow(controller).to receive_messages(ldap_attributes: { 'WEBAUTH_LDAPPRIVGROUP' => 'ldap:group1|ldap:group2' })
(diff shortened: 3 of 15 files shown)


app/controllers/scans_controller.rb
```diff
@@ -4,6 +4,28 @@
...
+      redirect_to delegated_new_request_path(@request, new_request_params), flash: {
+        html_safe: true,
+        error: render_to_string(partial: 'ineligible_exception')
+      }
+    else
+      super
+    end
+  end
+
+  def new_request_params
```
cbeer: Can you think of a better way to handle this parameter translation?

jkeck: Hrrmm, not off the top of my head. :/  Sorry.  I'll think about it a bit more.

app/views/scans/_ineligible_exception.html.erb
```diff
@@ -0,0 +1,5 @@
+<strong class="h3">Scan-to-PDF not available.</strong>
+
+<p>
+This pilot service is currently available only to faculty, grad students, postdocs, academic staff, fellows, and visiting scholars. Your feedback will help us refine and expand the service!
+</p>
```
jkeck: This content is currently in an i18n string.  I think it was used somewhere early on in the project, but I think it has been removed (if it ever even made it into the UI). I think we should either set the flash message to the i18n string or kill it and define the message here (as you have done).

It seems like there is some logic around making the partial HTML safe for the flash message, and I wonder if the *_html i18n string would give us that for free.

spec/controllers/scans_controller_spec.rb
```diff
@@ -115,13 +115,27 @@
       end
     end
     describe 'invalid requests' do
-      let(:user) { create(:webauth_user) }
+      let(:user) { create(:webauth_user).tap { |u| u.ldap_group_string = Settings.scan_pilot_groups.first } }
```
jkeck: I wonder if we should create a special FactoryGirl user factory for users who can scan and set this `ldap_group_string` there (so we can re-use that particular user logic around different tests), like the one in the `create_scan_spec`.

spec/controllers/application_controller_spec.rb
```diff
@@ -28,4 +28,20 @@
...
+      allow(controller.request).to receive(:env).and_return('a' => 1)
+      expect(controller.send(:ldap_attributes)).to include 'a' => 1
+    end
+
+    it 'enriches the request env with local settings' do
+      allow(controller).to receive_messages(user_id: 'some-user')
+      allow(controller.request).to receive(:env).and_return('a' => 1)
+      allow(Rails.env).to receive(:development?).and_return(true)
+      Settings.fake_ldap_attributes ||= {}
+      allow(Settings).to receive(:fake_ldap_attributes).and_return('some-user' => { 'b' => 2 })
```
jkeck: This kinda surprised me a bit. Can you tell me more about what's going on here w/ this test (stubbing `#development?`, setting a rails_config setting if not set, then stubbing it).  I'm guessing there is a logical reason to do this, I'm just trying to figure out what it is.

cbeer: 100% test coverage. The method was incidentally used before (although, how, I'm not sure), and wasn't after the refactor.

spec/controllers/scans_controller_spec.rb
cbeer: :+1:

app/views/scans/_ineligible_exception.html.erb
cbeer: We need to do that silly flash[:html_safe] thing due to the way flash messages are stored across requests (which loses the html_safe bit). I think this needs to be a partial, not an i18n string, when we convert `"Your feedback"` into a link as designed.

jkeck: :+1: can we remove the i18n key so there is no confusion about where that text comes from?

spec/controllers/application_controller_spec.rb
jkeck: :+1:

app/views/scans/_ineligible_exception.html.erb
cbeer: :ok_hand:

PR #245, opened 2015-07-22
Add customizations to html5 date field for mediated pages

### LINKED ISSUE #90: Implement Date Picker Widget

jkeck:
Closes #90 (I think)

- [x] Set min date to the earliest paging estimate
- [x] Set date field as invalid when a date the destination library is not open is selected

### DIFF
--- a/app/assets/javascripts/date_selector.js
+++ b/app/assets/javascripts/date_selector.js
@@ -1,0 +1,72 @@
+var dateSelector = (function() {
+  var defaultOptions = {
+    mediatedPageFieldSelector: 'input[data-request-type="mediated_page"]'
+  };
+
+  return {
+    init: function(opts) {
+      var _this = this;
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/date_selector.js
```diff
@@ -0,0 +1,72 @@
...
+
+    setupListeners: function() {
+      var _this = this;
+      if ( _this.mediatedDateField().length > 0 ) {
+        pagingScheduleUpdater
+          .schedulerText()
+          .on('paging-schedule:updated', function(_, data) {
+            _this.setMinDate(data.date);
+            _this.restrictToOpenDates(
+              data.destination_business_days
```
jkeck: This comes from a json response (that we do control...) but seems like this is unimportant.

PR #247, opened 2015-07-24
Add required fields validation and feedback

cbeer:


app/assets/javascripts/item_selector/item_selector_zero_disable_submit.js
```diff
@@ -0,0 +1,48 @@
+var itemSelectorZeroDisableSubmit = (function() {
+
+  return $.extend(itemSelector, {
+    init: function() {
+      var _this = this;
+      $(document).on('ready page:load', function(){
+        if ($('[data-behavior="item-selector"]').length > 0) {
```
cbeer: After adding this check, I'm having a terrible time getting the teaspoon tests to pass (although I don't see why..) @jkeck, would you mind taking a look?

cbeer:
Ok @jkeck, I think I've fixed my testing problem (and, in a future commit, we should be able to refactor the different item selector modules for consistency!)

PR #252, opened 2015-07-30
Allow mediated holdings and ad-hoc-items to be approved

### LINKED ISSUE #231: Mediation functionality
mediation-buttons-before

mediation-buttons-after

* [x] Add comment column to the outer table 
* [x] Make the outer table headers sortable
* [x] Change the open/closed colours to grey so they don't imply status
* [x] Make the detail table width 90% with margin: 0 auto; so it is more clearly a child table
* [x] Remove the "no" column (they don't ever "deny", they just don't approve...)
* [x] Change "Yes" column header to "Approve"
* [x] Add "Status" and "Approved by" columns to detail table
* [x] Add "unlisted items" to the detail table

When the user clicks an Approve button:
* [x] add status text
* [x] add sunet ID of logged-in user, and time stamp
* [x] disable Approve button
* if it's a real item
  * [ ] send it to Darsi
* if it's an "unlisted" item
  * no real action is triggered. The approval action is only to be able to indicate to the patron that the request has been accepted. 

Status text:
* location is -30 = "Paged"
* current loc is -LOAN = "Item is on-site - hold for patron"
* other location = "Added to pick list"
* unlisted = "Approved for manual processing"

### Move past txns out of the table

Requests should be removed from the active table once their "needed on" date has passed, whether they have been approved or not. (Do we need a "history" table?)

jkeck:
Effectively closes #231 (will still need to hook up to symphony when we do that work)

mediated-item-approval

### DIFF
--- a/app/assets/javascripts/item_approval.js
+++ b/app/assets/javascripts/item_approval.js
@@ -1,0 +1,55 @@
+var itemApproval = (function() {
+  var defaultOptions = {
+    buttonSelector: '[data-behavior="item-approval"]',
+    approverInfoSelector: '[data-behavior="approver-information"]'
+  };
+
+  return {
+    init: function(opts) {
+      var _this = this;
+      _this.options = $.extend(defaultOptions, opts);
+      $(document).on('ready page:load', function(){
+        _this.addApprovalBehavior();
+      });
+    },
+
+    options: {},
+
+    addApprovalBehavior: function() {
+      var _this = this;
+      $(document).on('click', _this.options.buttonSelector, function() {
+        var row = $(this).closest('tr');
+        if(!_this.rowIsApproved(row)) {
+          _this.approveItem($(this));
+        }
+      });
+    },
+
+    approveItem: function(item) {
+      var _this = this;
+      var url = item.data('item-approval-url');
+      $.ajax(url).success(function(data) {
+         _this.markRowAsApproved(item);
+         _this.updateApproverInformation(item, data);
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/item_approval.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+
+    markRowAsApproved: function(item) {
+      item.closest('tr').addClass('approved');
+    },
+
+    updateApproverInformation: function(item, data) {
+      var approverInfo = item.closest('tr')
+                             .find(this.options.approverInfoSelector);
+      approverInfo.text(data.approver + ' - ' + data.approval_time);
```
jkeck: This comes from an (internal) json response so I'm not sure this matters.

app/assets/javascripts/item_approval.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+
+    approveItem: function(item) {
+      var _this = this;
+      var url = item.data('item-approval-url');
+      $.ajax(url).success(function(data) {
+         _this.markRowAsApproved(item);
+         _this.updateApproverInformation(item, data);
+       }).fail(function() {
+         alert('The item was not able to be approved.');
```
jkeck: We don't have proper failure handling designed yet.  There will most likely be a message put into the mediation table, so the `alert` is just a placeholder for the moment.

cbeer:
:ok_hand:

PR #268, opened 2015-08-01
Allow mediators to toggle between expired and active requests.

### LINKED ISSUE #256: Remove expired requests from mediation table
Default sort for mediation table should be Needed on.
Requests with Needed on date prior to today's date should not be included in the default view. 

* [x] Display a "Show archived requests" link to show a view of the table with expired requests. 
* When archived requests are shown:
  * [x] add "archived requests" to the title
  * [x] display a "Show current requests" link to return to the default table.
* [x] By the way, the library name should be h2, not h3.

screen shot 2015-07-31 at 12 10 47 pm

screen shot 2015-07-31 at 12 11 22 pm

jkeck:
Closes #256 
expired-mediated-pages

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -7,7 +7,7 @@ class AdminController < ApplicationController
 
   def show
     authorize! :manage, Request.new(origin: params[:id]).library_location
-    @mediated_pages = MediatedPage.where(origin: params[:id]).order(:origin).page(page).per(per)
+    @mediated_pages = mediated_pages
   end
 
   def holdings
@@ -16,6 +16,23 @@ class AdminController < ApplicationController
   end
 
   private
+
+  def mediated_pages
+    pages = if params[:expired]
+              archived_mediated_pages
+            else
+              active_mediated_pages
+            end
+    pages.for_origin(params[:id]).page(page).per(per)
+  end
+
+  def archived_mediated_pages
+    MediatedPage.archived
+  end
+
+  def active_mediated_pages
+    MediatedPage.active
+  end
 
   def page
     params[:page]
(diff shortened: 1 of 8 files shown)


app/models/requests/mediated_page.rb
```diff
@@ -7,6 +7,9 @@ class MediatedPage < Request
   validates :needed_date, presence: true
   validate :destination_is_a_pickup_library
 
+  scope :expired, -> { where('needed_date < ?', Time.zone.today) }
```
cbeer: If we haven't, we should index this field.

jkeck: :+1: added in 6b89a648fa182ba06ce7b5e71a762eddb6cda32c

jkeck:
Just realized the original ticket mixed the terminology archived and expired.  I used expired/active for the scopes but the link has archived/current. Not sure if that is something that should be rectified.

cbeer:
It's probably best to rename them now. `archived`, in my mind, suggests other types of non-current states  besides the needed date is over.

app/controllers/admin_controller.rb
```diff
@@ -17,6 +17,22 @@ def holdings
 
   private
 
+  def mediated_pages
+    if params[:expired]
+      expired_mediated_pages
+    else
+      active_mediated_pages
+    end
```
cbeer: If you're renaming things, I wonder if:
 -  the `.where(origin: params[:id]).order(:origin).page(page).per(per)` below should be moved into `mediated_pages`,
 - `where(origin: params[:id])` is a scope, and if
 - `order(:origin)` is redundant/unnecessary?

jkeck: :+1: I think I have addressed all three of these points.  Let me know if I'm missing something or if there's a better way to handle it.

jkeck:
Renamed the scope to be named `archived` instead of `expired`.

jvine:
Okay. I considered both "expired" and "archived" and chose "archived" because "expired" has a specific meaning for requests in Symphony. These requests are not expired in that sense. I agree that "archived" isn't great either, though. Can we use "past"? Other ideas?

PR #270, opened 2015-08-03
Make item selector use radio buttons instead of checkboxes when there…

### LINKED ISSUE #200: Single-item select not as designed
For scan requests, user needs to be constrained to selecting a single item from a list of volumes. This interaction was designed as a drop-down selector. (Design spec p 24; image at bottom.)

It's been implemented as the multi-item selector with a 1-item limit. This is not a good user experience. 

screen shot 2015-06-29 at 1 30 58 pm

Compromise:
- convert the checkboxes to radio buttons for scan requests
- remove the 1-item limit message, since the radio buttons will constrain the selection 

screen shot 2015-06-29 at 1 21 07 pm

Original spec for reference:

screen shot 2015-06-29 at 1 28 49 pm

jkeck:
… is a limit of 1.

Closes #200

single-select
---
single-select

### DIFF
--- a/app/assets/javascripts/item_selector/item_selector_single_select.js
+++ b/app/assets/javascripts/item_selector/item_selector_single_select.js
@@ -1,0 +1,28 @@
+var itemSelectorSingleSelect = (function() {
+
+  return $.extend({}, itemSelector, {
+    init: function() {
+      var _this = this;
+      $(document).on('ready page:load', function(){
+        _this.addDeselectBehaviorForRadioButtons();
+      });
+    },
+
+    addDeselectBehaviorForRadioButtons: function() {
+      var _this = this;
+      var selections = [];
+      _this.selectorElement()
+           .on('item-selector:selected', function(e, checkbox) {
+        if(checkbox.prop('type') == 'radio') {
+          selections.push(checkbox);
+          if(_this.numberOfSelectedCheckboxes() > 1) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/item_selector/item_selector_single_select.js
```diff
@@ -0,0 +1,28 @@
...
+    addDeselectBehaviorForRadioButtons: function() {
+      var _this = this;
+      var selections = [];
+      _this.selectorElement()
+           .on('item-selector:selected', function(e, checkbox) {
+        if(checkbox.prop('type') == 'radio') {
+          selections.push(checkbox);
+          if(_this.numberOfSelectedCheckboxes() > 1) {
+            _this.selectorElement()
+                 .trigger('item-selector:deselected', [selections[0]]);
```
cbeer: Shouldn't we deselect every selected one, instead of just the first?

jkeck: In the case of the radio button there is only ever 1 selected element (which is the currently selected radio when the change event is called). We need to know which was the previously selected element and pass it to the deselected event (since no change event is triggered on the radios that are being deselected).  Since we call `shift()` below, the first element is popped off of the array right after it is passed to the deselected event and the array will only ever contain 2 elements (and only for an extremely brief time before the first is shifted off).

PR #271, opened 2015-08-05
[WIP] Create workflow for submitting a scan request to symphony

cbeer:


cbeer:
@jkeck can you take a look at this and see if I'm on the right track? If so, I'll start polishing it up with tests and reporting back to the user..

jkeck:
:+1: generally looks good to me.  I don't have a 100% grasp on which callback we should be using in order to notify the user (and if we should be creating rows in the requests table for failed requests or not), but I think you will probably figure out the best approach when you implement reporting back to the user.

jkeck:
I'm holding off on merging this in for the moment.  I'm concerned about deploying this to -dev (even though we could just not configure the endpoint) since we would end up submitting SCAN requests to morison for every other type of request as well.  Generally I agree with this approach and it looks good though.

cbeer:
> we would end up submitting SCAN requests to morison for every other type of request as well

That's not the intended behavior -- it should only run the `SubmitScanRequestJob` on actual scan requests (and, if we want to disable that in -dev, maybe we should add a configuration option)

jkeck:
Ah, I didn't notice the separate `submit!` methods in `Request` and `Scan`.

PR #273, opened 2015-08-07
Fix column layout for use in modal

### LINKED ISSUE #234: Layout of initial page/deliver form in popup
The initial form is showing like this:
screen shot 2015-07-17 at 9 47 23 am

which is okay on a phone but not for all resolutions. 

Should be like this:
screen shot 2015-07-17 at 9 50 33 am

jkeck:
Closes #234 

## Before
modal-selection-before
modal-form-before

## After
modal-selection-after
modal-form-after

### DIFF
--- a/app/assets/stylesheets/variables.scss
+++ b/app/assets/stylesheets/variables.scss
@@ -13,7 +13,9 @@ $headings-font-weight: 300;
 $font-size-base: 14px;
 $line-height-base: 1.428571429;
 $font-family-base: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
-
+// Reducing screen-sm-min to (now depricated) screen-xs-min pixel size in order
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/variables.scss
```diff
@@ -13,7 +13,7 @@ $headings-font-weight: 300;
 $font-size-base: 14px;
 $line-height-base: 1.428571429;
 $font-family-base: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
-
+$screen-sm-min: 480px;
```
cbeer: Would it make sense to add a comment here about why we're changing the default, or is the commit message sufficient?

jkeck: :+1: good call! I added a comment.

PR #276, opened 2015-08-07
Return a 404 when there is no paging estimate found for the given…

### LINKED ISSUE #269: Set fall-back text for paging schedule when no schedule is available.

jkeck:
…locations.

Closes #269 

### Before
business-before

### After
business-after

### DIFF
--- a/app/controllers/paging_schedule_controller.rb
+++ b/app/controllers/paging_schedule_controller.rb
@@ -4,9 +4,12 @@
 class PagingScheduleController < ApplicationController
   layout false
 
+  rescue_from PagingSchedule::ScheduleNotFound do
+    render status: 404
+  end
+
   def show
     schedule = PagingSchedule.for(request_for_schedule)
-    return unless schedule.present?
     respond_to do |format|
       format.json { render json: schedule.estimate }
       format.html { render text: schedule.estimate.to_s }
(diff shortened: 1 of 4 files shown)


app/controllers/paging_schedule_controller.rb
```diff
@@ -6,10 +6,13 @@ class PagingScheduleController < ApplicationController
 
   def show
     schedule = PagingSchedule.for(request_for_schedule)
```
cbeer: I'm happy with this, but is not having a paging schedule exceptional (and we could rescue the exception with a 404..)?

jkeck: :+1: see below

app/controllers/paging_schedule_controller.rb
```diff
@@ -6,11 +6,12 @@ class PagingScheduleController < ApplicationController
...
-    return unless schedule.present?
     respond_to do |format|
       format.json { render json: schedule.estimate }
       format.html { render text: schedule.estimate.to_s }
     end
+  rescue PagingSchedule::ScheduleNotFound
```
cbeer: At the controller level using `rescue_from`?

jkeck: :ok_hand:

cbeer:
:+1:

PR #277, opened 2015-08-07
178 illiad openurl

jgreben:
Created an IlliadOpenurl class. Added illiad_url method to scans controller. Updated the corresponding scan_controller_spec and create_scan_spec. Tested round trip to sulils-test and back.

screen shot 2015-08-07 at 4 51 35 pm

screen shot 2015-08-07 at 4 52 10 pm


app/controllers/scans_controller.rb
```diff
@@ -27,20 +27,11 @@ def new_request_params
   end
 
   def redirect_to_success_with_token
-    redirect_to illiad_query(current_request)
+    redirect_to illiad_url
```
cbeer: This may be a stupid question, but why are we bouncing the user through illiad? Couldn't we handle this request on the back-end, like we do for submitting requests to Symphony?

app/controllers/scans_controller.rb
```diff
@@ -27,20 +27,11 @@ def new_request_params
...
-      'rft.jtitle': scan.item_title,
-      'rft.au': scan.data[:authors],
-      'rft.pages': scan.data[:page_range],
-      'rft.atitle': scan.data[:section_title],
-      'rft.volume': scan.holdings.first.callnumber,
-      'scan_referrer': successful_scan_url(scan)
-    }
-    Settings.sul_illiad + "#{illiad_params.to_query}"
+  def illiad_url
+    @illiad_url ||= IlliadOpenurl.new(current_request, successful_scan_url(current_request)).to_url
```
cbeer: One way to make this method more testable would be to pass the `current_request` in as a parameter, like we were doing in `#illiad_query`.

It doesn't look like we call `#illiad_url` more than once in a request, so the memoization may be unnecessary too.

app/models/illiad_openurl.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def illiad_item_data
+    { 'rft.location': @scan.origin,
+      'rft.call_number': @scan.holdings.first.callnumber,
+      'rft.item': @scan.holdings.first.barcode }
+  end
+
+  def illiad_citation
+    { 'rft.jtitle': @scan.item_title,
+      'rft.au': @scan.data[:authors],
```
cbeer: I think `@scan.authors` should work here, no?

app/models/illiad_openurl.rb
```diff
@@ -0,0 +1,35 @@
...
+    { 'Action': '10', 'Form': '30', 'rft.genre': 'scananddeliver' }
+  end
+
+  def illiad_item_data
+    { 'rft.location': @scan.origin,
+      'rft.call_number': @scan.holdings.first.callnumber,
+      'rft.item': @scan.holdings.first.barcode }
+  end
+
+  def illiad_citation
```
cbeer: These are generic openurl parameters, not anything specific to openurl, right? Would it make sense to promote them to a module that gets mixed into `Request`?

app/models/illiad_openurl.rb
```diff
@@ -0,0 +1,35 @@
...
+    @redirect_url = redirect_url
+  end
+
+  def illiad_page
+    { 'Action': '10', 'Form': '30', 'rft.genre': 'scananddeliver' }
+  end
+
+  def illiad_item_data
+    { 'rft.location': @scan.origin,
+      'rft.call_number': @scan.holdings.first.callnumber,
```
cbeer: Do we need to make sure the request has a holding (although maybe items without holdings aren't valid for scan + deliver in the first place?)

app/models/illiad_openurl.rb
```diff
@@ -0,0 +1,35 @@
...
+      'rft.item': @scan.holdings.first.barcode }
```
cbeer: Should this actually be `@scan.barcodes.first`? I think that's the barcoded item the user selected on the form (@jkeck, is that right?)

app/models/illiad_openurl.rb
```diff
@@ -0,0 +1,35 @@
...
+      'rft.pages': @scan.data[:page_range],
+      'rft.atitle': @scan.data[:section_title] }
+  end
+
+  def illiad_callback
+    { 'scan_referrer': @redirect_url }
+  end
+
+  def to_url
+    illiad_qy = [illiad_page, illiad_item_data, illiad_citation, illiad_callback].reduce({}, :merge)
```
cbeer: `illiad_qy` looks like it should be its own method (`query_params` or something)?

app/models/illiad_openurl.rb
```diff
@@ -0,0 +1,35 @@
+###
+#  Class to handle creation of ILLiad OpenURL request
+###
+class IlliadOpenurl
+  def initialize(scan, redirect_url)
+    @scan = scan
+    @redirect_url = redirect_url
+  end
+
+  def illiad_page
```
cbeer: I think you can safely drop the `illiad_` prefix from all these methods, since they are in an `Illiad` class anyway.

app/controllers/scans_controller.rb
jgreben: Ideally it would be better to do it on the back-end, but ILLilad does not have a supported API for handling request creation. If we wanted to do it on the back-end, we would have to insert rows directly into the MSSQL tables without intimately knowing the schemas and relationships between tables. It might be worth looking into this angle, but it would take a lot of investigation and testing before we could go live.

cbeer: I guess I don't know what happens at that URL. Does the user need to do something special (and, is it something we could script) once they get to Illiad?

jgreben: The user will not have to do anything special if all goes correctly. The rft.genre=scananddeliver parameter instructs illiad to use a specific set of html pages. On those pages exist javascript to automatically submit the form (hence modeling the request in illiad) and upon success take them back to the callback (successful_scan) url.

app/models/illiad_openurl.rb
jkeck: Yes, `Scan#barcodes` is the array of selected barcodes (I think that plumbing may have not been there when this code was first introduced).

However; in this context, a request object where holdings have been selected, they are effectively the same. Holdings will be limited to only the user selected holdings, and since a scan can only have 1 item requested, it will be the user's selection. https://github.com/sul-dlss/sul-requests/blob/master/app/models/concerns/holdings.rb

This approach may end up causing issues down the road if we ever decide to expand the number of items a user may select for a scan since this is hard-coded to the first.

jkeck: *sidenote:* We are (or were) DRY-ing up a bit of code in dealing with iterating through holdings when the item was persisted, but perhaps we should break out all holdings (for item selector purposes) and selected holdings (for success/status/email/mediation purposes) into separate methods.  I may have been trying to be too clever here.

app/controllers/scans_controller.rb
cbeer: If we don't need the user to do anything, why can't we automate this step? Is there something about the user credentials? Is there configuration we can do to allow the app to send proxy requests on their behalf? I know we've done some clever Apache configuration before to do this, and I'd guess we could do something similar here.

app/models/illiad_openurl.rb
cbeer: Ah, cool. Never mind.

app/controllers/scans_controller.rb
jgreben: The user's sunset id does get passed along in the request header to match the user to their illiad account. The illiad.dll file is the 'controller' for that. As long as we can still mimic that on the back end I suppose it could be possible.

jkeck:
:+1: @jgreben I think we can ship this. Would you mind squashing the commits?

PR #280, opened 2015-08-11
Add mouseflow js to modal layout under development.

jkeck:


jkeck:
:clap: thanks @jgreben

PR #285, opened 2015-08-13
Re-styled broadcast messages

jgreben:
Re-styled broadcast messages view to include buttonized links, alert-style message box, styled active/inactive notes, and the button link moved next to the heading.

Ps. the branch name should have been "206-..." not "260-..."

### DIFF
--- a/app/assets/stylesheets/modules/messages.scss
+++ b/app/assets/stylesheets/modules/messages.scss
@@ -23,3 +23,33 @@
     }
   }
 }
+
+.messages {
+  margin-left: 2em;
+
+  h2 {
+    color: $black;
+    float: left;
+    font-size: 1.2em;
+    font-weight: normal;
+  }
+
+  .message {
+    clear: both;
+  }
+
+  .status {
+    color: $cardinal-red;
+  }
+
+  .active {
+    color: $green;
+    font-weight: bold;
+  }
+
+  .edit-message {
+    left: 14px;
+    position: relative;
+    top: 14px;
+  }
+}
--- a/app/views/messages/_messages_for_library.html.erb
+++ b/app/views/messages/_messages_for_library.html.erb
@@ -1,24 +1,48 @@
 <div class="messages">
-<% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
-<% messages.each do |message| %>
-  <div class="message">
-      <% if message.scheduled? %>
-      <div class="status <%= 'active' if message.active? %>">
-        <% if message.active? %>
-          Active <%= time_tag message.start_at.to_date, :short %> through <%= time_tag message.end_at.to_date, :short %>
-        <% else %>
-          Inactive (was <%= time_tag message.start_at.to_date, :short %> through <%= time_tag message.end_at.to_date, :short %>)
-        <% end %>
-      </div>
-      <% end %>
 
-      <div class="text well-sm">
+  <% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
+
+  <% if library_code == 'SAL' %>
+    <% library_code = "SAL1&2" %>
+  <% elsif library_code == "SAL-NEWARK"%>
+    <% library_code = "SAL Newark" %>
+  <% end %>
+
+  <h2><%= request_type.titleize + " from " + library_code if messages.none? %></h2>
+  <%= link_to " Add message", new_message_path(library: library_code, request_type: request_type), class: 'btn btn-sm btn-default edit-message glyphicon glyphicon-pencil' if messages.none? %>
+
+  <% messages.each do |message| %>
+  <h2><%= request_type.titleize + " from " + library_code %></h2>
+  <% if message.text.length > 0 %>
+    <%= link_to [:edit, message], class: 'btn btn-sm btn-default edit-message' do %>
+      <i class="glyphicon glyphicon-pencil"> Edit message</i>
+    <% end %>
+  <% else %>
+    <%= link_to new_message_path(library: library_code, request_type: request_type), class: 'btn btn-sm btn-default edit-message' do %>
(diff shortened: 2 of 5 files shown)


app/views/messages/_messages_for_library.html.erb
```diff
@@ -1,24 +1,48 @@
 <div class="messages">
-<% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
-<% messages.each do |message| %>
+
+  <% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
+
+  <% if library_code == 'SAL' %>
+    <% library_code = "SAL1&2" %>
```
jkeck: What is the purpose of reassigning this variable in the view?

app/views/messages/_messages_for_library.html.erb
```diff
@@ -1,24 +1,48 @@
...
+  <% elsif library_code == "SAL-NEWARK"%>
+    <% library_code = "SAL Newark" %>
+  <% end %>
+
+  <h2><%= request_type.titleize + " from " + library_code if messages.none? %></h2>
+  <%= link_to " Add message", new_message_path(library: library_code, request_type: request_type), class: 'btn btn-sm btn-default edit-message glyphicon glyphicon-pencil' if messages.none? %>
```
jkeck: It seems like this `link_to` is identical to the one below.  Maybe these should be in a helper so we don't violate DRY.

app/views/messages/_messages_for_library.html.erb
jgreben: Well, the spec looks for SAL1&2 but since I am dynamically generating the h2's out of the library_code and request_type it doesn't quite match what's in the spec (and presumably what we want the user to see).

app/views/messages/_messages_for_library.html.erb
jkeck: Both of these are only done if `messages.none?` evaluates to true.  Should we wrap both those lines in the conditional instead?  (I notice that the conditional is in the `h1 in the line above, this means that there can be an empty `h2` on the page, which I don't think is desired)

app/views/messages/_messages_for_library.html.erb
```diff
@@ -1,24 +1,48 @@
...
+
+  <% messages.each do |message| %>
+  <h2><%= request_type.titleize + " from " + library_code %></h2>
+  <% if message.text.length > 0 %>
+    <%= link_to [:edit, message], class: 'btn btn-sm btn-default edit-message' do %>
+      <i class="glyphicon glyphicon-pencil"> Edit message</i>
+    <% end %>
+  <% else %>
+    <%= link_to new_message_path(library: library_code, request_type: request_type), class: 'btn btn-sm btn-default edit-message' do %>
+      <i class="glyphicon glyphicon-pencil"> Add message</i>
```
jkeck: Move the text out of the `<i>` tag you won't get the weird spacing. This will be a bit more complicated with the `link_to` on line 12 but that's one of the reasons it's better to be consistent between them.

app/assets/stylesheets/modules/messages.scss
```diff
@@ -23,3 +23,33 @@
...
+
+  .active {
+    color: $green;
+    font-weight: bold;
+  }
+
+  .edit-message {
+    left: 14px;
+    position: relative;
+    top: 14px;
```
jkeck: This and the floating of the `h2` feels weird to me.  Can we make the `<h2>` `display: inline` instead of floating (which allows us to removing this left/top positioning) and then maybe give one of the surrounding divs a little `bottom-margin` to achieve the spacing we want. Experiment with it and let me know what you think.

app/views/messages/_messages_for_library.html.erb
jkeck: What about this: https://github.com/sul-dlss/sul-requests/blob/master/app/models/library_location.rb#L41-L43

Which you can see an example of in the `Messages` class here https://github.com/sul-dlss/sul-requests/blob/d9abd56e7a508fc98682c11b1cb9c9f3956e68f5/app/models/message.rb#L15-L17

(which it appears that you've actually duplicated that code a few times in the view, which I'll add comments about separately)

app/views/messages/_messages_for_library.html.erb
```diff
@@ -1,24 +1,48 @@
...
+    <% library_code = "SAL1&2" %>
+  <% elsif library_code == "SAL-NEWARK"%>
+    <% library_code = "SAL Newark" %>
+  <% end %>
+
+  <h2><%= request_type.titleize + " from " + library_code if messages.none? %></h2>
+  <%= link_to " Add message", new_message_path(library: library_code, request_type: request_type), class: 'btn btn-sm btn-default edit-message glyphicon glyphicon-pencil' if messages.none? %>
+
+  <% messages.each do |message| %>
+  <h2><%= request_type.titleize + " from " + library_code %></h2>
```
jkeck: `Message.title?` 

https://github.com/sul-dlss/sul-requests/blob/d9abd56e7a508fc98682c11b1cb9c9f3956e68f5/app/models/message.rb#L15-L17

app/views/messages/_messages_for_library.html.erb
```diff
@@ -1,24 +1,48 @@
...
+
+  <% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
+
+  <% if library_code == 'SAL' %>
+    <% library_code = "SAL1&2" %>
+  <% elsif library_code == "SAL-NEWARK"%>
+    <% library_code = "SAL Newark" %>
+  <% end %>
+
+  <h2><%= request_type.titleize + " from " + library_code if messages.none? %></h2>
```
jkeck: So this exact string seems to come up about 3 different places in the codebase now.  Perhaps that's an indication that it should be put into the internationalization file (en.yml) and referenced that way.

PR #288, opened 2015-08-13
Re-styled broadcast messages

jgreben:
Re-styled broadcast messages view including re-styled headings, buttonized links next to the heading, and alert-styled message box.

Renamed branch to match issue 206 and committed some changes that @jkeck suggested in closed PR #285.

### DIFF
--- a/app/views/messages/_messages_for_library.html.erb
+++ b/app/views/messages/_messages_for_library.html.erb
@@ -1,24 +1,41 @@
 <div class="messages">
-<% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
-<% messages.each do |message| %>
-  <div class="message">
+
+  <% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
+  <% library = LibraryLocation.library_name_by_code(library_code).to_s %>
+
+  <% if messages.none? %>
+    <h2><%= request_type.titleize + " from #{library.presence || 'anywhere'}" %></h2>
+    <%= link_to_add_message(library_code, request_type) %>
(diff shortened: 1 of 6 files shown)


jgreben:
@jkeck: I believe that I addressed most if not all of the issues from closed PR #285. I DRY'd up the code, constructed the titles in a hopefully better way, and re-styled the H2's. When you get a chance can you take a look at this PR and let me know if there is anything I missed?

app/views/messages/_messages_for_library.html.erb
```diff
@@ -1,24 +1,45 @@
...
+
+  <% messages = @messages.select { |m| m.library == library_code && m.request_type == request_type } %>
+  <% @library = LibraryLocation.library_name_by_code(library_code).to_s %>
+
+  <% if @library.empty? %>
+    <% @library = 'anywhere' %>
+  <% end %>
+
+  <% if messages.none? %>
+    <h2><%= request_type.titleize + " from " + @library %></h2>
```
jkeck: The best I can tell, this is the only place where the `@library` instance variable is used (I don't think this should be an instance variable w/ the `@`).  Can we just change this to be `<%= request_type.titleize + " from #{library || 'anywhere'}" %>`.  Then we could remove lines 6-8 entirely.

jkeck:
Just one minor comment and I think we're ready to :shipit:

app/views/messages/_messages_for_library.html.erb
jgreben: This does not appear to return 'anywhere' if there is no library returned.

PR #300, opened 2015-08-14
297 css tweaks

jgreben:
I tried to put the styles into existing classes and put the new ones into the most logical file. Let me know if you had something else in mind and if it otherwise looks the way you wanted

### DIFF
--- a/app/assets/stylesheets/modules/forms.scss
+++ b/app/assets/stylesheets/modules/forms.scss
@@ -1,8 +1,4 @@
 .form-group {
-  .input-like-text {
-    padding-top: 7px;
-  }
-
   hr {
     border: 0;
     border-top: 1px solid $grayish-yellow;
@@ -10,8 +6,13 @@
     margin-top: 20px;
   }
 
+  .input-like-text {
+    padding-top: 7px;
+  }
+
   .or-divider {
     background-color: $body-bg;
+    color: $dark-gray;
     font-size: 1.1em;
     margin: -32px auto 10px;
     text-align: center;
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/modules/item-selector.scss
```diff
@@ -13,6 +13,8 @@
     i {
       @extend .glyphicon-tiny;
       @extend .pull-right;
+      float: none !important;
```
jkeck: The float comes from extending the `.pull-right` class.  If you want to remove the float remove the `@extend` line instead.

app/assets/stylesheets/modules/forms.scss
```diff
@@ -36,14 +37,35 @@
...
+  margin: 0 0 0.5em;
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
 }
 
+.required:after {
+  content: 'Required';
+  margin-left: 5px;
+}
```
jkeck: `.required:after` is already defined above https://github.com/sul-dlss/sul-requests/pull/300/files#diff-22681f8e52bfcda88e3b3a6e9a7982a0L21

Gemfile.lock
```diff
@@ -341,6 +341,3 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   web-console (~> 2.0)
-
-BUNDLED WITH
-   1.10.4
```
jkeck: Hey @jgreben would you mind updating your bundler gem so that this doesn't get removed in your commits?

Check out [this comment](https://github.com/bundler/bundler/issues/3697#issuecomment-108299463) from one of the bundler devs about why this is happening.

PR #301, opened 2015-08-14
Always put comments on SAL-NEWARK pages.

### LINKED ISSUE #263: SAL-Newark items not all getting comments box
It looks like SAL-Newark items are getting a comments box only if they don't have a legit barcode.

Single item, real barcode, but no comment box:  2833045, 356446
Multiple barcoded items, no comment box: 5086259
No real barcode, gets comment + holdings:  4279331, 4472091

All SAL-Newark items should have a comments box plus holdings (if they exist), in addition to the item selector (if there are multiple barcoded items).

jkeck:
Closes #263 

`/requests/new?item_id=2833045&origin=SAL-NEWARK&origin_location=PAGE-GR`

## Before
2833045-before

## After
2833045-after

### DIFF
--- a/app/models/requests/page.rb
+++ b/app/models/requests/page.rb
@@ -16,11 +16,6 @@ class Page < Request
     true
   end
 
-  def item_commentable?
-    return super unless origin == 'SAL-NEWARK'
-    !barcoded_holdings?
-  end
-
   private
 
   def page_validator
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -8,9 +8,9 @@ module Commentable
   end
 
   # Currently used to comment which items you would like if the
-  # location you're requesting from does not have any barcoded items
+  # location you're requesting from is configured as item commentable
   def item_commentable?
-    false
+    SULRequests::Application.config.item_commentable_libraries.include?(origin)
   end
 
   def request_commentable?
--- a/app/models/concerns/holdings.rb
+++ b/app/models/concerns/holdings.rb
@@ -11,10 +11,6 @@ module Holdings
     holdings_object.all
   end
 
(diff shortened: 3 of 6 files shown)


app/models/requests/page.rb
```diff
@@ -17,8 +17,7 @@ def requestable_by_all?
   end
 
   def item_commentable?
-    return super unless origin == 'SAL-NEWARK'
-    !barcoded_holdings?
+    origin == 'SAL-NEWARK'
```
cbeer: Should this be `origin == 'SAL-NEWARK' || super`? (also, what about making SAL-NEWARK a config parameter?)

app/models/requests/page.rb
```diff
@@ -17,8 +17,7 @@ def requestable_by_all?
   end
 
   def item_commentable?
-    return super unless origin == 'SAL-NEWARK'
-    !barcoded_holdings?
+    SULRequests::Application.config.item_commentable_libraries.include?(origin) || super
```
jkeck: I wonder if this should actually be at the request class level (also potentially some other of these library specific behaviors).  The reason is, if a SAL-NEWARK item were to get a hold/recall for some reason I **think** we would want this behavior to apply there as well.

cc: @jvine in case she has any insight (and about other things that are potentially consistent across all request types based on library level attributes).

cbeer:
:+1:

jkeck:
Pushed that logic up to the request level (even though we don't think it's possible for a SAL-NEWARK request to get a Hold/Recall).  Leaving the others for now, but will be simple to move up in the future.

cbeer:
:+1: :shipit: :question:

jkeck:
:shipit::exclamation:

PR #304, opened 2015-08-15
[PROPOSAL] Style the destination field to not have the required label…

### LINKED ISSUE #284: "Deliver to" shouldn't be "required"
Note that "Deliver to" was not listed in #228 - it doesn't need to be marked required as it's impossible to leave it empty. 

screen shot 2015-08-13 at 11 34 30 am

jkeck:
… (since it can't be blank).

This is just a proposal, I'm not sure if there's a better way to handle this.  The `required` label is there because we validate the presence of the destination in the model level.

Closes #284 

destination-no-require

### DIFF
--- a/app/assets/stylesheets/modules/forms.scss
+++ b/app/assets/stylesheets/modules/forms.scss
@@ -18,7 +18,7 @@
     width: 3em;
   }
 
-  .required {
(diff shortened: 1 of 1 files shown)


cbeer:
:+1:

PR #307, opened 2015-08-19
Refactor i18n header keys and helpers to allow for variable mediated …

### LINKED ISSUE #262: Title for mediated pages is not always correct
It looks like all mediated pages are getting "Request on-site access". This is appropriate for SPEC-COLL, HOOVER and HV-ARCHIVE, but not for HOPKINS and PAGE-MP, where the items are circulating items, and the title should be "Request delivery to campus library".

example - 2833307
screen shot 2015-07-31 at 4 19 48 pm

jkeck:
…page headers.

Closes #262 

## PAGE-MP
`/requests/new?item_id=3161594&origin=SAL3&origin_location=PAGE-MP`
PAGE-MP

## SPEC-COLL
`/requests/new?item_id=9161931&origin=SPEC-COLL&origin_location=U-ARCHIVES`
SPEC-COLL

### DIFF
--- a/app/views/hold_recalls/_header.html.erb
+++ b/app/views/hold_recalls/_header.html.erb
@@ -1,3 +1,3 @@
 <h1 id='dialogTitle'>
-  <%= t("headers.hold_recall.#{i18n_title_key_for_hold_recall}", default: t('headers.hold_recall.default')) %>
+  <%= t(:".#{i18n_library_title_key}", default: [:".#{i18n_location_title_key}", :'.default']) %>
 </h1>
--- a/app/helpers/requests_helper.rb
+++ b/app/helpers/requests_helper.rb
@@ -39,14 +39,16 @@ module RequestsHelper
     end
   end
 
-  def i18n_title_key_for_hold_recall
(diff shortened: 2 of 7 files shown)


app/views/hold_recalls/_header.html.erb
```diff
@@ -1,3 +1,3 @@
 <h1 id='dialogTitle'>
-  <%= t("headers.hold_recall.#{i18n_title_key_for_hold_recall}", default: t('headers.hold_recall.default')) %>
+  <%= t(".#{i18n_library_title_key}", default: t(".#{i18n_location_title_key}", default: t('.default'))) %>
```
cbeer: You can write these fallbacks more succinctly as:

```
t(:".#{i18n_library_title_key}", default: [:".#{i18n_location_title_key}", :'.default'])) 
```

(being careful to use symbols -- if you use strings, they'll be treated as the string literal, not the i18n key)

jkeck: :+1: I knew there had to be a better way.  Thanks for pointing that out.

PR #308, opened 2015-08-19
Don't require date field for HOPKINS or PAGE-MP.

### LINKED ISSUE #266: PAGE-MP and HOPKINS mediated pages should not have "planned date of use"
Kind of related to #262. 

PAGE-MP and HOPKINS are not "plan in advance to visit the library" - they are just page requests with an in-between step. They should not have a "Planned date of use" field. When they are also a hold/recall, they should have a "Cancel request after" field.

screen shot 2015-07-31 at 5 18 49 pm
screen shot 2015-07-31 at 5 22 45 pm

jkeck:
Closes #266

### DIFF
--- a/app/models/requests/mediated_page.rb
+++ b/app/models/requests/mediated_page.rb
@@ -42,6 +42,8 @@ class MediatedPage < Request
   end
 
   def requires_needed_date?
+    return false if origin == 'HOPKINS'
+    return false if origin_location == 'PAGE-MP'
     true
   end
 
--- a/spec/models/requests/mediated_page_spec.rb
+++ b/spec/models/requests/mediated_page_spec.rb
@@ -109,4 +109,20 @@ describe MediatedPage do
       expect(subject.item_limit).to eq 5
     end
   end
+
+  describe '#requires_needed_date?' do
+    it 'is false when the library is HOPKINS' do
+      subject.origin = 'HOPKINS'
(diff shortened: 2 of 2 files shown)


app/models/requests/mediated_page.rb
```diff
@@ -42,6 +42,8 @@ def item_limit
   end
 
   def requires_needed_date?
+    return false if origin == 'HOPKINS'
+    return false if origin_location == 'PAGE-MP'
```
jkeck: I think this is another indication of a sub-type that exists somewhere (similar to https://github.com/sul-dlss/sul-requests/blob/24bcfcae4cae13fbb7b5f77f6da410cf9fc82415/config/locales/en.yml#L88-L89)  :disappointed:

cbeer: Should we add a configuration for these too? It might make the code a little more clear, if nothing else.

jkeck: I thought the same thing at first but didn't really know what that config should be.  We're kind of blacklisting a library and location here so do you think something like `date_needed_library_blacklist` and `date_needed_location_blacklist` would be appropriate?

cbeer: Ah, I didn't catch that. Yeah, ew. I guess ship it and worry about it later.

PR #310, opened 2015-08-19
Added aria-describedby page_range_help_block to page_range form field

jgreben:
:ok_hand:?

### DIFF
--- a/app/views/scans/_form_additional.html.erb
+++ b/app/views/scans/_form_additional.html.erb
@@ -1,4 +1,10 @@
-<%= f.text_field :page_range, help: 'Example: 1-5, 8, 7-13' %>
+<div class="form-group">
+  <%= f.label :page_range, class: "#{label_column_class} control-label" %>
+  <div class='<%= content_column_class %>'>
+    <%= f.text_field_without_bootstrap :page_range, class: 'form-control', aria: { describedby: 'page_range_help_block' } %>
+    <span class="help-block" id="page_range_help_block">Example: 1-5, 8, 7-13</span>
+  </div>
+</div>
 <%= f.text_area :section_title, rows: '3' %>
 <%= f.text_field :authors %>
 


app/views/scans/_form_additional.html.erb
```diff
@@ -1,5 +1,11 @@
-<%= f.text_field :page_range, help: 'Example: 1-5, 8, 7-13' %>
+<div class="form-group">
+  <%= f.label :page_range, class: 'control-label col-sm-3' %>
```
jkeck: We have a helper that has the label and content column sizes so we can change them all in one place if need be: https://github.com/sul-dlss/sul-requests/blob/master/app/helpers/application_helper.rb#L19-L21

app/views/scans/_form_additional.html.erb
```diff
@@ -1,5 +1,11 @@
...
+  <div class="col-sm-9">
```
jkeck: Sam as above, check out https://github.com/sul-dlss/sul-requests/blob/master/app/helpers/application_helper.rb#L27-L29

app/views/scans/_form_additional.html.erb
```diff
@@ -1,5 +1,11 @@
...
+    <%= f.text_field_without_bootstrap :page_range, class: 'form-control', 'aria-describedby' => 'page_range_help_block' %>
```
jkeck: The aria bit could use the new hash syntax by using `aria: { describedby: 'page_range_help_block' }`

app/views/scans/_form_additional.html.erb
```diff
@@ -1,5 +1,11 @@
-<%= f.text_field :page_range, help: 'Example: 1-5, 8, 7-13' %>
+<div class="form-group">
+  <%= f.label :page_range, class: label_column_class + ' control-label ' %>
```
jkeck: Not to be pedantic, but could you use string interpolation here instead of concatenation?  Check out this example: https://repl.it/BCec

If you try to concatenate a nil variable and a string you'll get an error.  However; using interpolation `to_s` will be called on the nil variable and an empty string will be returned. `nil.to_s => ""`

jgreben: I did try this:

`class: '#{label_column_class} control-label’`

But it was rendered literally:

`<label class="#{label_column_class} control-label" for="request_page_range">Page range</label>`


On Aug 19, 2015, at 3:13 PM, Jessie Keck <notifications@github.com<mailto:notifications@github.com>> wrote:


In app/views/scans/_form_additional.html.erb<https://github.com/sul-dlss/sul-requests/pull/310#discussion_r37475444>:

> @@ -1,5 +1,11 @@
> -<%= f.text_field :page_range, help: 'Example: 1-5, 8, 7-13' %>
> +<div class="form-group">
> +  <%= f.label :page_range, class: label_column_class + ' control-label ' %>


Not to be pedantic, but could you use string interpolation here instead of concatenation? Check out this example: https://repl.it/BCec

If you try to concatenate a nil variable and a string you'll get an error. However; using interpolation to_s will be called on the nil variable and an empty string will be returned. nil.to_s => ""

jkeck: You need to use double quotes to take advantage of string interpolation.

So `"#{label_column_class} control-label"` instead of `'#{label_column_class} control-label'`.

PR #314, opened 2015-08-21
Pass modal param along w/ the illiad redirect URL.

jkeck:


app/controllers/scans_controller.rb
```diff
@@ -31,7 +31,7 @@ def redirect_to_success_with_token
   end
 
   def illiad_url
-    IlliadOpenurl.new(current_request, successful_scan_url(current_request)).to_url
+    IlliadOpenurl.new(current_request, successful_scan_url(current_request, modal: params[:modal])).to_url
```
cbeer: Happy to :shipit:, but i wonder if we should pull `{ modal: params[:modal] }` into a method ("request_context_params" or something..) so it's clear why we're doing this.

jkeck: :+1: good call.  Let me know if how I implemented that is correct.

cbeer:
:+1:

PR #317, opened 2015-08-21
Update scan form to include scan paging schedule.

### LINKED ISSUE #261: No estimated date for scans
Scan form is meant to have an earliest delivery date. 

Initial intent was that it should be the "delivery to Green" time plus one increment - so, if delivery would happen in the morning, scanning would be that afternoon. If delivery would happen in the afternoon, scanning would be morning the following day. 

I don't know if this is affected by staff time - i.e. whether weekends are "closed" for scanning or not. Will verify with Sarah S.

screen shot 2015-07-31 at 4 16 08 pm

jkeck:
Closes #261 

scan-schedule

### DIFF
--- a/app/models/library_hours.rb
+++ b/app/models/library_hours.rb
@@ -1,9 +1,8 @@
 # LibraryHours is responsbile for determining
 # if a library is open on a given day
 class LibraryHours
-  attr_reader :library
-  def initialize(library)
-    @library = library
+  def initialize(library_code)
+    @library = library_code
   end
 
   def next_business_day(date = Time.zone.today)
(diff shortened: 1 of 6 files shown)


app/models/library_hours.rb
```diff
@@ -19,6 +18,10 @@ def business_days
     end.compact
   end
 
+  def library
+    config.scanning_library_proxy[@library] || @library
```
cbeer: I'm not thrilled by this. I wonder if we could teach the `config.hours_api_location_map` hash how to alias the hours for us..

PR #355, opened 2015-09-01
Add a default min date of today on the date field and validate server…

### LINKED ISSUE #330: Not trapping dates in the past?
User entered 02/09/15 in the needed on date (in Firefox where there's no date control) intending September 2. Two things happened:

1. the app interpreted 15 as 1915 and trapped it as a closed day, which ... at least it trapped it?
2. once the user changed the year to 2015, the app interpreted 02/09 as Feb 9, and allowed the request to be created.

Needed on dates should not ever be before today's date.

jkeck:
… side.
Closes #330

### DIFF
--- a/app/models/concerns/request_validations.rb
+++ b/app/models/concerns/request_validations.rb
@@ -1,0 +1,32 @@
+###
+# Module to group request object validations together
+###
+module RequestValidations
+  extend ActiveSupport::Concern
+
+  included do
+    validates :item_id, :origin, :origin_location, presence: true
+    validate :requested_holdings_exist
+    validate :needed_date_is_not_in_the_past, on: :create
+  end
+
+  protected
+
+  def destination_is_a_pickup_library
+    return if library_location.pickup_libraries.include?(destination)
+    errors.add(:destination, 'is not a valid pickup library')
+  end
+
+  # This will currently stil pass if the request has no barcodes.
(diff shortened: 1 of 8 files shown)


app/models/concerns/request_validations.rb
```diff
@@ -0,0 +1,30 @@
...
+  # I'm not sure we strongly enforce WHEN requests require barcodes
+  # (it seems like it may be variable depending on the record).
+  def requested_holdings_exist
+    holdings_barcodes = holdings.map(&:barcode)
+    return if barcodes.all? { |b| holdings_barcodes.include?(b) }
+    errors.add(:base, 'A selected item is not located in the requested location')
+  end
+
+  def needed_date_is_not_in_the_past
+    errors.add(:base, 'Date cannot be earlier than today') if needed_date && needed_date < Time.zone.today
```
cbeer: Why not:

```
errors.add(:needed_date, 'cannot be earlier than today')
```

?

cbeer: I also wonder if it's worth pulling in https://github.com/codegram/date_validator to help us.. maybe it's overkill?

jkeck: The label that's provided to the user is either "Not needed after" or "Planned date of use".  Date seemed generic enough to handle them both.

Perhaps we can leverage i18n here somehow here so we don't show the users "Needed date cannot be earlier that today" in all cases?

PR #356, opened 2015-09-01
Submit all types of requests to Symphony

cbeer:


app/models/searchworks_item.rb
```diff
@@ -163,6 +163,7 @@ def approve!(user)
           approval_time: Time.zone.now.to_s,
           approver: user
         }
+        @request.send_to_symphony!(barcodes: [@id])
```
jkeck: Do you think we should have a test for this? Since it's going through the request object I'm not sure what to even test so ¯\_(ツ)_/¯

jkeck:
One super minor comment, other than that I'm ready to :shipit:.

app/models/searchworks_item.rb
cbeer: :+1: good call. I'll add a unit test for it.

Issue #293, opened 2015-08-14
Add shim for HTML5 date field


jkeck:
Webshims seems to be do-able: http://afarkas.github.io/webshim/demos/index.html http://jsfiddle.net/trixta/BMEc9/

Are there other shims worth evaluating for accessibility?
https://github.com/chemerisuk/better-dateinput-polyfill
http://www.useragentman.com/blog/2010/07/27/creating-cross-browser-html5-forms-now-using-modernizr-webforms2-and-html5widgets-2/

jgreben:
Using: https://github.com/whatcould/webshims-rails
Run: 
bundle install
rake webshims:update_public_versioned

PR #370, opened 2015-09-02
Added webshim for html5 date

### LINKED ISSUE #336: HTML5 date field usability issues
The HTML5 date field is technically accessible, but not very usable. Many don't know that there is a calendar popup available, and the field template can be unforgiving. Wish I had video to show this.

Date selection was one of the primary usability issues we wanted to address with the new form. We've done a lot toward this by preventing selection of invalid dates, but people expect a calendar, and they are not seeing it. 

It looks like the webshim has the option to "openOnFocus", which would help a lot. Can we override Chrome's native behaviour with the webshim?

jgreben:
Fixes #336 #293

Using CDN version: http://cdn.jsdelivr.net/webshim/1.15.8/polyfiller.js

### DIFF
--- a/app/assets/javascripts/html5_webshim.js
+++ b/app/assets/javascripts/html5_webshim.js
@@ -1,0 +1,14 @@
+$(document).on('ready page:load', function(){
+    window.webshim.setOptions('forms-ext', {
+    lazyCustomMessages: true,
+    replaceUI: true,
+    waitReady: true,
+    type: 'date',
+    date: {
+      startView: 2,
+      openOnFocus: true
+    }
+  });
+
+  window.webshim.polyfill('forms-ext');
+});
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -16,6 +16,7 @@
     <link href="https://www.stanford.edu/su-identity/css/su-identity.css" rel="stylesheet">
     <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700" %>
     <%= javascript_include_tag "application" %>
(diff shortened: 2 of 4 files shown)


jgreben:
Closing PR for time being. Waiting for comments from @jvine about usability and accessibility before pull request.

jkeck:
@jgreben it seems like you've included some upstream commits in a merge commit (62d1356fd0353a1215e90ae4afe9b00487b3ec66). I think what you may want to do is rebase your branch on top of master and fix any conflicts there.  I would be more than happy to look into this w/ you next week.

I may be completely wrong about this being a problem though, @cbeer let me know if I'm wrong here and I'll :shipit:.

jgreben:
I went ahead and pulled from master, rebased, and fixed a conflict in the item_selector_spec. Pushing again and making new PR. Let me know if it looks good now.

app/assets/javascripts/html5_webshim.js
```diff
@@ -0,0 +1,15 @@
...
+    waitReady: true,
+    type: 'date',
+    date: {
+      startView: 2,
+      openOnFocus: true
+    }
+  });
+
+  window.webshim.polyfill('forms-ext');
+  window.onunload = function(){};
```
jkeck: Just out of curiosity, what does this empty `window.onunload` function do for us?

jgreben: The pollyfill does not work if you navigate away from the page and back again. This fixes it somehow.. I think it has something to do with the bfcache. I forgot where I came across it, maybe it was this http://stackoverflow.com/questions/158319/cross-browser-onload-event-and-the-back-button

jkeck: Hmm, that seems fishy to me.  Is there any way I can get you to give me the steps to replicate that issue (or show me)?  From what I understand, binding to the `ready` and `page:load` events should handle that case.

PR #394, opened 2015-09-05
Added 'sorry' route for redirect failures from illiad

jgreben:


app/assets/javascripts/sorry.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
+// All this logic will automatically be available in application.js.
```
jkeck: I know this was auto-generated, but would you mind removing it since there is no actual content?

app/assets/stylesheets/sorry.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the sorry controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
jkeck: I know this was auto-generated, but would you mind removing it since there is no actual content?

app/views/sorry/unable.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<div>
+  <h2>We're sorry...</h2>
+</div>
+
+<div>
+  We're unable to complete your request right now.
+</div>
+
+<div>
+  Please contact the circulation desk in Green Library at greencirc@stanford.edu or 650-723-1493.
```
jkeck: We have this information configured here: https://github.com/sul-dlss/sul-requests/blob/master/config/application.rb#L151-L154

spec/controllers/sorry_controller_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+describe SorryController, type: :controller do
+  describe 'get #unable' do
+    it 'returns http success' do
+      get :unable
+      expect(response).to have_http_status(:success)
```
jkeck: :+1: on this test.

It seems strange that this page would return a successful status code though.  We should probably respond with an appropriate http status code for an error state.

app/controllers/sorry_controller.rb
```diff
@@ -0,0 +1,23 @@
+# Page to display if something goes wrong with the request
+# e.g. redirect from ILLiad logon or ScanAndDeliver pages
+class SorryController < ApplicationController
+  def unable
+    @please_contact = please_contact
+    render status: :forbidden
```
jkeck: Why `forbidden/403`? Do we know for sure that is the reasoning for routing to this page?

jgreben: Unless we use this page for something in addition to the redirect from illiad, they should only get to this page if they hit the illiad logon page, which means that they are not a user in illiad or something else went wrong with the authentication to illiad.  So perhaps 401 (unauthorized) instead? Or 204 (no_content)?

PR #396, opened 2015-09-09
Expose relevant Symphony request statuses to the patron

cbeer:


app/helpers/requests_helper.rb
```diff
@@ -63,42 +51,16 @@ def label_for_item_selector_holding(holding)
     end
   end
 
-  private
+  def holding_request_status(holding)
+    text = case holding.request_status.msgcode
+           when 'P001B', 'P002B'
```
cbeer: I think there's a conversation @jvine needs to have with Darsi to figure out the right mapping, but I hope this one is at least partly right..

PR #397, opened 2015-09-10
Move Illiad redirect to a before action filter and pass create via…

### LINKED ISSUE #309: Do not submit scan request to symphony if illiad fails

jkeck:
…GET as referrer.

Closes #309

### DIFF
--- a/app/controllers/scans_controller.rb
+++ b/app/controllers/scans_controller.rb
@@ -2,6 +2,12 @@
 #  Controller to handle particular behaviors for Scan type requests
 ###
 class ScansController < RequestsController
+  # Redirect to the illiad_url unless the illiad_success
+  # parameter (that we in the illiad_url method) is present
+  before_action only: :create do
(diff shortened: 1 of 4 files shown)


app/controllers/scans_controller.rb
```diff
@@ -2,6 +2,13 @@
 #  Controller to handle particular behaviors for Scan type requests
 ###
 class ScansController < RequestsController
+  # Redirect to the Illiad url if we have a POST request
+  # OR if not a POST request then check if the user can create a Scan object
+  # (e.g. a GET request redirected from a webauth challenge)
+  before_action only: :create do
+    redirect_to illiad_url if request.post? || can?(:create, Scan)
```
cbeer: I'm having trouble grasping this condition. Why not just `can?(:create, Scan)`?

jkeck: Switched this up to use a param that we pass in the URL we tell redirect Illiad back to us with.

cbeer:
:+1:

PR #403, opened 2015-09-10
Preserve user values after validation failures

### LINKED ISSUE #319: Keep user values when error fails back to the form
When a request fails and redisplays the form, we need to keep the user-entered values. Example at Hopkins:  Select a different delivery library, login to submit, request fails, and form redisplays with the default delivery library selected. Same issue with multiple items selected - if the request fails, the form redisplays with no items selected.

cbeer:
Fixes #319

### DIFF
--- a/app/assets/javascripts/item_selector/item_selector_ad_hoc_items.js
+++ b/app/assets/javascripts/item_selector/item_selector_ad_hoc_items.js
@@ -1,3 +1,5 @@
+//= require 'item_selector/item_selector_breadcrumbs'
+
 var itemSelectorAdHocItems = (function() {
   var defaultOptions = {
     selector: '[data-behavior="ad-hoc-items"]'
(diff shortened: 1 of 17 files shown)


cbeer:
If you have any ideas how to test this behavior, I'm all :ear:. It seemed unnecessarily tricky because:

 - in order to get a validation failure, it has to bypass the html5 + js validation
 - in order to check the results, we need javascript to populate the breadcrumbs..

jkeck:
Yeah, that's a tough one.  The only thing that I can think of is using javascript (via `page.execute` or something like that) to add `novalidate` to the form in the context of these tests.  That may not be the greatest idea but I think it would allow us to submit the form required fields left blank....  I'll leave it up to you, I'm good w/ :ship:ing it.

cbeer:
ok, I'll look into that and :shipit: if nothing good comes of it.

cbeer:
:shipit: ?

PR #404, opened 2015-09-11
Send contentHeight and successPage boolean via postMessage to the…

### LINKED ISSUE #295: x + cancel + close + caching
User confusion about redundant close actions and "aggressive" caching.

* [x] remove the x from the upper right corner 
* [x] make the Cancel button clear the cache
* [x] do NOT clear the cache if the user clicks outside the popup (i.e. does not explicitly cancel)
* [x] on the success page, change Cancel to Close

### LINKED ISSUE #346: Fixed size lightbox is a usability issue
There are several aspects with the iframe-in-a-fixed-size-lightbox that affect usability. Note that it wasn't designed with this constraint in mind.

1. small forms currently get a large, awkwardly sized box (can be made smaller, but doesn't help the remaining issues)
2. but many (most?) requests require scrolling to get to the Send button
3. once the Send button is visible, the user may not see the "I don't have a SUNet ID" button unless they realize there is further scrolling to be done
4. users do a lot of fussy moving between controls and scrollbar
5. the size of the lightbox overrides the user's sense of the size of the form
6. page up/down doesn't work the same way as Librarian view
7. the "Cancel" link outside the lightbox is counterintuitive once the user reaches the success page (they all thought it might actually cancel their request)

Overall, scrolling within the constrained space made the interaction feel less positive. Everyone who had tried the form has commented on this, even when they were completely successful with their request. 

I don't have an overall solution, though there are a couple of mitigating tickets - #338, #295.

jkeck:
…parent frame.

Closes #346 
Closes #295

### DIFF
--- a/app/assets/javascripts/postMessage.js
+++ b/app/assets/javascripts/postMessage.js
@@ -1,0 +1,25 @@
+var iframePostMessage = (function() {
+  return {
+    init: function() {
+      $(document).on('ready page:load', function(){
+        if (top != self) {
+          setInterval(function(){
+            var message = {
+              contentHeight: $('html').outerHeight(true),
+              successPage: $('.success-page').length > 0
+            };
+
+            // If we ever extend the message to include something other than
+            // contentHeight or the successPage boolean then we should replace
+            // '*' with the origin of who we will allow our messages to be sent
+            // to. As it stands the data we are sending isn't sensitive and
+            // can't be used to identify the user in any way.
+            parent.postMessage(message, '*');
+          }, 1000);
+        }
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/postMessage.js
```diff
@@ -0,0 +1,25 @@
+var iframePostMessage = (function() {
+  return {
+    init: function() {
+      $(document).on('ready page:load', function(){
+        if (top != self) {
```
jkeck: ಠ_ಠ Really @houndci....really?

PR #416, opened 2015-09-23
Required fields for users without a sunet id

jgreben:
Tested this with Jennifer to make sure it fits the spec.


app/assets/javascripts/no_sunet.js
```diff
@@ -0,0 +1,43 @@
+$(document).on('click', '#no-sunet-button', function(){
+  setTimeout(function(){
```
jkeck: Why the 1ms timeout here?

app/assets/javascripts/no_sunet.js
```diff
@@ -0,0 +1,43 @@
...
+    if(e.which == 13) {
+      $('.form-group input').checkForm(e);
+    }
+  });
+
+  $('.form-control').on('mouseenter mouseleave focus blur',function(e){
+    $('.form-group input').checkForm(e);
+  });
+
+  $('#no-sunet-submit').click(function(e) {
```
jkeck: This and the `$('form').on('keypress')` are kinda strange to me here.  Is this simply to capture the form submission (which is better handled with `$('form').on('submit')`)?

(btw, the `$('form')` selector will grab the first form in the page, if there is more than one form this may cause an issue)

app/assets/javascripts/no_sunet.js
```diff
@@ -0,0 +1,43 @@
+$(document).on('click', '#no-sunet-button', function(){
+  setTimeout(function(){
+      $('#request_user_attributes_library_id').focus();
+  }, 1);
+
+  $('form').on('keypress', function(e) {
+    if(e.which == 13) {
+      $('.form-group input').checkForm(e);
```
jkeck: This simply references the first `input` in the first `.form-group`. Might we want to deal w/ non `input` form fields now or in the future?

app/assets/javascripts/no_sunet.js
```diff
@@ -0,0 +1,43 @@
...
+    if ($('.form-group input').checkForm(e)){
+      $('form').submit();
+    }
+  });
+});
+
+$.fn.checkForm = function(e) {
+  var empty = true;
+
+  $(this).each(function() {
```
jkeck: Best I can tell, this simply loops through every element supplied that are not `#request_user_attributes_name` (in our context just `input`s) and checks to see if their value is present.  I'm not sure that is really what we ultimately want.

I think this would also cause an issue if a non-required input was in a `.form-group` element and not filled out, right?

app/assets/javascripts/no_sunet.js
jgreben: This enables the Library ID field to get the focus when this script runs, otherwise when the no-sunet button is clicked that button still has the focus and the the library_id field can't get it.

app/assets/javascripts/no_sunet.js
jgreben: '.form-group input' is an array of objects. I suppose I don't really need to specify input, but I guess I don't see what you would want to validate with this script that is not a user input, since the reason for the script is to check whether they have values for (some of) the user inputs.

app/assets/javascripts/no_sunet.js
jgreben: The keypress function id there to capture when there user is in the (e.g.) library_id field, and presses the enter key. If I use the on submit function (if I remember correctly when I was debugging it) it still submits the form which we did not want.

app/assets/javascripts/no_sunet.js
jgreben: So then what do we want? I guess I am confused about the requirements. 

#request_user_attributes_name is a non-required element in the .form-group, so I guess other non-required elements that are added would have to be treated the same.

app/assets/javascripts/no_sunet.js
atz: You should be able to block on submit from finishing from within the event.

app/assets/javascripts/no_sunet.js
jkeck: Isn't that what's already happening from here: https://github.com/sul-dlss/sul-requests/blob/master/app/assets/javascripts/toggle_show_tabs.js ?

app/assets/javascripts/no_sunet.js
jkeck: Sorry, I misspoke. Yes, that returns all the `inputs` in `.form-group` elements, but why are we dealing with them unconditionally?  In `checkForm` below you loop through all the matching elements and check if they have been filled out.  What about if there is an input that isn't required and does not need to be filled out?  Would this code add the tooltip if a non-required input was left blank?

app/assets/javascripts/no_sunet.js
jkeck: :+1: @atz 

And for reference checkout the jquery [submit()](https://api.jquery.com/submit/) documentation.  See the `event.preventDefault()` in the example.

app/assets/javascripts/no_sunet.js
jkeck: I'm worried that this code is brittle (one of the *code smells* is the blacklist of elements by ID).  There are several things somebody else could do to the form that would break this functionality in unexpected ways. I would be more than happy to chat about ways to accurately identify which inputs need to be validated.

jkeck:
Closed by #470

PR #428, opened 2015-09-30
Send notifications to mediators when mediated pages are created

### LINKED ISSUE #411: Email to mediator when a request is made
For libraries that do not get requests daily, it would be helpful to get an email alert when a request is made and they need to do something in the mediation interface. 

HV-ARCHIVE
PAGE-MP
HOPKINS
HOOVER

Send email to mailing list (need to identify) with link to mediation UI.

cbeer:
Fixes #411

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -154,6 +154,10 @@ module SULRequests
       }
     }
 
+    config.mediator_contact_info = {
+
+    }
+
     config.hours_api_location_map = {
       'ARS' => { library_slug: 'ars', location_slug: 'archive-recorded-sound' },
       'ART' => { library_slug: 'art', location_slug: 'library-circulation' },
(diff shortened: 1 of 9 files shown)


config/application.rb
```diff
@@ -154,6 +154,10 @@ class Application < Rails::Application
       }
     }
 
+    config.mediator_contact_info = {
+
```
jkeck: Is the idea that we will add this in at some point?  I don't think we symlink this anywhere so we won't be able to do it on the server. If that's the case, I wonder if we should track down that info (or make sure we have a task to know that we need to do that in order for this feature to be implemented).

cbeer: Yes, that's the idea. I think @jvine has that as a todo.

jvine: @cbeer, can you clarify what my to-do is here? Provide the mediator email addresses?

cbeer: Yes. In #411, you said:

> Send email to mailing list (need to identify) with link to mediation UI.

PR #432, opened 2015-10-01
Only validate needed_date for mediated pages when it is required.

### LINKED ISSUE #430: Don't require needed_date unless requires_needed_date? is true
Causing issues with HOPKINS and PAGE-MP mediated pages.

jkeck:
Fixes #430

### DIFF
--- a/app/models/requests/mediated_page.rb
+++ b/app/models/requests/mediated_page.rb
@@ -4,7 +4,7 @@
 class MediatedPage < Request
   validate :mediated_page_validator
   validates :destination, presence: true
-  validates :needed_date, presence: true
+  validate :needed_date_is_required
   validate :destination_is_a_pickup_library
 
   scope :archived, -> { where('needed_date < ?', Time.zone.today) }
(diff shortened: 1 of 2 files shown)


app/models/requests/mediated_page.rb
```diff
@@ -4,7 +4,7 @@
 class MediatedPage < Request
   validate :mediated_page_validator
   validates :destination, presence: true
-  validates :needed_date, presence: true
+  validate :needed_date_is_required
```
jkeck: I could easily be convinced to promote this logic to the `Request` class.

cbeer: I think we were getting a `required` badge on the form from this, maybe?

jkeck: Yeah, I had the same panic moment but it looks like we're already handling this in a way that will work: https://github.com/sul-dlss/sul-requests/blob/master/app/views/requests/_form.html.erb#L15

cbeer: :+1:

PR #434, opened 2015-10-02
Don't lint factories when only running a single test.

jkeck:
It has been getting annoying to lint all the factories just to run a single test (particularly on conf. wifi).


cbeer:
¯\_(ツ)_/¯

jkeck:
:shipit: it's a waste of time to lint completely unrelated factories for a test that may not even invoke any factory it all.

PR #440, opened 2015-10-06
Add a sucard_number setter that will process the SUCARDNUMBER ldap…

### LINKED ISSUE #438: Omit first 5 characters from SUCARD Number for Library ID
SUCARD Numbers are 80609##########
The first 5 digits are not part of the Library ID so we need to drop them before sending to symphony.

jkeck:
…attribute and set the library_id to the library id portion.

Closes #438

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -18,6 +18,11 @@ class User < ActiveRecord::Base
     end
   end
 
+  def sucard_number=(card_number)
+    return unless card_number.present?
+    self.library_id = card_number[/\d{5}(\d+)/, 1]
+  end
+
   def email_address
     case
     when library_id_user?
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -18,6 +18,11 @@ def to_email_string
     end
   end
 
+  def sucard_number=(card_number)
+    return unless card_number.present?
+    self.library_id = card_number[/\d{5}(\d+)/, 1]
```
cbeer: Should we validate the first 5 digits are what we think they are?

jkeck: My concern is that I don't know of those 5 digits (`80609`) are legitimately the only ones that will ever show up (another downside of this being completely undocumented information).

PR #449, opened 2015-10-09
Allow mediated pages with no needed_date to show up under active scope.

### LINKED ISSUE #446: Items with no needed date do not show up on mediation page
Active and Archived scopes are based on needed_date.  If there is no needed_date the record does not show up under either.

jkeck:
Fixes #446

### DIFF
--- a/app/models/requests/mediated_page.rb
+++ b/app/models/requests/mediated_page.rb
@@ -8,7 +8,7 @@ class MediatedPage < Request
   validate :destination_is_a_pickup_library
 
   scope :archived, -> { where('needed_date < ?', Time.zone.today) }
-  scope :active, -> { where('needed_date >= ?', Time.zone.today) }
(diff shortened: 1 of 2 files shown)


cbeer:
I guess we're saying those requests never fall off the page?

`¯\_(ツ)_/¯`

jkeck:
Yeah, I guess so.  If we define `archived` as the needed date has passed, and we allow requests without needed date, then they must always be active requests! :-/

jkeck:
And I'm more than happy to let @jvine give her $.02.

cbeer:
:shipit:

jvine:
I don't have a problem with them being on the page if they are always at the bottom. Both Hopkins and PAGE-MP are pretty low volume - we probably have a year before we have to consider another solution for them.

PR #458, opened 2015-10-13
Redirect CanCan::AccessDenied errors (when the user is not a…

### LINKED ISSUE #457: Redirect CanCan::AccessDenied errors success/status pages

jkeck:
…webauth user) to the webauth login.
Closes #457

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -75,11 +75,17 @@ class RequestsController < ApplicationController
   end
 
   def rescue_can_can(exception)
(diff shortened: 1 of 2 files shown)


jkeck:
@cbeer please double check that I'm not crazy here.

Also I don't know if you have a better pattern for dealing with when we should raise an access denied and when we should redirect to the webauth.  The whole authenticate on submit kinda makes this more confusing that it should be....

cbeer:
Yeah.. works for me, I guess.

PR #462, opened 2015-10-15
Update requester info display for the mediators

cbeer:


app/helpers/requests_helper.rb
```diff
@@ -95,10 +95,12 @@ def searchworks_link(item_id, item_title)
...
-    if user.webauth_user?
-      mail_to user.to_email_string
-    elsif user.email.present?
-      mail_to user.email, user.to_email_string
+    if user.webauth_user? || user.email_address.present?
+      mail_to user.email_address, user.to_email_string
+    elsif user.library_id_user?
+      user.library_id
+    else
+      user.to_s
```
jkeck: Looks like this case is untested.

PR #464, opened 2015-10-16
Add _top target to my account links.

### LINKED ISSUE #459: Make MyAccount links target="_blank"
Or whatever it is that we want to make sure that my account doesn't come up in the iframe.

jkeck:
Closes #459 

@jvine is `_top` the target we want in this case?


jvine:
Yes. _top.

PR #469, opened 2015-10-20
Display 'needed date' validation messages on the form when a request errors

cbeer:


cbeer:
I can't figure out what's going wrong with setCustomValidity() and Firefox. Even though it gets set correctly, something is switching it back.

PR #470, opened 2015-10-20
Library ID & Name + Email Form Validation 

jkeck:
cc @jgreben we can chat about this tomorrow if you would like.

required-fields


jgreben:
:+1:

PR #479, opened 2015-10-22
Move mouseflow to partial and render in both app layouts. Also change…

jkeck:
… to be rendered in production.

### DIFF
--- a/app/views/shared/_mouseflow.html.erb
+++ b/app/views/shared/_mouseflow.html.erb
@@ -1,0 +1,10 @@
+<% if Rails.env.production? %>
+  <script type="text/javascript">
+    var _mfq = _mfq || [];
+    (function() {
+      var mf = document.createElement("script"); mf.type = "text/javascript"; mf.async = true;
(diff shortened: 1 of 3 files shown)


app/views/shared/_mouseflow.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if Rails.env.production? %>
+  <script type="text/javascript">
+    var _mfq = _mfq || [];
+    (function() {
+      var mf = document.createElement("script"); mf.type = "text/javascript"; mf.async = true;
+      mf.src = "//cdn.mouseflow.com/projects/07ae47b8-19d1-406d-9d9e-bf9c7fe048bc.js";
```
cbeer: Should this be in an environment-specific `Setting`? Or just :shipit:?

jkeck: I think just :shipit: I'm not if that has a particular benefit and mouseflow will only work on our public facing sites anyhow.

PR #485, opened 2015-10-28
Add an interstitial page to have WebAuth redirect to to complete requests

jkeck:
This allows us to take over from WebAuth as soon as possible and show the user a message while we're processing (instead of leaving them in the WebAuth form where subsequent requests cause errors).

@cbeer `¯\_(ツ)_/¯`

### DIFF
--- a/app/controllers/interstitial_controller.rb
+++ b/app/controllers/interstitial_controller.rb
@@ -1,0 +1,23 @@
+###
+#  This controller simply handles incoming requests (and responds immediately)
+#  then redirects the user to the requested url. This allows us to take the user from WebAuth
+#  and send them to their destination w/o allowing them to submit multiple superfluous requests
+###
+class InterstitialController < ApplicationController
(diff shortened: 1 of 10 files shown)


jkeck:
@jvine what happens here is if the user doesn't have javascript they will get a link that says `Click here if you are not redirected.`

Via javascript we immediately turn that text into `Processing...` and then send the user off to continue submitting their request.

Let me know if any of that wording should be changed.

jkeck:
btw, this branch is currently deployed to the stage server.  There is no SW UI pointing to it however.

cbeer:
`¯\_(ツ)_/¯`. I'll try it out later today.

cbeer:
Looks good (enough) to me. Even if I spam the submit button, the javascript redirect doesn't get processed.

jkeck:
Okay, I guess let's :shipit: and see if the dup requests cease.

PR #488, opened 2015-10-31
Change item_title column to text so it can take variable title length.

jkeck:
@cbeer what do you think about this?  Is this a good approach or should we add something in the model to limit the item title?


cbeer:
:+1:

PR #525, opened 2016-08-16
Use turbolinks 5.x events

cbeer:


ndushay:
PR looks good.  Travis failing tests, less so.  Tests ran clean the second time I ran 'em on my laptop.

:+1: for merging if this fixes #566

jkeck:
We need to figure out why this is breaking the build before merging.  Otherwise all of our travis build will be borked.

Gemfile
```diff
@@ -26,7 +26,7 @@ gem 'config'
 # (pinning to < 4.1 until issue w/ webshims is resolved https://github.com/aFarkas/webshim/issues/560)
 gem 'jquery-rails', '< 4.1'
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
-gem 'turbolinks'
+gem 'turbolinks', '~> 5.0'
```
ndushay: should there be a comment about wanting version 5?

ndushay:
and yes, I got to "rails 5" from the solutions suggesting turbolinks 5:  http://stackoverflow.com/questions/38381022/link-to-links-not-loading-javascript-in-rails  is one of them;  http://stackoverflow.com/questions/18770517/rails-4-how-to-use-document-ready-with-turbo-links  in the first answer, edit 2, is the second.

Gemfile
jkeck: I don't even know what I comment like that might say.  TBH, this will implicit when we move to rails 5 anyway, so I'm not sure it's necessary to leave a breadcrumb at this point.

ndushay: no worries.  merging.

PR #563, opened 2016-10-05
Fixing some consistently broken tests

jkeck:


ndushay:
LGTM;  👍 

what about the other 2 spec files with failing tests?  will this help them too?

jkeck:
@ndushay not quite sure what you mean.  Are you still getting failures locally w/ this branch?

ndushay:
@jkeck I didn't try it originally;  i just assumed each of the 3 spec files that kept failing would need tweaking.  what you did works beautifully.  It's all green on my machine.

jkeck:
Ah, I see what you're saying. ac06010aec0bc9ce28ac0663232b2016583a870f fixed 5 of the 8 failing tests.  And then the other three were all in the `item_selector_spec` which needed the `wait_for_ajax` behavior.

PR #567, opened 2016-10-08
Improve visibility of on-site note.

### LINKED ISSUE #491: Remote users aren't seeing the "on site" note
Increased coverage of request form means more remote users finding items via Google and requesting (what they think will be) an e-version of the item. Spec is following up with users on a daily basis, and canceling requests when the user learns they must be on site.

Need a more effective solution than a note on screen.

jkeck:
Closes #491 

 `/requests/new?item_id=10738397&origin=SPEC-COLL&origin_location=MSS-30`

## Before
form-before

## After
form-after

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -17,15 +17,15 @@ module ApplicationHelper
   end
 
   def label_column_class
-    'col-sm-3'
+    'col-sm-4'
   end
 
   def label_column_offset_class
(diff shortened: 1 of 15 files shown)


ndushay:
LGTM; 👍  if @jvine signs off.

jkeck:
Updated screenshot after reviewing with @jvine.  We 👌 for review.

PR #570, opened 2016-10-10
confirmation email: include year in 'not needed after' date

### LINKED ISSUE #528: Add year to "Not needed after" date in confirmation email
SW-1679

ndushay:
closes #528 

Once I figured out 
* where the email text came from (`app/views/confirmation_mailer/request_confirmation.text.erb`) 
* where `format: :quick` came from (`config/locales/en.yml`  `date/formats`) 

then it was easy to use the `long` format instead of the `quick` format in the email.

@jkeck Presumably only the confirmation email needs the change, not the mediation email or the admin/library_table.


ndushay:
yeah - i was pretty sure this was okay, but not 100% confident.

PR #571, opened 2016-10-11
status page title includes item title

### LINKED ISSUE #532: Add item title to the page title of Status page
Before:
`<title>SUL Requests</title>`

After:
`<title>Request status: Buckminster Fuller papers</title>`

ndushay:
closes #532 

## Before  (look at text on tab)
status page title before

## After  (look at text on tab)
status page title after

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -10,7 +10,7 @@
     <!-- Mobile IE allows us to activate ClearType technology for smoothing fonts for easy reading -->
     <!--[if IEMobile]> <meta http-equiv="cleartype" content="on"> <![endif]-->
 
-    <title><%= SULRequests::Application.config.application_name %></title>
+    <title><%= yield(:page_title).blank? ? SULRequests::Application.config.application_name : yield(:page_title)%></title>
     <%= favicon_link_tag 'favicon.ico' %>
     <%= stylesheet_link_tag "application", media: "all" %>
     <link href="https://www.stanford.edu/su-identity/css/su-identity.css" rel="stylesheet">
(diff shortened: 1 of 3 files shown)


jkeck:
@ndushay you can get to the status page for an item by clicking on the `Status` link in the requests dashboard.  Also, I believe it's just `/status` after a request's show url (which as a super admin you should be able to see any of them).

app/views/mediated_pages/_status_header.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :page_title do %>Request status: <%= current_request.item_title %><% end %>
```
jkeck: I believe this would only affect the status page for `MediatedPages`.

I think a more appropriate place for this would be in the [`status.html.erb`](https://github.com/sul-dlss/sul-requests/blob/master/app/views/requests/status.html.erb) view as that renders the `status_header` partial (which can be overridden and customized for different requests types).

jkeck:
Also, FWIW, we should never have empty item titles.  These titles come from the SearchWorks Blacklight app and there is logic there to make sure we get something in the absence of a title field (e.g. an identifier).

ndushay:
I think this is now ready, unless I need to change the title on the status page for hold/recall and for scan requests.

jkeck:
#532 says to add the item title to the status page (and does not indicate that this is particular to any kind of request, therefore it should be for **all** request types).

Please just add the `content_for` line to `status.html.erb` and leave the partials alone.

ndushay:
Ya know, I tried that, but then got "SUL Requests" as the title for a mediated page.  :-(   I will continue to experiment.  (sad trombone)

jkeck:
If you get to that point again, I would be more than happy to take a look at the branch at that point to figure out what's going on.  That should certainly not work out like that.

PR #573, opened 2016-10-11
Add a data flag on the item approval link that allows us to ensure…

jkeck:
…that a single click only triggers a single ajax request.

I believe that this will alleviate some of the issues `User already has a hold` with item approval in the mediation admin.

## Before
approval-before

## After
approval-after


ndushay:
it would be AWESOME if this gets rid of that user experience problem.

PR #574, opened 2016-10-12
mediation page: title and status links should open in a new window

### LINKED ISSUE #509: Open links from Mediation page in new window
Add  `target="_blank"` to links in the Title and Status columns in mediation pages.

screen shot 2016-09-12 at 3 38 52 pm

jmartin-sul:
* requests_helper.rb: allow specification of additional html_options for link_to
* _library_table.html.erb: set target to '_blank' and rel to 'noopener noreferrer' for title and status links
* mediation_table_spec.rb: add tests for mediation page title and status links, check that link is defined to open in new window.

screen shot 2016-10-11 at 6 42 23 pm

closes #509
connects to #509

### DIFF
--- a/app/helpers/requests_helper.rb
+++ b/app/helpers/requests_helper.rb
@@ -100,8 +100,8 @@ module RequestsHelper
     Time.zone.parse(date.to_s).strftime('%Y-%m-%d %I:%M%P')
   end
 
-  def searchworks_link(item_id, item_title)
-    link_to item_title, "#{Settings.searchworks_link}/#{item_id}", 'data-behavior': 'truncate'
+  def searchworks_link(item_id, item_title, html_options = {})
+    link_to item_title, "#{Settings.searchworks_link}/#{item_id}", html_options
   end
 
   def requester_info(user)
--- a/app/views/admin/_library_table.html.erb
+++ b/app/views/admin/_library_table.html.erb
@@ -21,11 +21,11 @@
           </span>
         </td>
         <td class='needed_date'><%= time_tag(request.needed_date, :quick) %></td>
-        <td class='title'><%= searchworks_link(request.item_id, request.item_title) %></td>
+        <td class='title'><%= searchworks_link(request.item_id, request.item_title, target: '_blank', rel: 'noopener noreferrer', 'data-behavior' => 'truncate') %></td>
         <td class='requester'><%= requester_info(request.user) %></td>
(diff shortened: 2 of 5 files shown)


jmartin-sul:
also added `rel="noopener noreferrer"` to the links as per @jkeck's advice

app/helpers/requests_helper.rb
```diff
@@ -100,8 +100,9 @@ def format_date(date)
     Time.zone.parse(date.to_s).strftime('%Y-%m-%d %I:%M%P')
   end
 
-  def searchworks_link(item_id, item_title)
-    link_to item_title, "#{Settings.searchworks_link}/#{item_id}", 'data-behavior': 'truncate'
+  def searchworks_link(item_id, item_title, html_options = {})
+    html_options['data-behavior'] = 'truncate'
+    link_to item_title, "#{Settings.searchworks_link}/#{item_id}", html_options
```
jkeck: Do you think it would be worth adding a spec to `requests_helper_spec.rb` to validate that our HTML options get passed? I'm on the fence on if it would be useful or not, but it should be an easy enough spec to write.  Totally up to you.

app/helpers/requests_helper.rb
```diff
@@ -100,8 +100,9 @@ def format_date(date)
     Time.zone.parse(date.to_s).strftime('%Y-%m-%d %I:%M%P')
   end
 
-  def searchworks_link(item_id, item_title)
-    link_to item_title, "#{Settings.searchworks_link}/#{item_id}", 'data-behavior': 'truncate'
+  def searchworks_link(item_id, item_title, html_options = {})
+    html_options['data-behavior'] = 'truncate'
```
jkeck: BTW, I really like the `html_options` approach.  We could have taken this approach w/ the `data-behavior` attribute initially.

Setting this attribute like this won't allow a consumer of this method to prevent truncation or passing an alternate `data-behavior` attribute.

What would you think about moving this attribute into the two places that currently call `searchworks_link`?

jmartin-sul: since it seems like we want that as the default behavior, what about making it the default in the absence of the caller setting it explicitly, that way we don't have to add it to either of the places that currently call `searchworks_link`.  i.e.:
`html_options['data-behavior'] = 'truncate' unless html_options['data-behavior']`

jkeck: I would be fine with that.

PR #575, opened 2016-10-12
Don't lock the capistrano version.

jkeck:
This hasn't really bitten anybody and we should figure out deploy problems as they occur (since we'll also be managing our version of capistrano).


ndushay:
yes please.  I was wondering about that.  I suspect that practice started here when cap v3 came out ...

👍 👍 👍

jkeck:
Yeah, I think this is just holdover from old Cap configuration code.

PR #580, opened 2016-10-14
Add WebAuth email for users, so we can distinguish between users that…

### LINKED ISSUE #497: @stanford.edu email isn't available for all sunetids
greencirc@stanford.edu is getting bounces from request confirmation emails to users who do not have a sunetid@stanford.edu address. That means the users aren't getting their confirmations.

Apparently users such as visiting scholars can have a sunetid and library privileges, but no sunetid@stanford.edu email account. Their actual email is in Symphony. 

Can we do anything with this? Can we tell from the webauth response whether they have a full service sunetid and if not, treat them like libraryID-only users (which I think has no confirmation email)?

tingulfsen:
… have or do not have a SUNet email.

Fixes #497 

Note that in order for this to work we have to change configurations as shown for example in sul-dlss/shared_configs#154


app/controllers/application_controller.rb
```diff
@@ -36,9 +36,16 @@ def update_ldap_attributes(user)
     user.ldap_group_string = ldap_attributes['WEBAUTH_LDAPPRIVGROUP']
     user.sucard_number = ldap_attributes['WEBAUTH_LDAP_SUCARDNUMBER']
     user.affiliation = ldap_attributes['WEBAUTH_LDAP_SUAFFILIATION']
+    user.ldap_email = ldap_email
     user.save if user.changed?
   end
 
+  def ldap_email
+    return ldap_attributes['WEBAUTH_EMAIL'] unless ldap_attributes['WEBAUTH_EMAIL'].nil?
+    "#{user_id}@stanford.edu" if ldap_attributes['WEBAUTH_LDAP_SUEMAILSTATUS'] == 'active'
```
jkeck: Do you mean to `return` here?  I think it will always fall to `nil` in this case.

app/models/request.rb
```diff
@@ -110,8 +110,7 @@ def proxy?
   def notification_email_address
     if proxy? && user.proxy_access.email_address.present?
       user.proxy_access.email_address
-    elsif user.email_address.present?
-      user.email_address
+    else user.email_address
     end
```
jkeck: This reads a little weird (and TBH, I'm not sure what the behavior is for this).

Would you mind putting `user.email_address` on a separate line?

app/models/user.rb
```diff
@@ -33,7 +33,7 @@ def email_address
     when library_id_user?
       email_from_symphony
     when webauth_user?
-      "#{webauth}@stanford.edu"
+      ldap_email || "#{webauth}@stanford.edu"
```
jkeck: Do we want the `OR` case for this?  we already return the address `"#{webauth}@stanford.edu"` in [ldap_email](https://github.com/sul-dlss/sul-requests/pull/580/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR45).  Although this might be here because that is returning `nil` in that case (as I mention in my comment there).

app/controllers/application_controller.rb
tingulfsen: Whoa! Good catch!

tingulfsen:
I deployed this branch to stage and it looks good to me, with some limited testing using my own login.

jkeck:
👍 LGTM.  Want to give other folks a chance to review and comment.

app/models/user.rb
```diff
@@ -33,7 +33,7 @@ def email_address
     when library_id_user?
       email_from_symphony
     when webauth_user?
-      "#{webauth}@stanford.edu"
+      ldap_email
```
ndushay: what I'm missing in the PR is a test that actually exercises this change.  Is there a spec for the `email_address` method already?  Or do you need to add one?

PR #581, opened 2016-10-17
Add a click-through note to remote users for certain requests

### LINKED ISSUE #551: Add IP check to mediated, in-library requests
This would apply to Spec, Rumsey, and Hoover Archives.

Can we do an IP check when the user clicks the Request button, and if they are outside the "campus + VPN" range, popup up an interim page, like so:

on-site-access-ip address check

IP ranges are in Consul
https://consul.stanford.edu/display/ERU/IP+Addresses+-+Default

jkeck:
Closes #551 

not-at-su

### DIFF
--- a/app/models/ip_range.rb
+++ b/app/models/ip_range.rb
@@ -1,0 +1,58 @@
+##
+# A class to model a set of IP ranges.  The default IPAddr library in ruby
+# can define broad IP ranges (e.g. 172.12.0.0/16 would define 172.12.*.*)
+# but doesn't handle custom ranges as well (e.g. 172.22.24.* - 172.22.255.*)
+# This class allows you to check if a given IP is in a network of IP ranges.
+#
+# There is a network of IP ranges injected by default, but an object can be
+# injected that provides an array of singleton IP addresses (or ranges defined
+# by the slash prefixes) and an mutli-dimensional array of custom IP ranges.
+# Example network definition:
+# example_network = OpenStruct.new(
+#   singletons: ['172.12.0.0/16'],
+#   ranges: [['172.12.12.0', '172.12.56.255']]
+# )
+#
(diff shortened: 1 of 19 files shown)


ndushay:
LGTM;  👍

app/models/ip_range.rb
```diff
@@ -0,0 +1,58 @@
...
+  class << self
+    def includes?(ip_to_check)
+      new(ip_to_check).included?
+    end
+
+    def default_network
+      Settings.stanford_ips
+    end
+  end
+
```
tingulfsen: After merging, I realized that I didn't quite understand your intent here. What were you trying to achieve with lines 33 through 41?

spec/models/ip_range_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  context 'ranges' do
+    it 'IP addresses configured as custom ranges are properly identified' do
+      expect(described_class.new('171.10.15.6', test_network)).not_to be_included
+      expect(described_class.new('171.10.15.7', test_network)).to be_included
+      expect(described_class.new('171.10.15.20', test_network)).to be_included
+      expect(described_class.new('171.10.15.35', test_network)).to be_included
+      expect(described_class.new('171.10.15.36', test_network)).not_to be_included
+    end
+
+    it 'IP addresses configured as ranges using slash suffix are properly identified' do
```
jmartin-sul: seems like a totally reasonable test, but does the slash suffix actually come into play on these?

app/models/ip_range.rb
jkeck: They are two completely unrelated methods.  They are both providing class methods so we can say things like `self.class.default_network` as is in the initializer or `IPRange.includes?` as is used in the helper.

PR #589, opened 2016-10-19
Add live_lookup accessors to allow symphony live lookups to be optional

jkeck:


app/views/admin/holdings.html.erb
```diff
@@ -25,7 +25,7 @@
           </button>
         </td>
         <td class='col-xs-2'><%= item.home_location %></td>
-        <td class='col-xs-2'><%= item.current_location.code %></td>
```
jkeck: Removing the current location here as this will be filled in with the data from the asynchronous call to SymphonyWebServices.

PR #590, opened 2016-10-20
Use BOTH library ID and email if provided by non-SUNet users for requ…

### LINKED ISSUE #493: Users who enter invalid library ID are then not able to request by name + email
User enters a library ID plus name + email in a page request. Library ID is bad - either expired, or not the right kind, or whatever. Request fails. User then tries to enter only name + email, but we have permanently linked the name + email to the ID, and they continue to get "request failed" messages. User workaround is to use a different email address. Backend workaround is to manually delete the library ID from the requests database. Would be better to just use the info entered, and not link name + email to a previously entered ID.

SW-1588 plus December discussion in Slack channel around user marvina.

tingulfsen:
…ests.

Fixes #493

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -83,9 +83,17 @@ class Request < ActiveRecord::Base
     if user.webauth_user?
       User.find_by_webauth(user.webauth)
     elsif user.library_id_user?
-      User.find_by_library_id(user.library_id)
+      find_existing_library_id_user
     elsif user.non_webauth_user?
       find_existing_email_user
+    end
+  end
+
+  def find_existing_library_id_user
+    if user.email
+      User.find_by(library_id: user.library_id, email: user.email)
+    else
+      User.find_by_library_id(user.library_id)
     end
   end
 
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -98,7 +98,7 @@ class Request < ActiveRecord::Base
   end
 
   def find_existing_email_user
-    User.find_by_email(user.email).tap do |u|
+    User.find_by(email: user.email, library_id: user.library_id).tap do |u|
       next unless u
       u.update_attributes(name: user.name)
     end
--- a/spec/models/request_spec.rb
+++ b/spec/models/request_spec.rb
@@ -294,6 +294,27 @@ describe Request do
         expect(User.where(library_id: '12345', email: 'jstanford@stanford.edu').length).to eq 1
       end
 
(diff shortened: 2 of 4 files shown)


app/models/request.rb
```diff
@@ -80,12 +80,20 @@ def find_existing_user
...
-      User.find_by_library_id(user.library_id)
+      find_exisisting_library_id_user
     elsif user.non_webauth_user?
       find_existing_email_user
     end
   end
 
+  def find_exisisting_library_id_user
+    if user.email
+      User.find_by('library_id = ? AND email = ?', user.library_id, user.email)
```
jkeck: Active record provides methods to do this type of thing natively.

`User.where(library_id: user.library_id, email: user.email).first`

app/models/request.rb
```diff
@@ -80,12 +80,20 @@ def find_existing_user
     if user.webauth_user?
       User.find_by_webauth(user.webauth)
     elsif user.library_id_user?
-      User.find_by_library_id(user.library_id)
+      find_exisisting_library_id_user
```
jkeck: Is there a good way to test for this behavior?

ndushay: `exisisting` ... typo

spec/models/request_spec.rb
```diff
@@ -268,6 +268,22 @@
...
+          item_id: '1234',
+          origin: 'GREEN',
+          origin_location: 'STACKS',
+          user_attributes: {
+            library_id: '12345',
+            email: 'jstanford@stanford.edu'
+          }
+        )
+        expect(User.where(library_id: '12345', email: 'jstanford@stanford.edu').length).to eq 1
+      end
```
jkeck: Hmm, I'm curious if this actually tests our bug.  Does this fail before your change?

jkeck:
@tingulfsen and I paired on this a bit and think that we have a small change and tests that address the issue.  If anybody else wants to look over it before we merge, please feel free.

spec/models/request_spec.rb
```diff
@@ -278,6 +278,43 @@
...
+
+      it 'should not duplicate user records when both library ID and email is provided' do
```
ndushay: if this test is doing something different from the above test, this description is not making it clear - I am confused.

spec/models/request_spec.rb
ndushay: jkeck asks a good question - was it answered F2F?

PR #591, opened 2016-10-20
Add model to lookup current location from a barcode

ndushay:
@jkeck

### DIFF
--- a/app/models/symphony_curr_loc_request.rb
+++ b/app/models/symphony_curr_loc_request.rb
@@ -1,0 +1,55 @@
+require 'json'
+
+# API for retrieving the current location from a barcode via Symphony Web Services
+class SymphonyCurrLocRequest
+  include ActiveModel::Model # allows initialization with a hash of attributes
+
+  attr_accessor :barcode
+
+  def current_location
+    return '' if json.empty?
+    json['fields']['currentLocation']['key']
+  rescue NoMethodError => e
+    Rails.logger.warn("currentLocation not available for #{barcode}; failed with: #{e}")
+    return ''
+  end
+
+  private
+
+  def response
+    @response ||= faraday_conn_w_req_headers.get
+    return empty_response(@response.body) unless @response.success?
+    @response
+  rescue Faraday::Error::ConnectionFailed => e
+    empty_response(e)
+  end
+
+  def json
+    return {} unless response.success?
+    @json ||= JSON.parse(response.body)
+  rescue JSON::ParserError => e
+    Rails.logger.warn("Couldn't parse JSON from #{url}: #{e}")
+    {}
+  end
+
+  BASE_URL = "#{Settings.symphony_web_services.base_url}#{Settings.symphony_web_services.curr_loc_path}".freeze
+  QUERY_STR = { includeFields: 'currentLocation' }.to_query.freeze
+
+  def url
+    @url ||= "#{BASE_URL}#{barcode}?#{QUERY_STR}"
+  end
+
+  def empty_response(error = nil)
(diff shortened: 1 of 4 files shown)


app/models/symphony_curr_loc_request.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# API for retrieving the current location from a barcode via Symphony Web Services
+class SymphonyCurrLocRequest
+  include ActiveModel::Model  # allows initialization with a hash of attributes
+
+  attr_accessor :barcode
+
+  def current_location
+    return '' if json.blank?
+    json['fields']['currentLocation']['key'] || ''
```
ndushay: i was following a pattern here.  Not sure if we should silently return ''  if there was a problem.  (we log it, but ...)

app/models/symphony_curr_loc_request.rb
```diff
@@ -0,0 +1,53 @@
...
+    @response
+  rescue Faraday::Error::ConnectionFailed => e
+    empty_response(e)
+  end
+
+  def json
+    return {} unless response.success?
+    @json ||= JSON.parse(response.body)
+  rescue JSON::ParserError
+    {}
```
ndushay: similarly, not sure if this should return empty hash if there was an error ...   esp. as `currrent_location` method above kinda barfs.   Maybe it should say .empty?  instead of .blank?

config/settings/test.yml
```diff
@@ -13,3 +13,8 @@ token_encrypt:
   salt: 'dGcTBhXrQjsvQkbDQCkAIdogYjzOiIQpjToNLjqyzslKShfTvxIJHmMZhzdNjOqtDZKtxutyHBZmNHKOcGMfHzxwCFeOTBeEHYHNedtrmdfLzHMTRSDJCRMURaAjRrDq'
 
 mailer_host: 'example.com'
+
+symphony_web_services:
+  # trailing slashes expected unless it's the end of the URI path
+  base_url: 'http://example.com/symws/'
+  curr_loc_path: 'v1/catalog/item/barcode/'
```
ndushay: my thinking here is that there will be other symphony web services paths we will be using in the future, such as the one for lookupTitleInfo ...

app/models/symphony_curr_loc_request.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def url
+    @url ||= "#{base_url}#{barcode}?#{query_str}"
+  end
+
+  def base_url
+    @@base_url ||= "#{Settings.symphony_web_services.base_url}#{Settings.symphony_web_services.curr_loc_path}"
```
ndushay: is the idea that I would do something like
```ruby
BASE_URL = "#{Settings.symphony_web_services.base_url}#{Settings.symphony_web_services.curr_loc_path}"
```
because it will be a constant while the app is running.   Or is it suggesting I do something else?

app/models/symphony_curr_loc_request.rb
```diff
@@ -0,0 +1,50 @@
...
+  include ActiveModel::Model # allows initialization with a hash of attributes
+
+  attr_accessor :barcode
+
+  def current_location
+    return '' if json.empty?
+    json['fields']['currentLocation']['key']
+  rescue
+    return ''
+  end
```
ndushay: Not sure this should silenty return empty string if bad or missing json ...  kinda followed the pattern from `symphony_user_email_requests`

ndushay:
And I don't understand how to fix the display_coveralls_coverage rake task ... or why we should actually care about testing all those lines?

PR #593, opened 2016-10-21
Add the ability for users to create admin comments on requests...

### LINKED ISSUE #519: Add staff note field to request mediation space
Staff in Branner have requested an area in the staff mediation space for Branner Maps to record notes on a particular request.  They would use the notes field to record and easily communicate the logic behind a decision to approve or not approve a request.  While they didn't specify if the notes should be at the request level or the item level, I think it would have to be at the item level.  Any request based notes could just be recorded on the first item that is listed.

jkeck:
...they can manage.

Closes #519 

admin-comments-spec

### DIFF
--- a/app/assets/javascripts/adminComments.js
+++ b/app/assets/javascripts/adminComments.js
@@ -1,0 +1,77 @@
+/* global alert */
+
+var adminComments = (function() {
+  var adminCommentFormSelector = '#new_admin_comment';
+  var adminCommentListSelector = '[data-behavior="admin-comments-list"]';
+  var adminCommentToggleSelector = '[data-behavior="toggle-admin-comment-form"]';
+  var adminCommentFormCancelButton = '[data-behavior="admin-comment-cancel"]';
+
+  function commentTemplate(comment) {
+    return '<li>' +
+             comment.comment +
+             '<span class="text-muted">' +
+               ' - ' +
+               comment.commenter + ' - ' +
+               comment.created_at +
+              '</span>' +
+           '</li>';
+  }
+
+  return {
+    init: function(holdingsTable) {
+      this.holdingsTable = $(holdingsTable);
+      this.form = this.holdingsTable.find(adminCommentFormSelector);
+
+      this.addCommentToggleBehavior();
+
+      this.onAjaxSuccess();
+
+      this.onAjaxError();
+
+      this.addCancelButtonBehavior();
+    },
+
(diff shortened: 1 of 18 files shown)


app/assets/javascripts/adminComments.js
```diff
@@ -0,0 +1,75 @@
+var adminComments = (function() {
```
jkeck: This is used in another javascript library, but I guess Hound doesn't catch that.

app/assets/javascripts/adminComments.js
```diff
@@ -0,0 +1,75 @@
...
+  var adminCommentToggleSelector = '[data-behavior="toggle-admin-comment-form"]';
+  var adminCommentFormCancelButton = '[data-behavior="admin-comment-cancel"]';
+
+  function commentTemplate(comment) {
+    return '<li>' +
+             comment.comment +
+             '<span class="text-muted">' +
+               ' - ' +
+               comment.commenter + ' - ' +
+               comment.created_at +
```
jkeck: This comes from the active record attribute name, so I think we should keep it.

spec/requests/admin_comments_requests_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  let(:headers) { { 'HTTP_REFERER' => 'http://example.com' } }
+  let(:url) { "/mediated_pages/#{mediated_page.id}/admin_comments" }
+
+  before do
+    stub_current_user(user)
+  end
+
+  describe '#create' do
+    context 'when successful' do
+      it "merges the current user's webauth into the persisten comment as the commenter" do
```
jmartin-sul: typo? (`persisten`)

app/assets/javascripts/adminComments.js
```diff
@@ -0,0 +1,77 @@
...
+      this.onAjaxError();
+
+      this.addCancelButtonBehavior();
+    },
+
+    addCommentToggleBehavior: function() {
+      var _this = this;
+      var commentToggle = _this.holdingsTable.find(adminCommentToggleSelector);
+      commentToggle.on('click', function() {
+        _this.form.slideToggle();
```
jkeck: D'oh! Totally what I was intending by moving that into its own function.  Good catch!

jkeck:
@jmartin-sul good :eye:! I think I have addressed the issues you pointed out.

Issue #547, opened 2016-10-03
Get current location for barcode for mediation via symphony web services


jkeck:
Instead of using the SearchWorks Availability API to get the current location use the symphony web services API.

See #546 for details on the web services API.

ndushay:
Pieces:
- requests app to SW LiveLookup (current location, status ...): (send in flag?) needs to stop going to symphony for live availability info
- requests app to use Symphony Web Services directly to get live lookup by barcode as it will be faster than the SW call to symphony by catkey

jkeck:
For those following along at home, @ndushay and myself are submitting PRs to the [improved-item-lookup](https://github.com/sul-dlss/sul-requests/tree/improved-item-lookup) branch.  We'll then submit a PR from that branch to master once we're ready for review.

ndushay:
Also for those following at home, Shelley found a better symphony web services call for this purpose than the one suggested in #546.  Details in the code.

ndushay:
I'm probably confused - but how does this ticket differ from ticket #546?

PR #594, opened 2016-10-22
use SymphonyCurrLocRequest to get current location by barcode

ndushay:
The last piece ... for a much faster, but still synchronous, lookup.

Plan is to see if this is fast enough.  If not, we'll go asynch.

connected to #547

### DIFF
--- a/app/views/admin/holdings.html.erb
+++ b/app/views/admin/holdings.html.erb
@@ -33,7 +33,7 @@
           </button>
         </td>
         <td class='col-xs-2'><%= item.home_location %></td>
-        <td class='col-xs-2'></td>
+        <td class='col-xs-2'><%= SymphonyCurrLocRequest.new(barcode: item.barcode).current_location %></td>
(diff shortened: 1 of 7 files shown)


app/views/admin/holdings.html.erb
```diff
@@ -25,7 +25,7 @@
           </button>
         </td>
         <td class='col-xs-2'><%= item.home_location %></td>
-        <td class='col-xs-2'></td>
+        <td class='col-xs-2'><%= SymphonyCurrLocRequest.new(barcode: item.barcode).current_location %></td>
```
ndushay: Working on a feature;  if you want to deploy as is for the demo, that's cool.   It's actually deployed to requests-stage right now.

ndushay:
@jkeck your commit looks great.  shall we merge to `improve-item-lookup` branch and then rebase commits?

jkeck:
Merging this as it is into our feature branch.  Look for a PR to master soon-ish.

PR #597, opened 2016-10-25
Add a feature flag (initially set false by default) to determine if…

jkeck:
…we should render the remote user check.

The IP ranges we are working from are incomplete so this allows us to ship with this feature behind a flag until we can setting on what IP addresses truly identify on vs off campus.


tingulfsen:
This one is simple enough that I'm just going to merge it myself.

PR #608, opened 2016-10-27
Persist LDAP derived email addresses for webauth users in the…

jkeck:
…User.email attribute.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -32,8 +32,11 @@ class User < ActiveRecord::Base
     case
     when library_id_user?
       email_from_symphony
-    when webauth_user?
-      ldap_email
+    when webauth_user? && !email
+      # Fallback for users who were created before we started
+      # setting the email attribute for webauth users from LDAP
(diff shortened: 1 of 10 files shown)


app/models/user.rb
```diff
@@ -83,4 +86,12 @@ def email_from_symphony
       SymphonyUserNameRequest.new(libid: library_id).email
     end if library_id_user?
   end
+
+  private
+
+  def notify_honeybadger_of_unknown_webauth_email!
+    Honeybadger.notify(
+      "Webauth user record being saved without a valid email address. Using #{webauth}@stanford.edu instead."
+    ) if new_record?
```
tingulfsen: Is it possible for this method to be called when retrieving and old user, and that we're not actually trying to save the user at this point? Just wondering if the Honeybadger message needs to be massaged a little here.

jkeck: Ah, fair point.  I'll update the message to say

```
Webauth user record being created without a valid email address...
```

Does that sound good?

tingulfsen: 👍

PR #609, opened 2016-10-27
Add ok computer for app and dependency monitoring.

### LINKED ISSUE #548: Switch to OK Computer

jkeck:
Closes #548 

@ndushay what would our next steps be for swapping over?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,9 @@ gem 'honeybadger'
 # Use is_it_working to monitor the application
 gem 'is_it_working'
 
+# Use okcomputer to monitor the application
+gem 'okcomputer'
+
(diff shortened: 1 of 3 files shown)


jkeck:
We don't explicitly use any cache, so I don't think we should check that.

The ruby version check failed when I copied it over, it was unclear to my how that was supposed to work.

If we can't get to the SymphonyAPI users cannot make requests (which we would need to know about immediately).

If we cannot get to the web services API, the mediators cannot see the current location, which I think is non-crucial.

jkeck:
I re-ordered the registering of checks and added a comment denoting where the optional ones begin.

ndushay:
hound was hung;  i went ahead with the merge.

Issue #492, opened 2016-02-03
Scrolling="no" setting is hiding essential controls for some users


jvine:
We turned off scrolling on the iframe to prevent the bouncy behaviour (when the scrollbar pulses on and off). But some browsers are not resizing the window correctly, and are cutting off the bottom portion of the form. This often includes the Submit button... so the patron isn't able to send the request at all. 

Report in SW-1587 of cutting off the bottom of the calendar:
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/601.3.9 (KHTML, like Gecko) Version/9.0.2 Safari/601.3.9
Viewport: width:1173 height:811

Same thing on iPhone in Safari and Chrome:

![img_2460](https://cloud.githubusercontent.com/assets/6945691/12796335/0d793bf0-ca73-11e5-94ee-a883db84745e.PNG)

![img_2461](https://cloud.githubusercontent.com/assets/6945691/12796339/1400783a-ca73-11e5-99f6-890eec19bdec.PNG)

jvine:
See also SW-1610

ndushay:
Is this the same as "Modal Resizing" from our top five goals?  Marking this as high-priority.

I experience this behavior when coming from SW using Firefox on Mac.

jkeck:
@ndushay yeah.

jkeck:
A couple of issues that we might want to tackle to solve the various issues around this.

- [ ] Position the calendar above the date-field input (because the overlay will never change the height of the container)
- [ ] Allow for a scroll bar (or a default height) to be set if we can identify that we're not in the requests app anymore (lack of postMessage?).  This will allow the WebAuth form to be completely visible.
- [ ] Figure out other height related issues (once we can reliably reproduce them)

jkeck:
Last issue has been moved to #662

PR #611, opened 2016-10-28
Position the html5 popover shim to work better in the modal context.

jkeck:
Part of #492 
## Before

before

## After

after

### DIFF
--- a/app/assets/javascripts/html5_webshim.js
+++ b/app/assets/javascripts/html5_webshim.js
@@ -7,7 +7,14 @@ $(document).on('turbolinks:load', function(){
       type: 'date',
       date: {
         startView: 2,
-        openOnFocus: true
+        openOnFocus: true,
+        popover: {
+          position: {
+            my: 'center bottom',
+            at: 'center top',
+            collision: 'none'
+          }
+        }
       }
     });
 
--- a/app/assets/stylesheets/modules/forms.scss
+++ b/app/assets/stylesheets/modules/forms.scss
@@ -83,6 +83,13 @@ hr {
   display: none;
 }
 
+// This is done because the html5 shim uses an element with the .help-block class
+// to display error messages. When a successful date is chosen the shim will remove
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/html5_webshim.js
```diff
@@ -7,7 +7,14 @@ $(document).on('turbolinks:load', function(){
       type: 'date',
       date: {
         startView: 2,
-        openOnFocus: true
+        openOnFocus: true,
+        popover: {
```
ndushay: i love popovers.  My mom used to make them whenever she served standing rib roast.  So yummy!

app/assets/stylesheets/modules/forms.scss
```diff
@@ -83,6 +83,12 @@ hr {
   display: none;
 }
 
+// This is done because the html5 shim removes elements
+// with the .help-block class after selecting a date
```
ndushay: thanks for this comment!   I'm not clear enough on "html5 shim removes elements" meaning tho - does the .help-block class remove particular things in the dom?   This is just generic enough to make me wonder more about the problem this is fixing.  Is there a reference that could be supplied?   Or am I being too picky?

jkeck: I'll update the comment to be more clear.

I don't think we can reliably test this because the HTML5 shim + Capybara/Poltergeist don't play as nicely as they could.

jkeck:
@ndushay I believe I have addressed all the changes you have requested.  Please let me know if I missed anything.

app/assets/stylesheets/modules/forms.scss
```diff
@@ -83,6 +83,13 @@ hr {
   display: none;
 }
 
+// This is done because the html5 shim uses an element with the .help-block class
+// to display error messages. When a successful date is chosen the shim will remove
+// elements with a .help-block class which removes our note.
```
ndushay: much clearer, thank you.

Issue #514, opened 2016-04-27
Sort and search requests table


jvine:
Allow search and sort across all archived requests (not just in the current page).

jvine:
SW-1651
SW-1665

jvine:
Problems around:

- items moving to "archived" based on needed-on date
  - requests with no needed-on date never get archived
  - requests not approved for the requested date roll out of view
- difficulty for spec in seeing all the requests for one date when some may fall across a page break
- difficulty navigating a long list to find a specific date

Proposal:

- [ ]  filter list by "done" and "not done" 
- [ ] allow staff to mark items as "done" to move them out of pending view [separate issue #539]
- [ ] for libraries with "needed on" dates, add filters for next three business days
   - [ ] bonus points:  don't show a date that has no requests (pending or done)
- [ ] for libraries without "needed on" dates (Branner, Hopkins), show only Pending/Done
- [ ] within a selected date, show done and not done together
- [ ] add a "jump to date" calendar picker (same control as Hours)

![screen shot 2016-09-16 at 10 36 06 am](https://cloud.githubusercontent.com/assets/6945691/18597567/6f800332-7c03-11e6-88c9-bb6cfef9bfbc.png)

![screen shot 2016-09-16 at 12 05 35 pm](https://cloud.githubusercontent.com/assets/6945691/18598063/f33c3162-7c05-11e6-9b06-0b166771d35b.png)

![screen shot 2016-09-16 at 12 01 06 pm](https://cloud.githubusercontent.com/assets/6945691/18598210/81e53de6-7c06-11e6-8c13-00478b7f5891.png)

saseestone:
While it's true that usually we have the date (or a date range) for the request that we're looking for, that's not always the case.  We always know the name of the Requester.  So the ability to search the data table by requester would be extra helpful.

PR #612, opened 2016-11-01
Add a button to allow mediators to mark a request as completed/done…

### LINKED ISSUE #539: Allow mediator to mark a request as Done
Mediator wants to be able to move a request out of "Pending" state without approving it. Applies to all mediating libraries.
- [x] Add a "mark as done" button in the open request, above the table.

mediate - mark done and comment

Mediator clicks "mark as done":
- [x]  button becomes disabled, request is marked as "Done". 
- [x]  ideally, requests that are artificially closed can be differentiated from done-by-approval requests
- [x]  any approve buttons in the request should remain active: mediator should be able to approve an request at any time (see also #512 )

mediate - marked done disabled

"Marked done" vs "actually done", in context:

mediate - marked done

jkeck:
…without approving all items.

Closes #539 

_Marking this as a **[WIP]** because I notice there is something in #514 that may affect the approach we take here_

mark-as-done

### DIFF
--- a/app/assets/javascripts/item_approval.js
+++ b/app/assets/javascripts/item_approval.js
@@ -71,10 +71,9 @@ var itemApproval = (function() {
 
     updateAllApprovedNote: function(item) {
       if(this.allItemsAreApproved(item)) {
-        item.closest('tr.holdings')
-            .prev('tr')
-            .find('[data-behavior="all-approved-note"]')
-            .show();
+        var tableRow = item.closest('tr.holdings').prev('tr');
+        tableRow.find('[data-behavior="all-approved-note"]').show();
+        tableRow.find('[data-behavior="mixed-approved-note"]').hide();
       }
     },
 
--- a/app/assets/javascripts/markAsComplete.js
+++ b/app/assets/javascripts/markAsComplete.js
@@ -1,0 +1,53 @@
+/* global alert */
+
+var markAsComplete = (function() {
+  var buttonSelector = '[data-behavior="mark-as-complete"]';
+
+  return {
+    init: function(holdingsTable){
+      this.holdingsTable = $(holdingsTable);
+
+      this.onAjaxSuccess();
+      this.onAjaxError();
+    },
+
+    onAjaxSuccess: function() {
+      var _this = this;
+      _this.markAsCompleteForm().on('ajax:success', function(_, request) {
+        if(request.approval_status === 'marked_as_done') {
+          _this.disableMarkAsCompleteButton();
+          _this.markRequestRowAsMixedStatus();
+        }
+      });
+    },
+
+    onAjaxError: function() {
+      this.markAsCompleteForm().on('ajax:error', function() {
+        alert('There was a problem marking this request as complete.');
+      });
+    },
+
+    disableMarkAsCompleteButton: function() {
+      this.markAsCompleteButton().attr('disabled', 'true');
+    },
+
+    markRequestRowAsMixedStatus: function() {
+      var requestRow =  this.holdingsTable
+                            .closest('tr.holdings')
+                            .prev('tr');
+      if(requestRow.find('[data-behavior="all-approved-note"]').is(':hidden')) {
+        requestRow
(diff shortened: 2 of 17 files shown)


app/assets/javascripts/markAsComplete.js
```diff
@@ -0,0 +1,51 @@
+/* global alert */
+
+var markAsComplete = (function() {
```
jkeck: It is used in another file, but hound does not seem to catch that.

jkeck:
@jvine a question that came up when I was going over this with @ndushay: Is it okay that there is no way to undo the "Mark as done" action?

jvine:
@jkeck, question back to you:  When a request is marked as done, is it possible to go back and actually change it? like, approve another item, make a comment?

jkeck:
Just as a heads up, we'll want to do a little data migration after we deploy a release with this change.  It should be something along the lines of:

```ruby
MediatedPage.archived.each do |request|
  if request.all_approved?
    request.approval_status = :approved
  else
    request.approval_status = :marked_as_done
  end
  request.save!
end
```

We'll also probably want to do this after work hours (once the folks mediating requests have finished for the day).

jkeck:
And I answered @jvine's question in person, but just for posterity, marking as done does not restrict the request from taking any other action (it's simply a state that is stored in the database).

app/assets/javascripts/markAsComplete.js
```diff
@@ -0,0 +1,53 @@
...
+      this.holdingsTable = $(holdingsTable);
+
+      this.onAjaxSuccess();
+      this.onAjaxError();
+    },
+
+    onAjaxSuccess: function() {
+      var _this = this;
+      _this.markAsCompleteForm().on('ajax:success', function(_, request) {
+        if(request.approval_status === 'marked_as_done') {
```
jkeck: This is coming from an active record attribute.

app/assets/javascripts/item_approval.js
```diff
@@ -71,10 +71,14 @@ var itemApproval = (function() {
...
-            .find('[data-behavior="all-approved-note"]')
-            .show();
+        var tableRow =  item.closest('tr.holdings')
+                            .prev('tr');
+
+            tableRow.find('[data-behavior="all-approved-note"]')
+                    .show();
+
+            tableRow.find('[data-behavior="mixed-approved-note"]')
+                    .hide();
```
ndushay: if I'm reading the above correctly, the indentation is misleading.  Lines starting `tableRow` are at the same level of nesting as `var tableRow`, correct?    There's also no particular reason to put the last `.blah()` part on the next line, is there?   Lastly, there's an extra space after `=` in line 74

db/schema.rb
```diff
@@ -35,8 +35,8 @@
     t.string   "type"
     t.date     "needed_date"
     t.string   "status"
-    t.datetime "created_at",         null: false
-    t.datetime "updated_at",         null: false
+    t.datetime "created_at",                     null: false
+    t.datetime "updated_at",                     null: false
```
ndushay: @another wacko alignment thing - doesn't really matter here, and probably came from automagic rails stuff anyway.

ndushay:
Please pardon my ignorance, but is there any way the data migration could be part of the db migration in the rails app??   (i'm guessing not or you would have done it that way already)

db/schema.rb
jkeck: Yeah, this is auto-generated by rails.  I figure it is best not to hand-edit the schema.

jkeck:
@ndushay wrt putting data migrations in with rails database migrations, I believe that is typically considered [bad practice](https://robots.thoughtbot.com/data-migrations-in-rails).

I would be happy to add something akin to what's in that blog post though.

ndushay:
ah - thanks for the reference.  I think a temporary rake task is a good solution - should one be added to this PR?

jkeck:
@ndushay: 538dfb611ef0b3a51505393d920cda5eef5b96f9

PR #616, opened 2016-11-01
Run background jobs with Sidekiq/Active Job.

### LINKED ISSUE #587: Add background workers for requests
Should this reside locally on the requests VMs or on a separate VM?

The worker basically executes long-running HTTP API calls.  It takes an active record object as a parameter and needs to update the active record object with the results from the HTTP response.

tingulfsen:
Fixes #587 

This is a shameless green introduction of Sidekiq as a background job processor.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,9 @@ gem 'web-console', '~> 2.0', group: :development
 # test coverage in static code analysis GUI
 gem 'codeclimate-test-reporter', group: :test, require: false
 
+# Sidekiq is our background processing framework, run via Active Job
+gem 'sidekiq'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug'
--- a/app/jobs/submit_symphony_request_job.rb
+++ b/app/jobs/submit_symphony_request_job.rb
@@ -3,12 +3,26 @@
 class SubmitSymphonyRequestJob < ActiveJob::Base
   queue_as :default
 
-  def perform(request, options = {})
+  def perform(request_id, options = {})
     return true unless enabled?
+    request = find_request(request_id)
+
+    return true unless request
 
     response = Command.new(request, options).execute!
     request.merge_symphony_response_data(response.with_indifferent_access)
     request.save
+  end
+
+  def find_request(request_id)
+    request = begin
+      Request.find(request_id)
+    rescue ActiveRecord::RecordNotFound
+      Honeybadger.notify(
+        "Attempted to call Symphony for Request with ID #{request_id}, but no such Request was found."
+      )
+    end
+    request
   end
 
   def enabled?
--- a/app/models/concerns/symphony_request.rb
+++ b/app/models/concerns/symphony_request.rb
@@ -5,8 +5,12 @@ module SymphonyRequest
     user.webauth_user?
   end
 
-  def send_to_symphony!(options = {})
-    SubmitSymphonyRequestJob.perform_now(self, options)
+  def send_to_symphony_now!(options = {})
+    SubmitSymphonyRequestJob.perform_now(id, options)
+  end
+
+  def send_to_symphony_later!(options = {})
+    SubmitSymphonyRequestJob.perform_later(id, options)
   end
 
   def symphony_request
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -120,7 +120,7 @@ class Request < ActiveRecord::Base
   end
 
   def submit!
-    send_to_symphony!
+    send_to_symphony_later!
   end
 
   def barcode_present?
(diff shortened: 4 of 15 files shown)


app/jobs/submit_symphony_request_job.rb
```diff
@@ -3,14 +3,28 @@
 class SubmitSymphonyRequestJob < ActiveJob::Base
   queue_as :default
 
-  def perform(request, options = {})
+  def perform(request_id, options = {})
```
jkeck: Hmm, maybe this is my naiveté (given I haven't worked much with background jobs) but I was under the impression that with `ActiveJob` we didn't have to pass the ID any longer as long as the object that we're passing into the job implements `GlobalID` (which `ActiveRecord` does).

I'm basing this off the [GlobalID section in the ActiveJob Rails Guide](http://edgeguides.rubyonrails.org/active_job_basics.html#globalid).  Was that not working for you in the requests app?

app/models/concerns/symphony_request.rb
```diff
@@ -6,7 +6,7 @@ def appears_in_myaccount?
   end
 
   def send_to_symphony!(options = {})
-    SubmitSymphonyRequestJob.perform_now(self, options)
+    SubmitSymphonyRequestJob.perform_later(id, options)
```
jkeck: I was thinking we could pass something into this method that would allow us to switch on `perform_now` vs `perform_later` (so that mediated item approval can be run inline).   More than happy to take a look at that with you if you would like.

jkeck:
Lookin' good so far @tingulfsen!  Just a few comments (even though I know this is just a `WIP`).  Hope you don't mind. :)

app/jobs/submit_symphony_request_job.rb
tingulfsen: I think that will work, but the author of Sidekiq seems to think otherwise:

https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

It may only be a preference though:

http://stackoverflow.com/questions/30962979/should-you-still-pass-only-the-object-id-when-using-activejob

app/models/concerns/symphony_request.rb
tingulfsen: I wrote that code before the comment regarding mediated item approval, so yes, this has to change.

app/jobs/submit_symphony_request_job.rb
tingulfsen: I'll set up a couple of fake Slack accounts, and see if I can start a flamewar on this topic 😜

Gemfile
```diff
@@ -61,6 +61,8 @@ gem 'web-console', '~> 2.0', group: :development
 # test coverage in static code analysis GUI
 gem 'codeclimate-test-reporter', group: :test, require: false
 
+gem 'sidekiq'
```
ndushay: in keeping with this gemfile practice, there should be a comment indicating why we are including this gem.

app/jobs/submit_symphony_request_job.rb
jkeck: I'm 👍 to keeping the approach you've gone with.

app/models/request.rb
```diff
@@ -120,7 +120,7 @@ def notification_email_address
   end
 
   def submit!
-    send_to_symphony!
+    send_to_symphony_later!
```
jkeck: Can we add a `submit!` method to the `Scan` class that overrides this to do a `send_to_symphony_now!` (given the complication about requiring a browser with webauth credentials to pass through Illiad for scan requests).

app/models/requests/scan.rb
```diff
@@ -25,6 +25,10 @@ def destination
     'SCAN'
   end
 
+  def submit!
+    send_to_symphony_now!
```
jkeck: Thanks for adding this!

One more minor thing, would you mind writing a test in `scan_spec` (akin to what we have in `request_spec`) that asserts this behavior?

jkeck:
Random question.  Sidekiq provides an admin UI of its own to look at the jobs queue I believe.   I know we plan on creating some additional tickets, so I just wanted to call that out as one we'll want to create as part of this work.

👍 looking god.

app/models/requests/scan.rb
tingulfsen: Good point - I added a test in spec/models/requests/scan_spec.rb. This is also exercised in spec/controllers/scans_controller_spec.rb.

tingulfsen:
I've added tickets #627, #628, #629 and #630 as additional work that should follow up on this.

jkeck:
I'm dismissing @ndushay's requested changes as it appears that you have addressed them (and this is going to a feature branch anyhow).

PR #617, opened 2016-11-02
Make Hoover library no longer mediatable

### LINKED ISSUE #578: Remove Hoover Library from mediated library list
Hoover Library does not need to be in the mediated library list.  Paul Thomas and Maria Quinonez  have indicated that any items physically stored at SAL3, with a Library = HOOVER and a Location = STACKS-30 should have a request link, and the items can be added to the SAL3 picklist without prior staff approval.

jkeck:
(which means it will fall-back to a simple Page request).

Closes #578 

We will need to make a PR to the Requests shared_configs branches to remove `HOOVER` from the `origin_admin_groups`, but we should get this deployed first to ensure that no requests end up in their mediation queue in the interim.

### DIFF
--- a/spec/models/concerns/mediateable_spec.rb
+++ b/spec/models/concerns/mediateable_spec.rb
@@ -40,28 +40,15 @@ describe Mediateable do
         expect(subject).to_not be_mediateable
       end
     end
-    describe 'HOOVER' do
-      describe 'library' do
-        before { subject.library = 'HOOVER' }
-        it 'should return true if the item is in a *-30 location' do
-          subject.location = 'SOMEWHERE-30'
-          expect(subject).to be_mediateable
-        end
-        it 'should return false if the item is not in a *-30 location' do
-          subject.location = 'SOMEWHERE-ELSE'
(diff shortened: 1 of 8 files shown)


spec/models/concerns/mediateable_spec.rb
```diff
@@ -40,28 +40,15 @@ class MediateableTestClass
...
-        end
+    describe 'HV-ARCHIVE' do
+      before { subject.library = 'HV-ARCHIVE' }
+      it 'should return true if the item is in a *-30 location' do
+        subject.location = 'SOMEWHERE-30'
+        expect(subject).to be_mediateable
       end
-      describe 'archives' do
-        before { subject.library = 'HV-ARCHIVE' }
-        it 'should return true if the item is in a *-30 location' do
```
ndushay: wondering why this test is no longer valid.

jkeck: I've just reformatted the tests since a Hoover describe block with a Hoover Archives describe block.  This test still exists above.

PR #618, opened 2016-11-02
Allow for location specific item limits to be configured...

### LINKED ISSUE #606: Add 5-item limit to SAL3 PAGE-SP items
For items that are not in Spec as a library, but are paged to Spec, add 5-item limit.

Example record https://searchworks.stanford.edu/view/35581 

SW-1709

jkeck:
...(and configure a PAGE-SP limit).

Closes #606 

`/requests/new?item_id=2789859&origin=SAL3&origin_location=PAGE-SP`

## Before
2789859-before

## After
2789859-after

### DIFF
--- a/app/models/requests/mediated_page.rb
+++ b/app/models/requests/mediated_page.rb
@@ -2,12 +2,6 @@
 #  Request class for making page requests that require mediation
 ###
 class MediatedPage < Request
-  ITEM_LIMITS = {
-    'HV-ARCHIVE' => 20,
-    'RUMSEYMAP' => 5,
-    'SPEC-COLL' => 5
-  }.freeze
-
   validate :mediated_page_validator
   validates :destination, presence: true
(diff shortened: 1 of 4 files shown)


app/models/requests/mediated_page.rb
```diff
@@ -40,10 +34,6 @@ def requestable_with_sunet_only?
     false
   end
 
-  def item_limit
-    ITEM_LIMITS[origin] || super
```
ndushay: huh - was super just doing nothing??  odd.

jkeck: In this context, super fell back to `nil`.  This PR effectively moves this logic up to the common viewer and since `ITEM_LIMITS['unconfigured-origin']` is also `nil` so the behavior is effectively identical (although we've sort-of removed a layer of indirection).

PR #620, opened 2016-11-03
update picklist

### LINKED ISSUE #495: Spec needs additional info in the picklist
Need to add the following to the picklists:
- [x] Comment field (not the "add volumes" comment, which is already there - the "I want to bring my dog" comment) (**request_comment**)
- [x] Planned date of use (**needed_date** I think)
- [x] Copy note (the public note that displays below the item in the Availability block and in the Request form) example: https://searchworks.stanford.edu/view/2148836

ndushay:
closes #495

This adds all desired info to the Symphony Request sent to the API Darsi set up;  Darsi is ready to deploy her side to production.

The branch for this PR is currently deployed to  `requests-stage`.

# Before
status before picklist chgs

# After

- `req_comment` is new
- `item_comments` used to be `comments`  (none in example below)
- `copy_note` is new
- `requested_date` is new

status after picklist chgs


ndushay:
#634 will add the remaining bit, `copy_note` to the requests sent to symphony.

Once #634  is merged into `495-add-info-to-picklist` branch, and @dlrueda is ready, the deployment will go like this:

1.  Darsi will deploy her new code to prod.  It accepts `comments` and `item_comments` both, so it will work with current production and with these changes.
2.  after 1, we may deploy this fix to production.

app/jobs/submit_symphony_request_job.rb
```diff
@@ -85,6 +90,15 @@ def barcodes
       items
     end
 
+    def copy_notes
+      return if request.public_notes.blank?
+      result = []
+      request.public_notes.each_pair do |name, val|
```
jkeck: This kind of confused me.  Few questions.  Why use `each_pair` here?  Would `each` be more idiomatic?

Also, the block variables may be more clear as `barcode` and `note`?

ndushay: > Why use each_pair here? Would each be more idiomatic?
I guess i thought i needed each_pair.  whoops.

> the block variables may be more clear as barcode and note
Great suggestion

app/assets/javascripts/item_selector/item_selector_ad_hoc_items.js
```diff
@@ -5,12 +5,12 @@ var itemSelectorAdHocItems = (function() {
     selector: '[data-behavior="ad-hoc-items"]'
   };
 
-  var santizeCallnumber = function(callnumber) {
+  var sanitizeValue = function(callnumber) {
```
ndushay: ah - good call.  I think I did that at one point and lost the commit.  will fix anon.

PR #635, opened 2016-11-09
Set the remote_ip_check feature flag to true.

jkeck:


ndushay:
silly question:  do you want to do this in a feature branch as well?

jkeck:
I'm not sure why we would use a feature branch for this.

ndushay:
I meant while you get the kinks out.  To avoid inadvertently turning it on in prod.

ndushay:
Jessie needs to test something because he was getting errors for off-campus (?) IPs

PR #640, opened 2016-11-10
requests dashboard filter by type

### LINKED ISSUE #625: Allow requests dashboard to be filtered by request type

jmartin-sul:
allows users to filter the requests dashboard by type.  each type name (under its respective count) links to a version of the page filtered by type.  if a type filter is on, an `[x]` link to its right provides a way to remove that filter and show all requests.  i was just running with @jkeck's idea for how to implement this simply, it could definitely use design vetting from @jvine.

attaching screenshots from the capybara tests because i did something to mess up my local test data when instantiating test objects from the factory, and the default unfiltered version of the page now 500s on me.  would be nice to test this manually, either locally or on stage, before merging.

capybara-201611101050404614639155

capybara-201611101050402221951178

capybara-201611101050401483245569

closes #625

### DIFF
--- a/spec/features/admin_requests_spec.rb
+++ b/spec/features/admin_requests_spec.rb
@@ -13,7 +13,14 @@ describe 'Viewing all requests' do
                       request_comment: 'I can has this item?',
                       user: User.create(name: 'Joe', email: 'joe@xyz.com')
               )
+        create(:mediated_page, ad_hoc_items: ['ZZZ-123'],
+                               item_title: 'A Different Type of Page',
+                               origin: 'SPEC-COLL',
+                               request_comment: 'I can has this mediated item?',
+                               user: User.create(name: 'Jane', email: 'jane@example.com')
+              )
       end
+
       it 'should list data in a table' do
         visit admin_index_path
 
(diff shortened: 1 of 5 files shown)


spec/features/admin_requests_spec.rb
```diff
@@ -23,8 +30,45 @@
         expect(page).to have_css('td a[data-behavior="truncate"]', text: 'An American in Paris')
         expect(page).to have_css('td a[href="mailto:joe@xyz.com"]', text: /Joe \(joe@xyz.com\)/)
 
+        expect(page).to have_css('td a[data-behavior="truncate"]', text: 'A Different Type of Page')
+        expect(page).to have_css('td a[href="mailto:jane@example.com"]', text: /Jane \(jane@example.com\)/)
+
         expect(page).to have_selector('table.table-striped', count: 1)
       end
+
+      it 'give a link to filter by request type' do
```
ndushay: better?  `allow filtering by request type` ?

jvine:
👍  The interaction looks fine to me. We do a fancier version with the x icon in SW, but I don't think the added complication is necessary for this context.

spec/features/admin_requests_spec.rb
jmartin-sul: advice taken

app/models/dashboard.rb
```diff
@@ -2,6 +2,14 @@
 #  Class to handle dashboard metrics
 ###
 class Dashboard
+  METRIC_TYPE = {
+    custom: 'Custom',
+    hold_recalls: 'HoldRecall',
+    mediated_pages: 'MediatedPage',
+    pages: 'Page',
+    scans: 'Scan'
+  }.freeze
```
jkeck: This seems a bit strange to have.   Can we just use `classify` on the string representation of these metrics?

```
:custom.to_s.classify
=> "Custom"

:hold_recalls.to_s.classify
=> "HoldRecall"
```

jmartin-sul:
i think @jkeck's suggestion has been implemented, last travis run should hopefully go green in a couple minutes (capybara error, just restarted)...

jkeck:
👍

PR #645, opened 2016-11-11
install capistrano-sidekiq

### LINKED ISSUE #641: add capistrano-sidekiq
example in exhibits;  not clear what it does yet or if we need it.

jmartin-sul:
in the little testing i did, i was able to successfully stop, start, and restart sidekiq (on dev from my laptop) using the capistrano tasks.

the second commit is not strictly related to the capistrano-sidekiq stuff, but my rails server was crashing without the change when i tried to start it, since `config/environments/development.rb` expects it to be set.  seemed like it wasn't worth a separate PR, but happy to split out if people would prefer.

@tingulfsen?

closes #641

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -25,6 +25,7 @@ require 'dlss/capistrano'
 require 'capistrano/passenger'
 require 'capistrano/honeybadger'
 require 'capistrano/shared_configs'
+require 'capistrano/sidekiq'
 
 # Load custom tasks from `lib/capistrano/tasks` if you have any defined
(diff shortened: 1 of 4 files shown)


jmartin-sul:
i didn't end up adding `require 'capistrano/sidekiq/monit'` to `Capfile`, because running those tasks gave me errors when i did, and because i assumed the web UI would give us the monitoring we're interested in.  exhibits didn't set it up either, AFAICT.

Issue #630, opened 2016-11-05
Add sidekiq-statistic


tingulfsen:
[Sidekiq-statistic](https://github.com/davydovanton/sidekiq-statistic) provides additional display of stats for Sidekiq. We should consider adding this to Requests as an additional monitoring/debugging tool.

ndushay:
included as part of #629;  it was simply a matter of adding the gem to the Gemfile.

PR #646, opened 2016-11-11
set up /sidekiq route to web view, and ensure it's only avail to superusers

### LINKED ISSUE #629: Set up the Sidekiq web UI
Our chosen background processing framework - Sidekiq - comes with a rather nice [web application/UI](https://github.com/mperham/sidekiq/wiki/Monitoring) that displays statistics and statuses.

We should consider setting this up for Requests.

I suppose one issue would be who (what user level) gets to access this page?

ndushay:
closes #629

note that sidekiq-statistics is in (#630) because the sidekiq page has a statistics heading to the far right

# Before (logged in as superadmin)
before sidekiq

# After (logged in as superadmin)
## Home page
after sidekiq home page

## Followed link
note that sidekiq-statistics heading to the far right

after sidekiq page

# After (not logged in), going directly to route
after sidekiq not authorized


ndushay:
that refactor is thanks to Jessie.  I was stymied.

config/routes.rb
```diff
@@ -6,12 +6,17 @@
   # You can have the root of your site routed with "root"
   root 'home#show'
 
+  require 'sidekiq/web'
+  require 'sidekiq_constraint'
+  Sidekiq::Web.set :session_secret, Rails.application.secrets[:secret_key_base]
```
ndushay: I did what they did in exhibits.  What say you, @jkeck ?

jkeck: It seems to be what the sidekiq folks suggest (setting the session_secret in the routes).  I'm guessing because it has something to do with how the context in which the rack app is loaded and mounted vs. how the parent rails app is initialized (which is where `config/initializers` comes into play).  I think we shoulkd keep it as-is.

jmartin-sul: 👍

PR #656, opened 2016-11-15
Changes to the success and status pages for background jobs

### LINKED ISSUE #585: Change text on success page for move to background workers
The success page (where a user gets sent right after submitting their request) will need the text changes since the actual status of the request will be available via email and the status page.

We should not block work on getting background workers setup on this, but will want to get them deployed together.

jkeck:
Closes #585 

## Page after request is submitted
success

## Status page for pending request
pending-status

## Success page while items are being approved
mixed-results


config/locales/en.yml
```diff
@@ -97,8 +111,8 @@ en:
   requests:
     success:
       email_notification:
-        proxy: "(We've sent a copy of this request to your email and to the designated notification address.)"
-        default: "(We've sent a copy of this request to your email.)"
+        proxy: "We'll send an email to you and to the designated notification address when processing is complete."
+        default: "We'll send you an email when processing in complete."
```
jmartin-sul: `is complete`?  if so, appears to have been copy/pasted into some tests (which is actually how i noticed it).

jkeck: Good catch.  Thanks, I have updated that.

app/views/requests/success.html.erb
```diff
@@ -1,40 +1,25 @@
...
-  <% if current_request.is_a?(MediatedPage) || current_request.symphony_response.success? %>
-    <dl class='dl-horizontal dl-invert'>
-      <dt class='sr-only'><%= current_request.class.human_attribute_name :user %></dt>
-      <% unless current_request.user.library_id_user? %>
-        <dd>
-          <span class="requested-by"><%= current_request.user.to_email_string %></span>
-        </dd>
-      <% end %>
+  <h1 id='dialogTitle'>
+    <span class="sul-i-check-2" aria-hidden="true"></span> We've received your request
```
jkeck: Yes, the span simply provides the icon and the text is to the right of the icon.

jkeck:
I have found myself getting away from putting things in `lib` as I have found it can cause more confusion than creating a clean delineation of what should go in `lib` vs. what should go in `app/models`.

If you can't live with this in `app/models` I'm happy to move it.

ndushay:
I just was wondering if general practice had changed;  I can live with it where it is.  It's by no means the only class in `app/models` with that status.

PR #660, opened 2016-11-16
disable the "Approve" button when it's clicked to approve a mediated request, re-enable the button if the approval fails

### LINKED ISSUE #603: Disable Approve button while waiting for ajax request to respond
- [ ] Disable the button (similarly to how it is disabled when the ajax requests responds).
- [ ] Re-enable the button if the ajax response comes back as an error.

jmartin-sul:
* item_approval.js:  add disableElement and enableElement methods, call on attempted approval and approval failure, respectively.
* mediation_table_spec.rb:  confirm that "Approve" button is in correct enabled/disabled state after it's clicked.
* item_approval_spec.js:  unit tests for enableElement and disableElement methods

closes #603

### DIFF
--- a/app/assets/javascripts/item_approval.js
+++ b/app/assets/javascripts/item_approval.js
@@ -31,6 +31,8 @@ var itemApproval = (function() {
     approveItem: function(item) {
       var _this = this;
       var url = item.data('item-approval-url');
+
+      _this.disableElement(item);
       $.ajax(url).success(function(data) {
          _this.markRowAsApproved(item);
          _this.updateApproverInformation(item, data);
@@ -41,7 +43,16 @@ var itemApproval = (function() {
            _this.markRowAsError(item);
          }
          _this.addHoldingsLevelAlert(item, data);
+         _this.enableElement(item);
        });
+    },
+
+    disableElement: function(elt) {
+      elt.attr('disabled', 'disabled');
+    },
+
+    enableElement: function(elt) {
+      elt.removeAttr('disabled');
     },
 
     markRowAsApproved: function(item) {
--- a/spec/features/mediation_table_spec.rb
+++ b/spec/features/mediation_table_spec.rb
@@ -117,6 +117,7 @@ describe 'Mediation table', js: true do
           end
           expect(page).to have_css('tr.approved')
           expect(page).to have_css('td button', text: 'Approved')
+          expect(page).to have_button('Approve', disabled: true)
 
           within(first('tr')) do
             expect(page).to have_css('td', text: 'Added to pick list', visible: true)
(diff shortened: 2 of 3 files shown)


spec/features/mediation_table_spec.rb
```diff
@@ -197,6 +200,8 @@
         within('tbody td table tbody') do
           within(all('tr').last) do
             click_button('Approve')
+            expect(page).to have_button('Approve')
+            expect(page).to_not have_button('Approve', disabled: true)
```
jmartin-sul: since the `disabled` attribute is removed to re-enable the button, i couldn't check for `disabled: false`, and instead checked for the presence of the button and the absence of the attribute.  originally, i did try checking `disabled: false`, since the attribute value in the HTML is actually set to `disabled`, and not a boolean value, but that caused a test failure, whereas this approach didn't, and manual testing indicated that re-enabling worked fine.

spec/javascripts/item_approval_spec.js
```diff
@@ -46,4 +46,20 @@ describe('Item Approval', function() {
...
+      expect(lastButton.prop('disabled')).toBe(true);
+    })
+  });
+
+  describe('enableElement()', function() {
+    it('removes the disabled attr from the button', function() {
+      var lastButton = $('button:last');
+      itemApproval.enableElement(lastButton);
+      expect(lastButton.prop('disabled')).toBe(false);
+    })
```
jmartin-sul: teaspoon, however, always gives you the disabled status as a boolean, even if the attribute is no longer present.

app/assets/javascripts/item_approval.js
```diff
@@ -39,11 +41,20 @@ var itemApproval = (function() {
          if (data.responseJSON) {
            _this.addItemErrorInformation(item, data.responseJSON);
            _this.markRowAsError(item);
+           _this.enableElement(item);
```
jkeck: I wonder if we should enable the button regardless of if we get `responseJSON`.

spec/features/mediation_table_spec.rb
jkeck: One thought is we could find the button in the page (using `find` or `all` with a css selector) and then make assertions on the `disabled` attribute itself.  I have used this strategy when testing for `disabled` elements in the past as it can be kind of a PITA.

I would say only if you think it's worth it to be more explicit about what we're testing.  Otherwise, perhaps a brief comment in here explaining what this is actually asserting.

app/assets/javascripts/item_approval.js
jmartin-sul: ah, yeah, probably.  should i go ahead and change that?

jmartin-sul:
@jkeck: pushed with suggested changes, ready for re-review...

jmartin-sul:
added `wait_for_ajax` before the two button checks that were failing under ruby 2.2.1, and travis seems to be happy again.

PR #661, opened 2016-11-16
[WIP] Add 722 as a successful response code from symphony.

### LINKED ISSUE #531: Ignore some Symphony error responses  (like 722)
Any Symphony condition that results in the user having a hold on the material should be accepted as a success. Specifically:
- [ ]  "(User already has a hold on this material)" should not generate an alert to the user
- [ ]  are there more?

Related to #423

jkeck:
Closes #531

to remove WIP, coordinate with systems (darsi) ... she needs to get her code into the appropriate symphony instance.  to trigger 722 error, make the same request twice.

to deploy to production, need to coord with darsi, as she needs to time her deployment to production. 

_**[WIP]** until we can get this deployed to stage and tested to ensure the behavior is what we want._

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -163,7 +163,7 @@ module SULRequests
       }
     }
 
-    config.symphony_success_codes = ['209', 'S001', 'P001', 'P001B', 'P002', 'P005']
+    config.symphony_success_codes = ['209', '722', 'S001', 'P001', 'P001B', 'P002', 'P005']
 
     config.no_user_privs_codes = ['U003', 'U004']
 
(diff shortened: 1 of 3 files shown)


ndushay:
Darsi's fix is on  sulohsdev1 blah blah blah ,  so I have deployed this branch to requests-dev for vetting by whomever.

ndushay:
@dlrueda will deploy her Symphony side changes to production RSN, and so this will be merged to master and intended for deployment with sidekiq work (but if deployment happens any time after Darsi's update to Symphony prod, it's fine.)

PR #672, opened 2016-11-22
Warn user immediately when an invalid date is selected

### LINKED ISSUE #651: Help text being treated as error text on date select; error text missing
If I choose a date during winter closure, I get this effect:

screen shot 2016-11-14 at 4 30 01 pm

which is confusing because it doesn't explain the actual error situation. I think we used to have text that said something like "The library is closed on that date." 

Can we get something closer to this instead?

screen shot 2016-11-14 at 4 35 41 pm

tingulfsen:
Users are presented with date selector to notify the library when they want to access certain materials (e.g. special collections). The current approach relies exclusively on the Bootstrap validation tool, which only pops up a message when the entire form is submitted. This PR adds code to immediately display a warning below the date selector when an invalid date is chosen.

Closes #651 

After:

screen shot 2016-11-22 at 12 16 06 am

### DIFF
--- a/app/assets/javascripts/date_selector.js
+++ b/app/assets/javascripts/date_selector.js
@@ -2,7 +2,8 @@ var dateSelector = (function() {
   var defaultOptions = {
     mediatedPageFieldSelector: 'input[data-request-type="mediated_page"]',
     dropdownSelector: '[data-paging-schedule-updater="true"]',
-    singleLibrarySelector: '[data-single-library-value]'
+    singleLibrarySelector: '[data-single-library-value]',
+    neededDateWarningSelector: '#needed-date-warning'
   };
 
   return {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/date_selector.js
```diff
@@ -80,19 +81,24 @@ var dateSelector = (function() {
       this.mediatedDateField().closest('.form-group').removeClass('has-error');
       this.mediatedDateField()[0].setCustomValidity('');
       this.mediatedDateField().siblings('.help-block').remove();
+      this.mediatedDateWarning().text('');
```
ndushay: i read this as `mediatedDateDrowning`  💧

app/views/requests/_needed_date.html.erb
```diff
@@ -3,8 +3,9 @@
     <%= f.label :needed_date, class: "#{label_column_class} required control-label" %>
     <div class='<%= content_column_class %>'>
       <%= f.date_field_without_bootstrap :needed_date, class: 'form-control', required: true, min: Time.zone.today, data: { request_type: current_request.type.underscore } %>
+      <div id='needed-date-warning' class='has-error needed-date-help-block'></div>
```
ndushay: should this be inside of an if clause so we only get the div when the needed date is bad?

app/assets/javascripts/date_selector.js
```diff
@@ -80,19 +81,24 @@ var dateSelector = (function() {
...
     },
 
     setDateFieldAsInvalid: function() {
+      var message = 'This library is not open on ' + this.selectedDate();
+      this.mediatedDateField()[0].setCustomValidity(message);
+      this.mediatedDateWarning().text(message);
```
ndushay: why do we need two ways of using the same message?

jvine:
Looked at this on stage

> 1. does it leave html for valid dates alone?

Appears to, yes. Looks correct when I select a valid date, and removes the error message when i select a valid date after selecting an invalid date.

> 2. is there ever any customValidity text?

Nope.

ndushay:
It seems to allow you to submit a request when the date is invalid.  that is perhaps undesirable.

jvine:
Clicking the submit button pops up a message that prevents submit. I'm okay with this - I think it's what we were doing previously.

ndushay:
so it turns out that the warning message appears as a ?flash?  alert  thanks to bootstrap, so all of my concerns have been addressed.  merging.

PR #673, opened 2016-11-28
Make the feedback link have a blank target…

### LINKED ISSUE #663: Make feedback link open in a new tab

jkeck:
…(as well as the appropriate rel attr).

Closes #663


jmartin-sul:
merged because hound was taking forever, and the other passing builds meant that rubocop passed, so seemed fine to just go ahead and pull the trigger.

PR #677, opened 2016-11-29
send "approval status" email after async call to symphony 

### LINKED ISSUE #586: Change text on confirmation email for move to background workers
Will we need to make changes to the text on the confirmation email?  Should the email wait for the status of the symphony request before being sent, or send the email immediately once we receive the user's request?

If we want to wait for symphony's response, we should probably change the text to include information on the status of the request.

ndushay:
For pages and hold/recalls, we will send an email to the user about their request after the asynchronous call to symphony returns.   (For scans and mediated pages, we will stick with the existing mailer as the information (scans need a user's authenticated browser to redirect thru illiad;  mediated pages need the  approval process to work as is-- the user request for a mediated page doesn't hit symphony)

Per ticket #586, also tweaks the existing synchronous confirmation mailer slightly/

closes #586


ndushay:
whoops - wrong branch!

PR #678, opened 2016-11-29
send "approval status" email after async call to symphony

### LINKED ISSUE #586: Change text on confirmation email for move to background workers

ndushay:
For pages and hold/recalls, we will send an email to the user about their request after the asynchronous call to symphony returns. (For scans and mediated pages, we will stick with the existing mailer as the information (scans need a user's authenticated browser to redirect thru illiad; mediated pages need the approval process to work as is-- the user request for a mediated page doesn't hit symphony)

warning:  shameless green!

Per ticket #586, also tweaks the existing synchronous confirmation mailer slightly

connects to #586

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -82,7 +82,7 @@ Rails.application.configure do
   # put IP at beginning of log messages
   config.log_tags = [ :remote_ip ]
   # reduce noise in logs
-  config.lograge.enabled = true
+  #config.lograge.enabled = true
 
   # Use the Sidekiq adapter for Active Job if configured in settings
   if Settings.background_jobs == true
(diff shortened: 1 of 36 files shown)


config/environments/production.rb
```diff
@@ -85,7 +85,7 @@
   #config.lograge.enabled = true
 
   # Use the Sidekiq adapter for Active Job if configured in settings
-  if Settings.background_jobs == true
+  if Settings.background_jobs.enabled == true
```
jmartin-sul: FWIW: `Settings.background_jobs` isn't in the default `settings.yml`.  so, for safety's sake, it seems like the above should be `if Settings.background_jobs && Settings.background_jobs.enabled == true`, or the default `settings.yml` should have a `background_jobs` section with `enabled` set to `false`.

apologies if you're already on that, i know this was an in-the-middle-of-testing fix, but figured it was worth pointing out just in case...

Issue #681, opened 2016-11-30
Multiple item approvals overriding each other


jkeck:
If you click on multiple `Approve` buttons at once each time symphony returns the previous statuses get overwritten. 

[This code](https://github.com/sul-dlss/sul-requests/blob/sidekiq/app/jobs/submit_symphony_request_job.rb#L16-L17) is what is expected to handle that, but it's not capturing for this case.  We may need to do a `reload` on the request before merging in the symphony response, but since we've moved over to passing the ID to the job and calling find within, I figured that wouldn't be a problem (although maybe with long running symphony responses a reload would still be necessary).

ndushay:
possibly relevant:

https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#cannot-find-modelname-with-id12345


https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting
> Sidekiq is multithreaded so your Workers must be thread-safe.

therubyracer cowboyd/therubyracer/issues/270 is not thread safe

ndushay:
@cbeer suggests another approach would be to lock the request record in the `perform` method of the job

PR #685, opened 2016-12-02
[WIP] Address overwriting request status during item approval

jkeck:
Potentially addresses #681 

We have initially come up with two potential solutions for this problem:

1. This is the simplest (and most likely least performant) solution which is to lock the row when updating the item approval and symphony request status.  Since the row is being locked for update, it does cause the record to no be able to be update asynchronously (and if something were to run very long could cause lock timeout errors).

1. The other (read: correct) approach would be to refactor the serialized data structors that hold the symphony response and the item approval status into their own active record classes.  The affect that this will have will be that instead of needing to lock the request row in the database for updating, we'll instead be simply adding new rows to another table (thus not requiring any locking and being able to be completely asynchronous).

Approach 1 is what this PR does (and appears to work as expected with a few requests in stage).  Approach 2 will be more performant, and is clearly the correct way to structure this kind of information for clarity and performance.

Let's chat about this before we merge it in.

### DIFF
--- a/app/models/item_status.rb
+++ b/app/models/item_status.rb
@@ -52,15 +52,17 @@ class ItemStatus
 
   def approve!(user)
     @request.send_to_symphony_now!(barcodes: [@id])
-    reload_request # reloading to get any attributes saved to the database above
-    return unless symphony_item_successful?
-    self.status_object = {
-      approved: true,
-      approval_time: Time.zone.now.to_s,
-      approver: user
-    }.with_indifferent_access
-    @request.approval_status = :approved if @request.all_approved?
-    @request.save!
+    @request.with_lock do
+      @request.request_status_data ||= {}
+      next unless symphony_item_successful?
+      self.status_object = {
+        approved: true,
+        approval_time: Time.zone.now.to_s,
+        approver: user
+      }.with_indifferent_access
+      @request.approval_status = :approved if @request.all_approved?
+      @request.save
+    end
   end
 
   private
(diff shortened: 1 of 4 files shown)


app/models/item_status.rb
```diff
@@ -52,15 +52,17 @@ def approval_time
...
-    self.status_object = {
-      approved: true,
-      approval_time: Time.zone.now.to_s,
-      approver: user
-    }.with_indifferent_access
-    @request.approval_status = :approved if @request.all_approved?
-    @request.save!
+    @request.with_lock do
+      @request.request_status_data ||= {}
+      next unless symphony_item_successful?
```
ndushay: curious about using `next` here ... is that to cleanly get out of the lock?

spec/models/item_status_spec.rb
```diff
@@ -3,7 +3,7 @@
 def update_symphony_data_and_save(request, symphony_data)
   r = Request.find(request.id)
   r.symphony_response_data = symphony_data
-  r.save!
+  r.save
```
ndushay: this change is to avoid errors being raised?

app/models/item_status.rb
jkeck: Yes, it's to get out of the lock (it used to `return`, but I don't believe that works the same in this case).

spec/models/item_status_spec.rb
jkeck: It's to avoid validation, which can be a bit timely (I believe due to getting current location from the SW API).

PR #693, opened 2016-12-14
Add new Science library and remove math, bio, and chem.

jkeck:
I'm leaving the translations in so that old requests for these libraries will have the correct name.


jkeck:
🚢 ing because this needs to get deployed tonight. It's basically configuration and test changes.

PR #698, opened 2017-01-21
Modify Paging schedule for SAL3 SCAN

### LINKED ISSUE #697: Modify Paging schedule for SAL3 SCAN
Scanning staff in ILL have settled on a more realistic schedule for processing requested scans from SAL3, and they've requested we update the published "earliest delivery" time to users.

Here's the new SAL3 SCAN schedule:
When paging to SCAN before 12:00pm the item will arrive after 6:00pm 1 days later
When paging to SCAN after 12:00pm the item will arrive after 6:00pm 2 days later

This schedule can go into production anytime that is convenient.

cbeer:
Fixes #697

### DIFF
--- a/config/initializers/paging_schedule.rb
+++ b/config/initializers/paging_schedule.rb
@@ -10,12 +10,12 @@ PagingSchedule.configure do
   end
   # Scan (based on SAL3 -> GREEN with added days)
   when_paging from: 'SAL3', to: 'SCAN', before: '12:00pm' do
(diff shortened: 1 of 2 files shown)


cbeer:
@jkeck I guess we made a test that assumes the worst-case time.. do you think we should fix the test, or hard-code the worst-case time to make the test pass?

jkeck:
I think fixing the test will be 👌

We can address it differently down the road if it proves to be a persistent problem.

PR #717, opened 2017-06-15
Prevent item selection after reaching max limit

### LINKED ISSUE #715: Max items limit is not actually being applied
Special Collections requests have a maximum of 5 items that can be requested at one time. The UI isn't working correctly here. Select 5 items, and the alert appears - good. But then if you continue clicking in the list - the checkbox will not be checked, the row will not be highlighted, but the items will be added to the list of selected items below the scrolling box, and the items are added to the actual request. 

Example of a request that got more than 5:  https://requests.stanford.edu/mediated_pages/102851/status

A sample record with lots of items:
https://searchworks.stanford.edu/view/4085328

screen shot 2017-06-14 at 10 43 12 am

Intended behaviour once user reaches 5 items:
- display the alert
- prevent further selections in the scrolling list
- do not add any further items to the selected list

camillevilla:
Closes #715

pending: code review and discussion about search bar bug

## Before
- Input couldn't be checked off but were still active and clickable

## After 
- sets unchecked inputs to `disabled="true"` after max limit is reached
sul_requests

### DIFF
--- a/app/assets/javascripts/item_selector/item_selector_limit.js
+++ b/app/assets/javascripts/item_selector/item_selector_limit.js
@@ -23,13 +23,29 @@ var itemSelectorLimit = (function() {
 
     setupEventListeners: function() {
       var _this = this;
+      _this.searchListener();
       _this.selectorElement().on('item-selector:selected', function(_, item) {
         _this.increaseSelectedNumber();
         _this.enforceSelectedItemLimit(item);
       });
 
-      _this.selectorElement().on('item-selector:deselected', function() {
+      _this.selectorElement().on('item-selector:deselected', function(_, item) {
         _this.decreaseSelectedNumber();
+        _this.enforceSelectedItemLimit(item);
+      });
+    },
+
+    searchListener: function(){
+      var _this = this;
+      var limit = selectorLimit(_this);
+      $('#item-selector-search').on('blur', function(){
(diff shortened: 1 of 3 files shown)


camillevilla:
Working on a related bug where users can select 1 additional item per search context after reaching the max item limit

item_selector_search_bug

camillevilla:
Alrighty, @jkeck. I cleaned up my commits and have feature tests and JS unit tests. I have one more bit of behavior to fix. 

A user should:
- start a search and select their 5th item in that search facet
- be unable to select additional items in the current facet
- clear out the search bar and be unable to select additional items in the original context

Current behavior:
- checkboxes are not disabled until a user must enters an additional keystroke (non-backspace) 

remaining_search

camillevilla:
@jvine I finally have a fix for the bug mentioned above. After consulting with @jkeck, we decided the fix is not ideal behavior, but it's functional: the user can't add more than 5 items.

## Behavior
Checkboxes newly revealed by search still white and look selectable. Once the user tries to click on an item, the `disabled` attribute is triggered. This initial transition might be a little confusing for users, but the message about maximum items is still there to clarify what happened. 

## Demo
search_blur_fix

jvine:
@camillevilla Can you check the behaviour with keyboard only? If you are in the search box, and reveal some new apparently-checkable checkboxes, what happens when you try to tab to them?

PR #736, opened 2017-09-18
Update to rails 4.2.9.

jkeck:


greysteil:
@jkeck - forgive me if I've asked before, but would you be up for automatically creating PRs like this one? I built [Dependabot](https://dependabot.com) to do that, and it's totally free for open source. Trying to get as many people using it as possible so as the [compatibility scores](https://dependabot.com/compatibility-score.html) are accurate!

PR #739, opened 2017-12-14
Add a flash notice about the symphony outage on Dec. 28, 2017.

jkeck:


jkeck:
The failures in this build are to be expected since we're effectively turning off much of the request app's functionality.  The PR to remove the notice shows that the tests pass again once the request function is enabled again.

PR #747, opened 2018-03-22
Rails 4 --> 5

cbeer:


jkeck:
I validated the token encrypt issue locally.  I created a token for a request under Rails 4.2 and Ruby 2.3 (what's currently in production).  I then switched to this branch under ruby 2.5 and passed that token to our `Request#valid_token?` implementation and in returned `true`, so 👍.

PR #796, opened 2018-11-12
Add SidekiqLatencyCheck

### LINKED ISSUE #795: Add some sort of status check for background queues
It would be nice to be alerted if our app does not have any queues available to it.  This could be done through the Sidekiq API, but we can explore other mechanisms to be alerted when we our background jobs infrastructure isn't picking up any requests to be sent out.

cbeer:
Fixes #795?

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -6,7 +6,7 @@ OkComputer.check_in_parallel = true
 
 OkComputer::Registry.register 'confirmation_mailer', OkComputer::ActionMailerCheck.new(ConfirmationMailer)
 OkComputer::Registry.register 'searchworks_api', OkComputer::HttpCheck.new("#{Settings.searchworks_api}/status")
-
+OkComputer::Registry.register 'background_jobs', OkComputer::SidekiqLatencyCheck.new('default', 25)
 
 if Settings.symphony_api.enabled && Settings.symphony_api.url.present?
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -6,7 +6,7 @@
 
 OkComputer::Registry.register 'confirmation_mailer', OkComputer::ActionMailerCheck.new(ConfirmationMailer)
 OkComputer::Registry.register 'searchworks_api', OkComputer::HttpCheck.new("#{Settings.searchworks_api}/status")
-
+OkComputer::Registry.register 'background_jobs', OkComputer::SidekiqLatencyCheck.new('default', 25)
```
jkeck: Would this potentially give us false alerts if we're in a lull in requests?  (not sure what their def. of latency is here).

cbeer: Nope; from the docs:

> If you throw a lot of jobs into the queue, you can get false positives when monitoring the queue backlog. Instead, monitor the queue latency. Queue latency is the difference between when the oldest job was pushed onto the queue versus the current time. This code will check that jobs don't spend more than 30 seconds enqueued. 

https://github.com/mperham/sidekiq/wiki/Monitoring#monitoring-queue-latency

--

This is also the same check we're using on exhibits (and maybe other places?)

jkeck: 👍

PR #797, opened 2018-11-16
Implement a BorrowDirect find and request before submitting a Hold/Recall to Symphony

### LINKED ISSUE #794: Implement Borrow Direct in the Hold/Recall Workflow (ship after exams)
In order to help relieve the excessive recall issues, we are going to integrate with Borrow Direct to request items that we don't otherwise have available to our users on-campus.

Before submitting to Symphony we should try to submit the request via BorrowDirect (through their API).  We will most likely want to do this through ISBN, although they also accept OCLC and ISSN.

If the BorrowDirect request succeeds, then the user will get the item through them (and we do not need to submit the request to Symphony).  If the BorrowDirect request fails then we'll do what we had always done, which is to submit the request to Symphony.  

Their API should fail for one of 3 reasons.
- Their API isn't responding / our implementation is broken
- The do not have the item
- They believe the item is otherwise available at our institution.

Some changes will be made on the Symphony side to make this kind of request a title level hold, and can have some intelligence locally about if an item is a volume vs. copy and can ensure the appropriate item is either added to a pick list or recalled from a user (as a last ditch effort).

There are also some changes to the Status page, which are detailed in the doc linked below.

[There is a BorrowDirect gem](https://github.com/jrochkind/borrow_direct) that will make dealing with the API much easier.  @jkeck has our API key.

A few more details/discussion in [a Google Doc in our Team Drive](https://docs.google.com/document/d/1Twqv5us9l2G7VHZvpg_Trwip8psK-ATFrPtligAYMqA/edit)

## Tasks
- [x] Make request to BorrowDirect API (as described above)
- [x] Update status page with some text indicating that we're working on getting the item (described in Google Doc)
- [x] Update the status page for admins to include the BorrowDirect API response
- [x] Figure out how to send the appropriate pickup library data to the item request API

jkeck:
Closes #794


db/migrate/20181115194822_add_borrow_direct_boolean_column_to_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddBorrowDirectBooleanColumnToRequests < ActiveRecord::Migration
+  def change
+    add_column :requests, :via_borrow_direct, :boolean, default: false
```
jkeck: I've added this migration in a separate commit, and am totally willing to remove it.

At first, I was thinking that there would be state where we might want to set the via_borrow_direct flag at some point before making the official request to them, but it's all pretty much at the same time as far as the user will be concerned, so we could theoretically infer this from the existence of any `borrow_direct_response_data`.

PR #809, opened 2018-12-10
Handle radio button item selection properly now that we're in rails 5…

jkeck:
… land.


spec/controllers/scans_controller_spec.rb
```diff
@@ -102,7 +102,6 @@
 
       before do
         stub_searchworks_api_json(build(:sal3_holdings))
-        allow(controller).to receive(:current_request).and_return(create(:scan_with_holdings, barcodes: ['12345678']))
```
jkeck: This is what prevented us from identifying this issue during the upgrade.

app/controllers/requests_controller.rb
```diff
@@ -115,12 +115,25 @@ def delegated_new_request_path(request, url_params = nil)
...
 
+  # Radio buttons get sent in as an array (radios are used when only a single item is allowed from a mutli-volume set)
+  # This will modify the barcodes array to be a hash that mimics checkbox elements (for strong params purposes).
+  # This worked pre-rails 5 but now we're restricted to only passing one parameter format.
+  # modify_item_selector_checkboxes will change this back into an array (which is dumb) to be persisted, but this is
+  # seemingly the only way to get Strong Params + AR Serializer + Checkbox/Radio agnostic form element to work.
+  def modify_item_selector_radio_buttons
+    return unless params[:request][:barcodes].is_a?(Array) && params[:request][:barcodes].one?
+
+    create_params[:barcodes] = params[:request][:barcodes].map { |barcode| [barcode, '1'] }.to_h
```
cbeer: I think I'm confused what the contract is.. we're turning radio buttons into a hash, and we're turning checkboxes into an array?

PR #814, opened 2018-12-14
render page range content as html_safe fixes #813

### LINKED ISSUE #813: SCAN email confirmation showing html in Page range
The confirmation email message for SCAN requests is showing html in the "Page range" section of the email body.  Here's an example:

image

mejackreed:
Fixes #813

### DIFF
--- a/app/views/shared/_request_status_information.html.erb
+++ b/app/views/shared/_request_status_information.html.erb
@@ -41,7 +41,7 @@
   <% if current_request.data.present? %>
     <% if (page_range = current_request.data['page_range']).present? %>
       <dt><%= current_request.class.human_attribute_name(:page_range) %></dt>
-      <dd><%= page_range %></dd>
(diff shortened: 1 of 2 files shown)


jkeck:
Was the issue here that a user supplied HTML in their comment and we want to render that HTML on the status page side? (which could potentially be a bad thing)

Also, I think the original ticket mentions the content being in the email output, which I'm not sure this addresses.

camillevilla:
It would be good to pair on this and conduct some tests in stage to verify behavior. My understanding was that HTML was being produced even though the user was supplying plain text.

mejackreed:
Yep seems like a good thing to pair on. I'm not sure I even really understand what is happening in requests, so open to suggestions/pairs

mejackreed:
Closing this, turns out this doesn't really do what I thought it did.

PR #834, opened 2019-02-21
Access Services requests more detail about type of request we are "Sorry" about

jgreben:
Per Sarah Seestone at out Circ/LibSys meeting, they need the user to have more info on the page about the kind of request when they get to the `sorry/unable` page. Currently the page renders:

screen shot 2019-02-20 at 5 14 35 pm

This PR allows Sorry controller to accept a param that will display the request type on the rendered page.

AFAIKT Scans are the only type of request that end up calling `sorry/unable` (via ILLiad) so maybe this is a bit overkill, but if we ever use it for other kinds of requests, it might come in handy. The idea is to send the redirect back from ILLiad as `sorry/unable?request_type=scan` if the request fails in ILLiad. Then the page would render:

screen shot 2019-02-20 at 5 03 45 pm

### DIFF
--- a/spec/support/factory_bot.rb
+++ b/spec/support/factory_bot.rb
@@ -12,10 +12,8 @@ RSpec.configure do |config|
       factory.name =~ /_holdings?$/ || factory.name =~ /_searchworks_item$/ || factory.name =~ /^symphony_/
     end
 
-    begin
-      FactoryBot.lint factories_to_lint unless config.files_to_run.one?
-    ensure
-      DatabaseCleaner.clean
-    end
+    FactoryBot.lint factories_to_lint unless config.files_to_run.one?
+  ensure
+    DatabaseCleaner.clean
   end
 end
--- a/spec/support/factory_bot.rb
+++ b/spec/support/factory_bot.rb
@@ -12,8 +12,11 @@ RSpec.configure do |config|
       factory.name =~ /_holdings?$/ || factory.name =~ /_searchworks_item$/ || factory.name =~ /^symphony_/
     end
 
-    FactoryBot.lint factories_to_lint unless config.files_to_run.one?
-  ensure
-    DatabaseCleaner.clean
+    begin
+      FactoryBot.lint factories_to_lint unless config.files_to_run.one?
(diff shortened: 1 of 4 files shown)


spec/support/factory_bot.rb
```diff
@@ -12,8 +12,10 @@
       factory.name =~ /_holdings?$/ || factory.name =~ /_searchworks_item$/ || factory.name =~ /^symphony_/
     end
 
-    FactoryBot.lint factories_to_lint unless config.files_to_run.one?
-  ensure
-    DatabaseCleaner.clean
+    begin
+      FactoryBot.lint factories_to_lint unless config.files_to_run.one?
```
jkeck: This effectively makes our factory linting completely useless by wrapping in a `begin`.

jgreben: Why do I get `SyntaxError:
  /Users/jgreben/Projects/sul-requests/spec/support/factory_bot.rb:16: syntax error, unexpected keyword_ensure, expecting keyword_end` when running rspec with the original file?

jgreben: Thanks, this passes on travis. My local system needs updating to catch up with this project.

jkeck:
This looks good to me @jgreben.  One last question I have, is what kinds of values are we expecting to come through the `request_type` parameter?  Is this going to be a term that is in the SUL Requests domain or something that is just going to be a term specifically from the Illiad domain?

jgreben:
@jkeck It may be a term in the SUL Requests domain (perhaps just "scan" as in the example) but @saseestone  may want to provide something more descriptive like "SAL3 scan to PDF". I think ILLiad is the only service that currently uses this page. Maybe in the future it could be used by other request types within SUL Requests as well.

jkeck:
Thanks for the context.

If it's going to be a static string, perhaps we should just hard-code it until this page is used by another request type (and we can add comments about that for future edification).

jgreben:
I suppose so...but I was trying to be so clever! Ok I will close this PR and wait to find out exactly what they want the sorry page to say.

PR #841, opened 2019-03-26
Update the contact info and information on the sorry/unable page

jgreben:
The scans processing team should be the contact point for failed scans. Right now the only type of requests that use the sorry/unable page are scans.


jgreben:
Thanks @jkeck, no rush. It can go out on Monday.

PR #879, opened 2019-08-06
Update SU footer

mejackreed:
The mobile view on master seems to be jacked up a bit already. This doesn't fix it.


cbeer:
This has the same SUL logo bottom border issue.

PR #893, opened 2019-10-07
Update dependencies

sul-devops-team:


Gemfile.lock
```diff
@@ -339,16 +336,17 @@ GEM
...
-      concurrent-ruby (~> 1.0)
-      connection_pool (~> 2.2, >= 2.2.0)
-      rack-protection (>= 1.5.0)
-      redis (~> 3.2, >= 3.2.1)
-    sidekiq-statistic (1.2.0)
-      sidekiq (>= 3.3.4, < 5)
+    selenium-webdriver (3.142.6)
+      childprocess (>= 0.5, < 4.0)
+      rubyzip (>= 1.2.2)
+    sidekiq (6.0.1)
```
mejackreed: Not sure why sidekiq didn't get pinned here, but should we pin >6 ?

cbeer: < 6? Yes, probably. At least so we can defer figuring out how to do daemon stuff until we're on a modern OS.

mejackreed: Yes def < 6 😆

PR #903, opened 2019-10-23
Validate library ID (and make Name/Email required in some cases)

### LINKED ISSUE #896: LibID requests: require name/email when libID is optional
Part of #895 

In requests that can be submitted with name/email only; where we currently display this when the user selects "I don't have a SUNet ID":

request - optional current

Change the UI as follows:

- add help text below the libID input
   - `ID is a 10-digit number. It's optional, but may help you track your request in My Library Account.`
- add an `aria-describedby` connection between the input and the help text
- add `required` attribute to Name and Email inputs
- add `required` class to the labels for Name and Email

request - id optional final

### LINKED ISSUE #897: LibID requests: input validation on the client side
Part of #895 

Client-side validation may help prevent users from entering IDs from other universities, or catch missed digits. 

1. Add help text below the input. 
   - if LibID is required:  `ID is a 10-digit number.`
   - if LibID is optional:  `ID is a 10-digit number. It's optional, but may help you track your request in My Library Account.`
   - associate help text with the input using `aria-describedby`

2. Can we require 10 characters (possibly `minlength="10" maxlength="10"`? or is there a more graceful way?)
   - if we use `minlength`, I'd like to do something like `setCustomValidity()` to override the validation message to "Stanford Library ID must have 10 digits (you have entered 9)." because the default message is very default-y.
     request - length validation

Note that the "10-digit number" help text is for patron information only, but can't be enforced on the input. Pseudo-patron IDs will often (always?) include other characters, but they are only used by staff. Actual patron IDs are always numbers. 

I don't think there's anything else that we can consistently assert about the ID format.

### LINKED ISSUE #898: LibID requests: Validate libID before submitting request
Part of #895 

This is important for cases where the library ID is required to complete the request, where we currently display this when the user selects "I don't have a SUNet ID":

request - required current

Overall process:
1. User enters LibID and clicks Send
2. We check whether the LibID as entered exists in Symphony
3. If it does, we process the request as normal, and proceed to the "we got your request" page
4. If it does not exist, we display the error in place in the form, so the user can correct it and re-submit.

At the point where the user clicks the `Send request` button, we currently disable the button, and there's a pause before the "confirmation" page displays.

### Add a spinner to the button
Since the check might cause it to take more time than the usual Send.

request - spinner

### Display the error condition on the form

If the libID comes back as invalid:

- Change the `h1` to `Request not sent: Library ID is incorrect` **(1)**
- Add a cardinal-red border around the libID input **(2)**
- Add the error message immediately below the input (replaces the help block)  **(3)**
   - in `<div role="alert">` so it will announce to the SR 
   - use class `alert-danger` for styling 
   - add danger icon in front of the first line of text (I used `<p>` to separate the two sections of text.) **(4)**
   - use `aria-describedby` to link the error to the input 
   - add an `aria-label` for the `8 0 6 0 9` number so it reads as separate digits for SR: **(5)**
      - `<span aria-label="8 0 6 0 9">80609</span>`
      - style the span with `font-weight: bold; letter-spacing: .1em;`

request - error message

jkeck:
Closes #896 
Closes #897
Closes #898

## Library ID Is Too Short
Library ID Too short

## Library ID does not exist
Library ID Validation Error

### DIFF
--- a/app/views/requests/_no_sunet_form.html.erb
+++ b/app/views/requests/_no_sunet_form.html.erb
@@ -8,22 +8,32 @@
             <%= uf.text_field_without_bootstrap(
                   :library_id,
                   class: 'form-control',
+                  'aria-describedby' => 'library-id-help',
                   data: { behavior: 'single-user-field' },
                   autocomplete: 'off'
                 )
             %>
+            <p id="library-id-help" class="help-block">ID is a 10-digit number.
+              <% if f.object.requestable_by_all? %>
+                It's optional, but may help you track your request in My Library Account.
+              <% end %>
+            </p>
           </div>
         </div>
       <% end  %>
       <% if f.object.requestable_by_all? %>
-        <div class='form-group'>
-          <div class='<%= "#{label_column_offset_class} #{content_column_class}" %>'>
-            <h3>Don’t have a Stanford Library ID?</h3>
-            <p>Just enter your name and email.</p>
-          </div>
-        </div>
-        <%= uf.text_field :name, data: { behavior: 'grouped-user-field' } %>
-        <%= uf.email_field :email, data: { behavior: 'grouped-user-field' } %>
+        <%= uf.text_field(
+              :name,
+              label_class: 'required',
+              data: { behavior: 'grouped-user-field' }
+            )
+        %>
+        <%= uf.email_field(
+              :email,
+              label_class: 'required',
+              data: { behavior: 'grouped-user-field' }
+            )
+        %>
       <% end %>
     <% end %>
 
--- a/app/views/requests/_no_sunet_form.html.erb
+++ b/app/views/requests/_no_sunet_form.html.erb
@@ -8,6 +8,8 @@
             <%= uf.text_field_without_bootstrap(
                   :library_id,
                   class: 'form-control',
+                  minlength: (10 unless f.object.requestable_by_all?),
+                  maxlength: 10,
                   'aria-describedby' => 'library-id-help',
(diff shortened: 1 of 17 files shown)


app/views/requests/_no_sunet_form.html.erb
```diff
@@ -2,28 +2,49 @@
...
                 )
             %>
+            <% if f.object.library_id_error? %>
+              <div role="alert" id="library-id-help" class="alert alert-danger">
+                <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span>
+                <%= f.object.errors[:library_id].join('. ') %>.<br/>
+                A Stanford Library ID has 10 digits. On some University ID cards, it has a prefix of <span class="suid-prefix" aria-label="8 0 6 0 9">80609</span>.
+              </div>
+            <% else %>
+              <p id="library-id-help" class="help-block">
```
cbeer: I forget -- do we support entering a proxy borrowers ID in this field?

jkeck: From what I understand, anybody who requests items w/ an ID that is not this 10 digit number does not do so through this request form (e.g. through workflows directly, or some other mechanism)

jkeck: @saseestone/@jvine may have more details

jvine: Proxy borrower IDs should also be 10 digits. Pseudo-patron IDs may not be, but they can request through Workflows.

PR #921, opened 2020-02-01
Allow empty emails in the SymphonyUserNameRequest lookup.

jkeck:


spec/models/symphony_user_name_request_spec.rb
```diff
@@ -26,6 +26,22 @@
       end
 
       it { expect(subject.exists?).to be true }
+
+      context 'when there is only a name' do
+        let(:response) { double(success?: true, body: 'Some Patron ()') }
+
+        it { expect(subject.name).to eq 'Some Patron' }
+        it { expect(subject.email).to be_blank }
+        it { expect(subject.exists?).to be true }
```
jkeck: This is the case that was failing before this change (as the `#name` accessor also returned `nil`)

PR #933, opened 2020-03-11
Temporarily remove SAL (1/2) as a Pickup Library.

jkeck:


jkeck:
Merging w/ approval.  Coveralls is not going to finish.

Issue #932, opened 2020-03-09
Duplicate headings on messages dashboard


jkeck:
<img width="527" alt="Screen Shot 2020-03-09 at 1 26 16 PM" src="https://user-images.githubusercontent.com/96776/76254403-a1fa8900-6209-11ea-92ce-2f6d28022784.png">

Both `Add message` buttons go to the same place in this case.  Not sure if this is intentional or not (one of them will turn to an `Edit message` button when there is a message), but it can be a bit confusing why there are two things that do the same thing in this case.

jkeck:
Closed by #936

PR #936, opened 2020-03-12
Deal w/ "duplicate heading"  issue

jkeck:
This is an approach to handling the issue / confusion reported in #932 

## One Message (before)
one-message-after

## One Message (after)
one-message-before

## Empty Messages (before)
empty-message-before

## Empty Messages (after)
empty-message-after

## Another Example 
Trying to avoid this scenario where a user cannot edit or remove these messages (only add more)

cannot-delete

### DIFF
--- a/app/helpers/messages_helper.rb
+++ b/app/helpers/messages_helper.rb
@@ -25,4 +25,16 @@ module MessagesHelper
                     edit-message
                     glyphicon glyphicon-pencil'
   end
+
+  def link_to_delete_message(message)
+    link_to ' Delete message',
+            message,
+            method: :delete,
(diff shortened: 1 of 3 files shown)


jkeck:
@jvine / @saseestone thoughts?  Mainly trying to find a quick fix to this (plus, deleting is nice 🗑)

saseestone:
The approach looks good to me.  And deleting is nice.

To be honest, I had to sit and think about why we'd want the ability to have 2 different active messages, and realized that it would be handy if you need two different date ranges for the text.  Y'all are smart.  Now I can be smarter and more efficient too.

PR #957, opened 2020-05-19
Update URL path by removing v1

jgreben:
The `v1` part of the SymWS URL is deprecated and will be removed in the next version of SymWS.


jgreben:
Requires #958

jgreben:
Will make these in one pr.

PR #996, opened 2020-06-09
Ask which branch to deploy on dev/stage

jkeck:
This has burnt me for the last time

### DIFF
--- a/config/deploy/dev.rb
+++ b/config/deploy/dev.rb
@@ -1,5 +1,7 @@
 set :rails_env, 'development'
 
+ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call unless ENV['DEPLOY']
+
 server 'requests-dev.stanford.edu', user: 'requests', roles: %w(web db app)
 
 set :bundle_without, %w{deployment test}.join(' ')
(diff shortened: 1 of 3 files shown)


config/deploy/dev.rb
```diff
@@ -1,5 +1,7 @@
 set :rails_env, 'development'
 
+ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call unless ENV['DEPLOY']
```
jkeck, line 3: I suppose this is a proposal of how we might be able to handle both allowing deploying patches/production releases to go easily and to reduce the friction for getting development branches deployed to non-production servers.  I'm not strongly tied to this env var name if anybody has thoughts.  This would require that we use `DEPLOY=true` with our cap commands when doing our regular patching **IF** we want to avoid the branch ask (but then during normal development and deploying to non-production envs we get the prompt).

This also has the benefit of doing more to prevent you from deploying an unintended branch to production, which...uh....would never happen...right?

Happy to entertain other ideas, but the current pattern has bitten me on several occasions when needing to get branches deployed to dev/stage.

jkeck:
Duplicated by #1023

PR #1001, opened 2020-06-15
Update dependencies

sul-devops-team:


spec/features/admin_comment_spec.rb
```diff
@@ -25,9 +25,9 @@
       end
 
       within('.admin-comments') do
-        expect(page).to have_css('form#new_admin_comment', visible: false)
+        expect(page).to have_css('form#new_admin_comment', visible: :all)
```
jkeck, line 28: Just out of curiosity, did `visible: :hidden` work here by chance?  I know `:all` is functionally equivalent to `false`, but I think what we were trying to test for was `:hidden` in these cases.

cbeer: It wasn't clear to me offhand which of these tests were looking for exclusively hidden elements, so I just went with the previous behavior 🤷‍♂️ .

jkeck: I'll push something up shortly that will hopefully be more in-line with the intent.

spec/features/elibility_confirmation_spec.rb
```diff
@@ -11,27 +11,29 @@
     it 'shows an eligibility confiration overlay that can be cleared to reveal the form' do
       visit new_page_path(item_id: '1234', origin: 'GREEN', origin_location: 'STACKS')
 
-      expect(page).to have_css('#new_request', visible: false)
-      expect(page).to have_css('#eligibility-confirm-overlay', visible: true)
+      expect(page).to have_css('#new_request', obscured: true)
```
jkeck, line 14: This is something handy that I just learned about today (only available in the JS driver)

PR #1035, opened 2020-08-06
Co-opt the bootstrap info button for buttons to choose between scan …

### LINKED ISSUE #1033: Make Request and Scan button same color
When Scan is introduced as an option for users, make the "Request & pickup" and "Scan to PDF" buttons the same color so that they have the same weight as alternate options.

Jennifer and I landed on blue (like the blue background of the Request button in SearchWorks).

jkeck:
…and page.

Closes #1033 

Screen Shot 2020-08-06 at 9 32 09 AM

### DIFF
--- a/app/assets/stylesheets/sul-variables.scss
+++ b/app/assets/stylesheets/sul-variables.scss
@@ -6,6 +6,7 @@ $grayish-yellow: #d5d5d3;
 $green: #008000;
 $light-grayish-orange: #f0eeea;
 $light-grayish-yellow: #f7f7f4;
+$bright-blue: #006cb8;
 $pure-yellow: #ff0;
 $white: #fff;
 $dark-cyan: #007c5d;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/sul-variables.scss
```diff
@@ -6,6 +6,7 @@ $grayish-yellow: #d5d5d3;
 $green: #008000;
 $light-grayish-orange: #f0eeea;
 $light-grayish-yellow: #f7f7f4;
+$strong-blue: #006cb8;
```
cbeer, line 9:
```suggestion
$bright-blue: #006cb8;
```

That's what https://identity.stanford.edu/color.html#digital-color calls it, at least 🤷‍♂️

jkeck: Ah, I usually go by what [ColorHexa](https://www.colorhexa.com/006cb8) says, but we can use Stanford's made up color names. 😆

PR #1041, opened 2020-08-07
Symws

### LINKED ISSUE #1036: Consolidate item info lookup with symws
We're already making a symws call in `SymphonyCurrLocRequest`. After ###, we should consolidate that into the `SymphonyClient`

### LINKED ISSUE #1038: Pull in Symphony API client code from mylibrary
Whole-sale copy + paste, and strip out the calls we're not using (?) or include everything to maintain consistency?.

cbeer:
Fixes #1038
Fixes #1036


mejackreed:
Things to check:

 - [ ] No SCANs for group (not sure how to check this)

spec/jobs/submit_symphony_request_job_spec.rb
```diff
@@ -226,6 +226,59 @@
         subject.execute!
       end
 
+      context 'for a response where the user already has a hold on the material' do
```
mejackreed, line 229: 👏

jkeck (approved):

PR #1042, opened 2020-08-08
Enable scan requests for ETAS materials

### LINKED ISSUE #1027: Add knowledge of an item's ETAS status and disable Request & Pickup option
(Note, this will likely require a change in SearchWorks to provide this through the availability API)

Screen Shot 2020-07-27 at 11 35 46 AM

jkeck:
Closes #1027

## Request for record 12345

Screen Shot 2020-08-07 at 5 21 04 PM


jkeck:
This should be fine to ship now, we're not currently putting request links on any ETAS materials, but we would like to do that in the very near future (and this will make sure the request app will be ready to handle them by blocking non-scan requests).

PR #1082, opened 2020-09-17
Send mail to ILB staff if illiad request is not successful

### LINKED ISSUE #1081: Send email to Interlibrary Borrowing (ILB) Staff if ILLiad request fails
Sometimes we will get a `{ "Message": "...error..." }` response from the ILLiad API. When that happens no transaction will be recorded in ILLiad for a scan, but the item to scan may still be sent to ILB. We need to inform the staff so they can create a transaction in ILLIad manually.

Send an email to `stfborrowing@stanford.edu` and `sul-unicorn-devs@lists.stanford.edu` with the details of the `illiad_transaction_request` and the message from illiad (contained in the Scan's `illiad_response_data` field)

```
{
      'ProcessType': 'Borrowing',
      'RequestType': 'Article',
      'SpecIns': 'Scan and Deliver Request',
      'PhotoJournalTitle': scan_title,
      'PhotoArticleAuthor': @scan.authors,
      'Loaction': @scan.origin,
      'ReferenceNumber': @scan.origin_location,
      'PhotoArticleTitle': @scan.data[:section_title],
      'PhotoJournalInclusivePages': @scan.data[:page_range],
      'CallNumber': call_number,
      'ILLNumber': ill_number,
      'ItemNumber': item_number,
      'Username': @scan.user.webauth
    }
```

jgreben:
Fixes #1081 

- Creates a mailer for ILB staff
- Looks for an illiad "Messge" response in the scan_request_job, meaning there is some kind of error or other reason that the transaction could not be placed, otherwise the response would be the transaction details.
- Fixes a typo that is preventing the location information from being populated in the illiad transaction
- removes unused cruft from tests related to the old illiad request method
- removes old sorry/unable page, route, controller, and tests used by illiad prior to implementing the API. The new mailer will replace the need for the sorry page.

See also:
https://github.com/sul-dlss/shared_configs/pull/1550
https://github.com/sul-dlss/shared_configs/pull/1551
https://github.com/sul-dlss/shared_configs/pull/1552


app/jobs/submit_scan_request_job.rb
```diff
@@ -11,6 +11,8 @@ def perform(scan)
 
     scan.update(illiad_response_data: JSON.parse(response.body).select { |_, value| value.present? } || {})
 
+    scan.notify_ilb! if scan.illiad_response_data['Message']
```
jkeck, line 14: I'm wondering if we should add some sort of accessor into either `IlliadRequest` or the `Scan` class to let us know that the response was an error (so the concept of an Illiad error doesn't only exist in this job)

jgreben: Sounds good. I moved it to the Scan class

config/routes.rb
```diff
@@ -11,9 +11,6 @@
   Sidekiq::Web.set :session_secret, Rails.application.secrets[:secret_key_base]
   mount Sidekiq::Web => '/sidekiq', constraints: SidekiqConstraint.new
 
-  #sorry page route
-  get 'sorry/unable'
```
jkeck, line 15: 👏

jkeck (commented):
Just a few minor comments.  Thanks for tackling this.

config/settings/test.yml
```diff
@@ -25,6 +25,7 @@ symphony_web_services:
   curr_loc_path: 'v1/catalog/item/barcode/'
 
 sul_illiad: 'https://illiad-test/'
+illiad_email_address: 'stfborrowing@stanford.edu'
```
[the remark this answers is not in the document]
jgreben: Yeah, probably don't need this in the test.

jkeck (approved):
LGTM 👍   Thanks for making the changes.

jkeck:
@jgreben we're planning on rolling out some changes tomorrow, and am thinking about getting this shipped out along with the other symphony web services changes.

PR #1083, opened 2020-09-21
Update dependencies

sul-devops-team:


app/controllers/interstitial_controller.rb
```diff
@@ -23,8 +23,8 @@ def redirect_param_same_as_host?
 
     param_host = begin
                    URI.parse(URI.decode(params[:redirect_to]))
-                 rescue URI::InvalidURIError
-                   URI.parse(params[:redirect_to])
+    rescue URI::InvalidURIError
+      URI.parse(params[:redirect_to])
                  end
```
jkeck, line 28: Hmm, this doesn't quite look right, did the auto-fixer do something wonky here?

mejackreed: Yes this is what the autofix gave me. I've pushed up a change

jkeck (approved):

PR #1089, opened 2020-09-24
Look up patrons in symphony by webauth id or library id

cbeer:


app/models/patron.rb
```diff
@@ -5,24 +5,25 @@
...
+    patron_key ||= symphony_client.login_by_library_id(library_id)&.dig('key') if library_id
 
-    symphony_client = SymphonyClient.new
-
-    patron_key ||= symphony_client.login_by_library_id(library_id)&.dig('key')
-
-    return if patron_key.blank?
-
-    new(symphony_client.patron_info(patron_key))
+    new(symphony_client.patron_info(patron_key)) if patron_key.present?
```
jkeck, line 13: Do we want to notify Honeybadger or something like that if we run into the case of a patron w/o a key (to see how often this is actually happening)?

jkeck (approved):

PR #1114, opened 2020-10-22
Adds validation that page_range can only be up to 30 characrters

jgreben:
The illiad database record field length for PhotoJournalInclusivePages can currently only handle up to 30 chars. They will be fixing this sometime in the near-ish future, but in the meantime this will prevent failed illiad api transaction requests and the email being sent to ILB.

N.b. If we are OK with just getting the occasional (but sometimes daily) error we can consider foregoing this change, as we might then want to back it out when Atlas fixes the field length in the next version upgrade. I have no idea how long it will be before they release a new version upgrade.

### DIFF
--- a/app/models/requests/scan.rb
+++ b/app/models/requests/scan.rb
@@ -6,6 +6,7 @@
 class Scan < Request
   validate :scannable_validator
   validates :section_title, presence: true
+  validates :page_range, length: { maximum: 30 }
 
   def requestable_with_sunet_only?
     true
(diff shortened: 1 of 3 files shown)


jgreben:
Working with Atlas, we fixed the API error that happens when the page range field is more than 30 chars. However, it does truncate the field, so any extraneous information that the user enters here will be lost. In a different PR we should consider adding some kind of a note field because requestors often add additional information about the page range.

jgreben:
I recently upgraded to ILLiad 9.2 and the PhotoJournalInclusivePages field is still only 30 characters.

dnoneill:
Is this still a problem in Illiad or can we close this PR?

PR #1155, opened 2021-02-18
Remove teaspoon from the project paving the way for Ruby 3.0

mejackreed:
I aimed to keep the essence of all of the existing tests. jasmine-jquery did some magic so removing has some slight consequences but never seemed too bad in my opinion.


spec/javascripts/additional_user_validation_fields_spec.js
```diff
@@ -36,30 +34,29 @@ describe('Additional User Validation Fields', function() {
...
-    beforeAll(function() {
-      this.fixtures = fixture.load('additional_user_validation_fields_elements.html');
+  describe('groupedUserFields()', () => {
+    beforeEach(() => {
+      document.body.innerHTML = fixture_elements;
     });
 
-    describe('buttonWrapper()', function() {
-      it('returns the div wrapping the button', function() {
-        expect(additionalUserValidationFields.buttonWrapper()).toExist();
```
mejackreed, line 46: One of the consequences of the removal of jasmine-jquery.. Doesn't feel too bad.

jkeck (approved):
Thanks for tackling this! Ruby 3, here we come!

PR #1173, opened 2021-05-06
Update bundle to address security vulnerabilities

jgreben:


cbeer:
I think the dependency updates took care of this?

PR #1191, opened 2021-07-21
Refactor LibraryLocation class to extract a new `RequestAbility` class

cbeer:
This PR extracts some non-location information out of the `LibraryLocation` class, where additional behaviors seem to have accumulated over the years. Before starting to apply re-opening rules, it seemed like a good opportunity to clarify some of these components.

There are 3 major changes:
- extracting some of the business logic into YAML configuration
- extracting the `LocationRules` and `RequestAbility` classes to evaluate what types of requests can be made for a given library/location/item_type/etc
- cleaning up some loose ends in order to re-focus `LibraryLocation` on location-specific data accessors.

For slightly easier reviewability, I've left the original tests for the different types of requests intact (with a little mechanical transformation to use consistent naming, etc). Possible follow-up would be to move some/all of those tests onto `LocationRules`/`RequestAbility`.


mjgiarlo (approved):
Nice change!

Issue #1205, opened 2021-08-30
Restore pickup library list on 9/7/21 for HoldRecall and Page


saseestone:
We'll need to restore the pickup library list to it's pre-pandemic glory, allowing users the ability to pickup materials from any open campus library.  (Unless we have rules that restrict the pickup - i.e. "PAGE-" locations.) This should apply to both PAGE and HoldREcall request types.

We should plan to update the list on 9/7/21 when campus libraries open.

If there are no restrictions, the following library list should display:
Art & Architecture Library (Bowes)
Business Library
Earth Science Library (Branner)
East Asia Library
Engineering Library
Green Library
Law Library (Crown)
Marine Biology Library (Miller)
Music Library
Science Library (Li and Ma)

cbeer:
Just confirming we're dropping Education (obviously) and RWC, and adding MEDIA-MTXT?

saseestone:
Ack. Good catch Chris.  MEDIA-MTXT shouldn't go in the general list.  We need it as an option for restricted paging though (e.g. PAGE-LP & PAGE-MD and the new PAGE-MM)

Confirming we are dropping Education and RWC for now.

PR #1207, opened 2021-09-01
[Ship 9/7] Refactor paging to use LocationRules definitions provided by Settings

cbeer:
This is an attempt to unify and consolidate the business logic for paging (and mediated paging) into an easier-to-grok DSL in Settings to express:

- what is eligible for paging (or needs mediated paging): `library`, `locations`, `locations_match` (already in use for scanning + hold/recall)
- where it can go: `pickup_libraries`, `additional_pickup_libraries`, `default_pickup_library`
- special properties or information needed for pulling: `item_commentable`, `confirm_eligibility`

As with scanning rules, the first applicable rule is the one that the application uses for the item.

This PR also consolidates other library-related metadata (labels, contact info, hours).

The rules incorporate the picklist changes needed for #1205 and other reopening tasks (e.g. https://github.com/sul-dlss/SearchWorks/issues/2728... not that requests ever enforced what sort of requests could be submitted).


app/models/library_location.rb
```diff
@@ -11,82 +11,21 @@ def initialize(library, location)
...
     end
 
     def all_libraries
-      config.libraries
+      Settings.libraries
     end
 
     def pageable_libraries
-      all_libraries.select { |k, _| config.pageable_libraries.include? k }
+      all_libraries.map.select { |_, v| v.pageable }.to_h.stringify_keys
```
mjgiarlo, line 28: Hmm, interesting. What is the `.map` adding here?

app/models/request.rb
```diff
@@ -169,15 +170,23 @@ def library_id_error?
...
-      Settings.origin_admin_groups.to_hash.keys.map(&:to_s).select do |library_or_location|
-        MediatedPage.exists?(origin: library_or_location) || MediatedPage.exists?(origin_location: library_or_location)
-      end
+      Settings.mediateable_origins.map.select do |code, config|
+        if config.library_override
+          MediatedPage.exists?(origin_location: code.to_s)
+        else
+          MediatedPage.exists?(origin: code.to_s)
+        end
+      end.to_h.stringify_keys
```
mjgiarlo, line 189: Can you say more about the need for string-key hashes? Path of least resistance?

app/models/requests/page.rb
```diff
@@ -15,6 +15,13 @@ def token_encryptor_attributes
     super << user.email
   end
 
+  # Ideally, we'll be able to drop the wildcard rule because no requests will make it here
```
mjgiarlo, line 18: 👍🏻

mjgiarlo (approved):
I have a couple of minor questions, answers to which should not block this PR. Overall, I'm a big fan of moving from Rails config to a settings-based approach, and though I admit this codebase is pretty foreign to me and this is not a simple PR... I feel like the changes made increase the app's readability and make it easier to reason about. Support moving forward with this if @camillevilla concurs!

app/models/library_location.rb
cbeer: I couldn't get the RailsConfig `select` to work 🤷‍♂️

app/models/request.rb
cbeer: Yup. We had string keys before, tests that said they should be strings, and stuff that broke if they weren't 🤷‍♂️

app/models/library_location.rb
cbeer: (And.. you might ask `why not #to_h?`... `map` preserves the railsconfig objects further down the tree, and `to_h` turns everything in the tree into ordinary hashes)

app/models/request.rb
mjgiarlo: Meh, easy enough to stick with string keys.

app/models/library_location.rb
mjgiarlo: @cbeer I ... don't think I understand why that'd be the case at all, but if it works, great. Maybe toss these in a comment above this line as an act of kindness to posterity?

cbeer: Sure.

```
Settings.library.select # => nil
Settings.library.to_h # => { default: { ... }, ... }
Settings.library.map.to_h # { default: #<Config::Options  ... > }
```

Why it works that way, 🤷

app/models/library_location.rb
```diff
@@ -11,82 +11,23 @@ def initialize(library, location)
...
-    def config
-      SULRequests::Application.config
+      all_libraries[code]&.label
     end
 
+    # This is a super-clunky way to convert data from RailsConfig to something
+    # Enumerable, so we can use e.g. #select
     def all_libraries
-      config.libraries
+      Settings.libraries.map.to_h.with_indifferent_access
```
mjgiarlo, line 26: Even better! 👏🏻

camillevilla (approved):
This looks good to me! The Honeybadger notification for the fallthrough case is nice.

mjgiarlo:
👏🏻

Issue #1206, opened 2021-08-31
Update interstitial pages for all request types and libraries


saseestone:
Details here: https://docs.google.com/document/d/114MCYVXb_Fw0LcbbzQOff9c90Sh7D0AW2mhJDOAIQvU/edit#heading=h.csvgigbqdy8p

PR #1209, opened 2021-09-02
Update Rumsey eligibility statement

cbeer:
Part of https://github.com/sul-dlss/sul-requests/issues/1206


mjgiarlo (approved):
LGTM. Also wondering if it might be worth stuffing the repeated language in i18n or other config at some point, but it's no big deal.

cbeer:
I think the eligibility is just "coincidentally" the same (thanks to Sarah). Before the pandemic, it wasn't there at all, and hopefully will go away once things are fully open (or it'll diverge again based on service 🤷‍♂️ )

saseestone:
Yes, "coincidentally" the same because we're emphasizing limited services because of pandemic. 

Predicting if we have any text here after we fully reopen is difficult - but mainly because of the Aeon implementation.

Issue #1291, opened 2022-05-03
Non-Stanford researchers receive a "User is Blocked" message


anarchivist:
We got the following message from Tim:

> In doing the approvals it seems that the outside researchers are getting a "User is Blocked" message. The material is delivered from SAL 3 but we do not get a printout. Can't figure out why the outside researchers are getting this message.

I'm guessing the printout is essential to the SPEC Public Services workflow.

anarchivist:
From @saseestone:

> Had a minute to check email and saw the report from Spec that non-Stanford users are getting a “user is blocked” response. I saw a SW feedback from EAL late last night saying the same. I did check that the HOLD@XX pseudos had a status of OK (meaning not blocked). But that's as far as I got looking into it

anarchivist:
More from Tim:

> The gist is that outside researchers are making their requests but we don't get acknowledgment that the request has gone through. In other words, we're not getting the "done" acknowledgment when a successful request goes through. This is our view (don't know why it came out blurry) but this researcher is only here for a few days and we want to make sure we get these boxes:

![image 1  copy](https://user-images.githubusercontent.com/73732/166510239-9914fba2-327a-4cc8-8a9d-2d1c713e5228.png)

PR #1292, opened 2022-05-03
Only enforce patron standing for hold recalls

cbeer:
I don't think this is necessarily right, but:

scan falls back to a SCAN@ pseudopatron
page + mediated page fall back to the pseudopatron for blocked + expired users

Related to #1291


corylown (approved):
🤷

PR #1295, opened 2022-05-05
Use abilities to determine patron access to types of requests

cbeer:
Builds on #1293


app/controllers/requests_controller.rb
```diff
@@ -43,10 +42,22 @@ def update
     end
   end
 
+  def success
```
corylown, line 45: @cbeer what is this empty method doing (and `def ineligible` on the next lines)?

corylown: nevermind...just answered my own question

corylown (approved):

PR #1339, opened 2022-09-13
Aeon requests

cbeer:


app/views/aeon_pages/_info_modal.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <% if current_request.finding_aid? %>
+      <%= t('.steps.finding_aid_html') %>
+      <% elsif current_request.holdings.many? %>
+      <%= t('.steps.multi_vol_html') %>
+      <% else %>
+      <%= t('.steps.single_vol_html') %>
+      <% end %>
+    </ol>
+    <p class="dialog-more-info">
+      <%= t('.more_details_html', library: Settings.libraries[i18n_library_title_key].label, reading_room_url: Settings.libraries[i18n_library_title_key].reading_room_url) %>
```
cbeer, line 18:
```suggestion
      <%= t('.more_details_html', library: Settings.libraries[current_request.origin].label, reading_room_url: Settings.libraries[current_request.origin].reading_room_url) %>
```

My read of https://github.com/sul-dlss/sul-requests/commit/f63f940d310c13db62372014d21e2be3770f91f4 suggests it was implemented specifically for I18n purposes, and might not be the library code?

I wonder if adding a `Request#library_metadata` access or something would be even better?

app/views/aeon_pages/_info_modal.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    </p>
+    <div class="dialog-buttons">
+      <% if current_request.finding_aid? %>
+        <%= link_to t('.continue_button'), current_request.finding_aid, class: 'btn btn-md btn-primary' %>
+      <% elsif current_request.all_holdings.many? %>
+        <%= button_tag(t('.continue_button'), class: 'btn btn-md btn-primary', type: 'button', data: { behavior: 'close-overlay', 'close-target' => '#aeon-info-overlay' }) %>
+      <% else %>
+        <%= submit_tag t('.continue_button'), class: 'btn btn-md btn-primary', form: 'new_request' %>
```
cbeer, line 26: What's does the `form: 'new_request'` option do?

config/settings.yml
```diff
@@ -333,12 +342,17 @@ libraries:
...
     pageable: true
+    aeon_sites:
+      L-PAGE-EA: EASTASIA
   SAL3:
     label: SAL3 (off-campus storage)
     hours:
       library_slug: sal3
       location_slug: operations
     pageable: true
+    aeon_sites:
```
cbeer, line 353: I wonder if the `aeon_site` configuration would be better on the `pageable` rule, so it's harder to the data out of sync?

spec/controllers/requests_controller_spec.rb
```diff
@@ -57,7 +57,7 @@
     describe 'unscannable item' do
       it 'redirects to the new mediated page request form' do
         get :new, params: mediated_page_params
-        expect(response).to redirect_to new_mediated_page_path(mediated_page_params)
+        expect(response).to redirect_to new_aeon_page_path(mediated_page_params)
```
cbeer, line 60: I'm surprised this test is changing; does the underlying fixture data need to be updated to avoid trigger the Aeon rules?

app/views/aeon_pages/_info_modal.html.erb
thatbudakguy: allows this button to submit the underlying (hidden) form directly, by targeting it with the [`form` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form), while not needing to be located within the markup of the form.

spec/controllers/requests_controller_spec.rb
thatbudakguy: i think this was an old change before i started trying to shift some of the fixture data (e.g. for Hoover archive); I can probably find a way to alter the fixture to prevent it from changing.

config/settings.yml
thatbudakguy: I like that; I think it'll result in less nesting in the YAML too. I'll try it

thatbudakguy:
many tests were written assuming `SPEC-COLL` uses mediated paging and fail if it's configured to use Aeon. rather than change them all to use `ART` (and then maybe change them again later), I added some mocking so that `SPEC-COLL` goes back to being mediated paging for the purposes of those tests. idk if this is the best approach, but it seemed the least interventionist way to prevent the tight coupling between config and tests.

cbeer:
>  I added some mocking so that SPEC-COLL doesn't use Aeon for the purposes of those tests.

I don't love integration tests that mock so much stuff (and, potentially, it's confusing if things are true in test that won't ever be true in production). I'd say updating SPEC -> ART is entirely appropriate for integration tests (maybe followed closely by a fictional test library, but that's probably more trouble than it's worth)

thatbudakguy:
updated SPEC -> ART all the places I thought it made sense.

PR #1341, opened 2022-09-15
Make some seams for changing out the submit request job

cbeer:


thatbudakguy (approved):
this looks good — presuming the intention is that we can later implement a `FolioBase` and switch most of these models to inherit from that instead, to cut down on places we need to change things? I like that strategy.

PR #1345, opened 2022-09-23
Add debug gem and generate rdbg binstub

thatbudakguy:
Also regenerates bundler, because otherwise it will complain that
bundler was not generated by bundler when trying to run rdbg.


Gemfile
```diff
@@ -84,6 +84,7 @@ group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   # or call 'binding.pry'  (you may need require 'pry-byebug' first)
   gem 'pry-byebug'
```
cbeer, line 86: Can we remove `pry-byebug` if we're using `debug` instead?

thatbudakguy: for sure!

PR #1357, opened 2022-10-05
Don't bother enqueuing CDL jobs; if it fails, we'll just run it again…

cbeer:
… in a minute anyway


thatbudakguy:
rebasing so https://github.com/sul-dlss/sul-requests/pull/1356 will fix CI

thatbudakguy (approved):

Issue #1367, opened 2022-10-28
For 12/13/22: Replace current BorrowDirect API with new ReShare API


saseestone:
IPLC, the governing body for BorrowDirect, is moving away from Relais to ReShare. We will need to replace our current API with a new ReShare API when routing RecallHold requests to BorrowDirect.

There is API documentation available in basecamp: https://3.basecamp.com/5319802/buckets/26606684/documents/5132521312
Since it requires a basecamp account, I've copied it here: https://docs.google.com/document/d/1u8NouCFH1DVBV74vNtxv30zp9D0HmQI6yGrCMV9W7Wo/edit#
Hilary Thorsen arranged for my basecamp account, and we can request others through her as needed.

IPLC plans to cut over from Relais to ReShare on ~~12/1/22 OR on~~ ~12/6/22~ 12/13/22.

corylown:
Dropping some notes on this:

- The VuFind Search API is unrestricted at the moment; no authorization is required. It enables us to lookup a record by an ISN (or other fields) and find its lending status and its shared inventory record ID, which is required for placing a request via the OpenURL Resolver. (example VuFind Search API ISN request: https://borrowdirect.reshare.indexdata.com/api/v1/search?lookfor=0791056554&type=ISN&field%5B%5D=id&field%5B%5D=lendingStatus&sort=relevance&page=1&limit=20&prettyPrint=false&lng=en
- The Open URL Resolver will enable us to place a request for an item using the shared inventory ID. **It's IP restricted, so we'll need to share (with someone?) our server IP addresses for access.**
- mod-rs API will let us lookup request status. We'll need a **username**, **password**, and **tenant** **(do we need to request these?)**. I think this would be used by mylibrary to display request status to patrons. I've created a ticket over in mylibrary to represent this work: https://github.com/sul-dlss/mylibrary/issues/757

saseestone:
Just heard from Hilary that the earliest date for ReShare will be 12/6, and it might be delayed again.  IF there is a delay, James R. will announce by 11/17.

PR #1377, opened 2022-11-22
Implement BorrowDirect Reshare requests for hold recalls

cbeer:
BorrowDirect is changing back-end providers. The new provider (reshare) requires our application do 2 things:

- look up an item in their VuFind to figure out what id to request
- send a request to the "IPLC listener" to request the item for our patron

Fixes https://github.com/sul-dlss/sul-requests/issues/1367

### DIFF
--- a/app/jobs/submit_iplc_listener_job.rb
+++ b/app/jobs/submit_iplc_listener_job.rb
@@ -34,6 +34,7 @@ class SubmitIplcListenerJob < ApplicationJob
 
     request.borrow_direct_response_data = iplc_response.as_json
     request.via_borrow_direct = true
+    # We don't care if this save was successful or not, because if it went through the back-end service, then requests is no longer the source of truth for this.
     request.save
 
     request.send_approval_status!
--- a/app/jobs/submit_iplc_listener_job.rb
+++ b/app/jobs/submit_iplc_listener_job.rb
@@ -34,7 +34,8 @@ class SubmitIplcListenerJob < ApplicationJob
 
     request.borrow_direct_response_data = iplc_response.as_json
     request.via_borrow_direct = true
-    # We don't care if this save was successful or not, because if it went through the back-end service, then requests is no longer the source of truth for this.
+    # We don't care if this save was successful or not, because if it went through the back-end service,
+    # then requests is no longer the source of truth for this.
     request.save
 
     request.send_approval_status!
--- a/app/jobs/submit_iplc_listener_job.rb
+++ b/app/jobs/submit_iplc_listener_job.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+##
+# Background job to send requests to IPLC.
+# If Reshare has the item and we succesfully request it, let the user know it is on the way.
+# If Reshare does not have it or we cannot succesfully request it, trigger the normal Symphony request workflow.
+class SubmitIplcListenerJob < ApplicationJob
+  discard_on ActiveRecord::RecordNotFound do |job, _error|
+    Honeybadger.notify(
+      "Attempted to call Reshare for Request with ID #{job.request_id}, but no such Request was found."
+    )
+  end
+
+  def perform(request_id, instance_uuid)
+    request = Request.find(request_id)
+
+    Sidekiq.logger.info("Started SubmitIplcRequestJob for request #{request_id}")
+
+    begin
+      make_iplc_request(request, instance_uuid)
+    rescue StandardError => e
+      Honeybadger.notify("IPLC Request failed for #{request_id} with #{e}. Submitted to Symphony instead.")
+
+      request.send_to_symphony_now!
+    end
+
+    Sidekiq.logger.info("Completed SubmitIplcRequestJob for request #{request_id}")
+  end
+
+  def make_iplc_request(request, instance_uuid)
+    iplc_response = IplcWrapper.new(request, instance_uuid)
+
+    raise 'IPLC HTTP request failed' unless iplc_response.success?
+
+    request.borrow_direct_response_data = iplc_response.as_json
+    request.via_borrow_direct = true
+    request.save
+
+    request.send_approval_status!
+  end
+
+  # Basic client for working with the IPLC listener HTTP API.
+  class IplcWrapper
+    attr_reader :request, :instance_uuid
+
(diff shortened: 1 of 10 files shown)


corylown (approved):

app/jobs/submit_iplc_listener_job.rb
```diff
@@ -0,0 +1,91 @@
...
+  discard_on ActiveRecord::RecordNotFound do |job, _error|
+    Honeybadger.notify(
+      "Attempted to call Reshare for Request with ID #{job.request_id}, but no such Request was found."
+    )
+  end
+
+  def perform(request_id, instance_uuid)
+    request = Request.find(request_id)
+
+    Sidekiq.logger.info("Started SubmitIplcRequestJob for request #{request_id}")
```
jcoyne, line 17: Why do we need `Sidekiq` here? `logger` is an instance method on `ApplicationJob`, right?

jcoyne (changes requested):

app/jobs/submit_iplc_listener_job.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    begin
+      make_iplc_request(request, instance_uuid)
+    rescue StandardError => e
```
jcoyne, line 21: I feel like this might be too generic.  Would it make sense for line 33 to raise a specific error and rescue it here?

app/jobs/submit_iplc_listener_job.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def make_iplc_request(request, instance_uuid)
+    iplc_response = IplcWrapper.new(request, instance_uuid)
+
+    raise 'IPLC HTTP request failed' unless iplc_response.success?
+
+    request.borrow_direct_response_data = iplc_response.as_json
+    request.via_borrow_direct = true
+    request.save
```
jcoyne, line 37: Should this be `request.save!` or should we look for a truthy value?

app/jobs/submit_iplc_listener_job.rb
```diff
@@ -0,0 +1,91 @@
...
+        req_id: iplc_patron_id,
+        rft_id: instance_uuid,
+        'res.org': 'ISIL:US-CST',
+        'svc.pickupLocation': iplc_pickup_location_code,
+        rfr_id: request.to_global_id.to_s
+      }
+    end
+
+    def iplc_patron_id
+      request.user.patron.university_id
```
jcoyne, line 84: Does a patron always have a university_id?  I noticed that @corylown is testing for this in https://github.com/sul-dlss/mylibrary/pull/761/files/42f79639511ea51e2728fc2239d181bd2f7fd3c9..d21ea15dc62135fa44d1fb7c64a409f3f9f51d37#diff-12bc95bc120bd5dd5050c084ac44aaefea6a87f3bf01b373720a3e79e2ad1abeR21 and I wonder if it's the same value?

app/jobs/submit_iplc_listener_job.rb
cbeer: I'm guessing it's to support the special log level configuration for Sidekiq?

https://github.com/sul-dlss/sul-requests/blob/1f6e0dd89895cf6cf287d6e0aeb35a151162311f/config/initializers/sidekiq.rb#L3

This is just consistent with the other request jobs :shrug:

app/jobs/submit_iplc_listener_job.rb
cbeer: 🤷‍♂️ . The other request jobs don't seem to care (and... I think that's fine? If the request went through to the back-end service, requests is no longer the source of truth for that request)

jcoyne: Can you add a comment about that here. I think that's important.

jcoyne, line 37:
```suggestion
    # We don't care if this save was successful or not, because if it went through the back-end service, then requests is no longer the source of truth for this.
    request.save
```

app/jobs/submit_iplc_listener_job.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def make_iplc_request(request, instance_uuid)
+    iplc_response = IplcWrapper.new(request, instance_uuid)
+
+    raise 'IPLC HTTP request failed' unless iplc_response.success?
+
+    request.borrow_direct_response_data = iplc_response.as_json
+    request.via_borrow_direct = true
+    # We don't care if this save was successful or not, because if it went through the back-end service, then requests is no longer the source of truth for this.
```
corylown, line 37: Make rubocop happy
```suggestion
    # We don't care if this save was successful or not, because if it went through the back-end service,
    # then requests is no longer the source of truth for this.
```

app/jobs/submit_iplc_listener_job.rb
corylown: @cbeer may be able to speak to this more confidently, but my understanding is that we rescue from `StandardError` because if anything goes wrong we want to fall back to sending the request to Symphony.

FWIW the existing `SubmitBorrowDirectRequestJob#perform` method also wraps the request call in a rescue block that rescues `StandardError` -- so it's not a change from the current implementation: https://github.com/sul-dlss/sul-requests/blob/master/app/jobs/submit_borrow_direct_request_job.rb#L16

jcoyne: It seems to me this isn't exactly clear.  If that's indeed @cbeer's intent, it would be helpful to have some documentation on make_iplc_request that says what errors it expects to throw and when.

cbeer: > if anything goes wrong we want to fall back to sending the request to Symphony.

That's my understanding too. We'd prefer to get the book from some other library, but not getting the material in some way isn't an option. 

My suggestion would be... we merge this as-is, see how reshare actually breaks, and revisit it when we have time. If you have a comment in mind ("yes, we actually mean it when we're rescuing everything"?), feel free to add it.

corylown: Given that we're up against a deadline, I'm going to merge this as-is and we can revisit if needed.

PR #1387, opened 2022-12-13
Ensure IPLC pickup location code is uppercase

corylown:
We're seeing an issue where pickup location codes are not populated in reshare and suspect it is because they're being sent as `STA_green` instead of `STA_GREEN`


jcoyne (approved):
Feel free to merge, but this seems like the type of change that could benefit from a test.

corylown:
@jcoyne I had the same thought -- I'll merge this and see about some docs/tests.

PR #1389, opened 2022-12-13
Add comments and tests for SubmitIplcListenerJob::IplcWrapper

corylown:
There were previously copied/duplicate tests from `SubmitReshareRequestJob::ReshareVufindWrapper` in `spec/jobs/submit_iplc_listener_job_spec.rb`, which I've replaced with tests for the class that exists there. Also adds a comment about how to convert Symphony location codes to ReShare location codes.


jcoyne (changes requested):
Do these tests and documentation cover the need for the upcase?  I'm not sure I see it.

corylown:
@jcoyne `underscore` returns the lowercase form of the string (https://apidock.com/rails/String/underscore). Maybe it would be more direct to use `.tr('-', '_')`

jcoyne (approved):

PR #1426, opened 2023-03-31
Migrate analytics to GA4

marlo-longley:
Wait to deploy until https://github.com/sul-dlss/shared_configs/pull/1949 is deployed
Closes https://github.com/sul-dlss/sul-requests/issues/1425


app/assets/javascripts/analytics.js
```diff
@@ -1,55 +1,15 @@
...
 
 $(document).on("turbolinks:load", function(){
-  GoogleAnalytics.load();
-  if (GoogleAnalytics.analyticsId){
-    GoogleAnalytics.trackPageview();
+  // gtag config
+  const config = {}
+  // To turn off debug mode, exclude the parameter altogether (cannot just set to false)
+  // See https://support.google.com/analytics/answer/7201382?hl=en#zippy=%2Cgoogle-tag-websites
+  if (document.head.querySelector("meta[name=analytics_debug]").getAttribute('value') === "true") {
```
corylown, line 11: It looked like for the other migrations, at least the ones I saw, we just put this script in the application template. Curious why in this case you've opted to retain the compiled analytics.js? By moving the script to the template you'd also avoid having to write/read the analytics debug setting from the DOM.

app/assets/javascripts/analytics.js
```diff
@@ -1,55 +1,15 @@
...
-
-  return GoogleAnalytics;
-
-})();
+// gtag intial setup
+window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
 
 $(document).on("turbolinks:load", function(){
```
corylown, line 6: Is this still needed? I thought we observed that GA4 tracked turbolinks page views correctly without any additional handling.

corylown (changes requested):

app/assets/javascripts/analytics.js
marlo-longley: I didn't know we'd concluded that. Thanks. Sure, I will move the script into the template.

app/views/layouts/application.html.erb
```diff
@@ -15,6 +15,18 @@
     <%= stylesheet_link_tag "application", media: "all" %>
     <link href="https://www-media.stanford.edu/su-identity/css/su-identity.css" rel="stylesheet">
     <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700" %>
+    <meta name="analytics_debug" value="<%= Settings.analytics_debug %>">
```
corylown, line 18: I think you don't need this line anymore, right?
```suggestion
```

corylown (approved):

Issue #1442, opened 2023-05-11
Use FOLIO APIs to drive the display of bibliographic information


cbeer:
Requests currently queries Searchworks to retrieve bibliographic information (and very intentionally doesn't go to Symphony). It seems like FOLIO might be a more reliable endpoint to hit to retrieve this information.

[SearchworksItem](https://github.com/sul-dlss/sul-requests/blob/master/app/models/searchworks_item.rb) probably represents the most important information we're currently displaying.

This might also include retrieving the current location for an item (e.g. for https://github.com/sul-dlss/sul-requests/blob/master/app/helpers/mediation_helper.rb#L7)

This may be a little tricky because searchworks is handing off the catkey/hrid, library + location codes (as it understands them) which don't match FOLIO's data. We may need to pull over the same legacy mappings from the indexer for now.

Issue #1331, opened 2022-08-22
Place item + instance ("title") level holds using the mod-patron (edge-patron) API


cbeer:

thatbudakguy:
possible that our version of FOLIO doesn't even support title level holds yet?

thatbudakguy:
(sorta) blocked on #1434

thatbudakguy:
we think:
- title-level holds "officially" supported in nolana
- title-level holds for on-order instances still unclear.

thatbudakguy:
to complete this ticket, compare the mod-patron and edge-patron APIs in FOLIO to decide which one we want to use, based on the data that is available to requests at the time (who the user is, what catkey (title) or barcode (item) was requested, where they want it to go). ideally we can use mod-patron since it comes "for free"; we could install and configure edge-patron if needed. does the API actually work and match its docs?

jcoyne:
The primary difference between the mod-patron and edge-patron APIs is the value of the account id.  mod-patron requires "The UUID of a FOLIO user" and edge-patron uses "Patron's external system Id stored in FOLIO user record".  How would we link a FOLIO user id to a sunetid?

thatbudakguy:
blocked until we figure out why hold requests are timing out; there doesn't seem to be enough error output to get to the bottom of this. more investigation needed (from libsys?)

thatbudakguy:
see slack thread: https://stanfordlib.slack.com/archives/G018TJ20XGE/p1683211174057659

thatbudakguy:
closing since our retry strategy has unblocked this.

cbeer:
I'm reopening this because I think we only discovered which API endpoints to call and verified we could call them with some data.

I think we need to finish investigating the API parameters to send in order to create the same types of user experiences. In requests today, our users can send:
- a free-form comment (currently truncated to 50 characters to meet Symphony requirements; is there any restriction in FOLIO-land?)
- specifying a "fill by date"/expiration for their recall or mediated page

And the system decides whether or not a hold request triggers a recall or not (certain types of patrons can't recall material). Is this something we can control, or does FOLIO take care of it automatically?

An additional API investigation is... the current contract with Symphony involves using catkeys and barcodes, but it looks like FOLIO wants UUIDs. Do we know the right API endpoints to hit for converting from catkeys + barcodes to the appropriate UUIDs?

cbeer:
[...]

PR #1449, opened 2023-05-17
Start building out Folio-specific models

### LINKED ISSUE #1434: Finish setting up requests-folio-dev
we have a server created by ops, but that's about it. needs some work to ensure it's configured and running correctly.

cbeer:
Builds on #1448 and creates a clear(er) seam for logic that depends on Symphony:

- patron data lookups
- current location lookups for mediated pages

This also shoves the Symphony-coupled API models into app/models/symphony and namespaces them appropriately. Hopefully soon those model names only show up in CDL and tests.

Fixes #1434 

If this is a plausible base implementation, these can be follow-on tickets:
- [ ] #1442 (?): Create a seam for current location lookups (e.g. a FOLIO seam for https://github.com/sul-dlss/sul-requests/blob/master/app/helpers/mediation_helper.rb#L7, used only for mediated pages)... and perhaps validate we even need to continue doing that? 
- [ ] #1450: place title level holds for records without any items (e.g. on-order)
- [ ] figure out what to do about recalls for multi-volume works (currently using title level holds in Symphony; FOLIO seems like it'll just fulfill any volume)
- [ ] pass on comments and expiration dates (investigation needed in #1331)
- [ ] store useful information from FOLIO responses for debugging
- [ ] figure out whether FOLIO circ rules or policies can express restricted paging to specific locations  
- [ ] #1442 
- [ ] ???
- [ ] Backfill some tests for the FOLIO client
- [ ] Figure out whether/how to use the [service-points endpoint](https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/p/service-point.html) to populate pickup locations
- [ ] Figure out if FOLIO takes care of page vs hold vs recall, or if we need to add that logic here


jcoyne (approved):

app/services/folio_client.rb
```diff
@@ -109,7 +109,6 @@ def get_json(path, **kwargs)
   # @raises [StandardError] if the response was not a 200
   # @return [Hash] the parsed JSON data structure
   def parse_json(response)
-    raise response unless response.status == 200
```
jcoyne, line 91: This intentionally guarded against a 500 error status, which ended up showing up as a JSON parse error.  I think we should add this back and maybe look for `response.success?` instead.

cbeer: Maybe... although when I tried that, ruby complained about trying to raise a Faraday response object instead of an exception 🤷‍♂️ It seemed like the `check_response` call is already checking the status, so maybe we don't need it?

jcoyne: Okay, I may have written this guard prior to adding check_response, so it may have been redundant.

PR #1459, opened 2023-05-18
Place item level holds with all the necessary metadata in FOLIO

cbeer:
Part of #1331


app/services/folio_client.rb
```diff
@@ -66,13 +66,13 @@ def create_instance_hold(user_id, instance_id, pickup_location_id)
   # @param [String] user_id the UUID of the FOLIO user
   # @param [String] item_id the UUID of the FOLIO item
   # @param [String] pickup_location_id the UUID of the pickup locatio
-  def create_item_hold(user_id, item_id, pickup_location_id)
+  def create_item_hold(user_id, item_id, **attr)
```
jcoyne, line 69: I rather prefer we're explicit about the allowed params here. Is there a reason we don't add patronComments and expirationDate params? That way the reader of this class knows what is expected without having to look at all of the consumers.

jcoyne (changes requested):

app/services/folio_client.rb
cbeer: I'm worried about the argument list getting pretty crazy for stuff we're just passing through (that's why we took a similar approach over in Symphony client...), but sure, whatever.

jcoyne (approved):

PR #1476, opened 2023-05-23
Get bibliographic data from Folio

jcoyne:


app/models/folio/item_with_status.rb
```diff
@@ -0,0 +1,52 @@
...
+      status == STATUS_AVAILABLE ? 'available' : 'unavailable'
+    end
+
+    # TODO, we probably need to handle "Page", which is something Symphony had.
+    def status_text
+      status
+    end
+
+    # TODO, HUH?
+    def current_location_code
```
jcoyne, line 36: To be removed once https://github.com/sul-dlss/sul-requests/pull/1501 is merged and this PR is rebased.

app/models/folio/item_with_status.rb
```diff
@@ -0,0 +1,52 @@
...
+      'derp'
+    end
+
+    def due_date
```
jcoyne, line 40: This is the last big todo.  Maybe we want it in this PR or it could be a follow on.

jcoyne: I've added a new ticket https://github.com/sul-dlss/sul-requests/issues/1510 to deal with this.

thatbudakguy (approved):

app/models/folio/bib_data.rb
```diff
@@ -0,0 +1,71 @@
...
+module Folio
+  # Bibliographic data from Folio.
+  class BibData
+    def initialize(request, _ = nil)
+      @request = request
+    end
+
+    attr_reader :request
+
+    def title
```
cbeer, line 12: We should probably verify that edits to the MARC bib record sync to these FOLIO fields

app/models/folio/bib_data.rb
```diff
@@ -0,0 +1,71 @@
...
+      json.fetch('indexTitle', '')
```
cbeer, line 13: Is there a reason to pick `indexTitle` over `title`? The docs suggest indexTitle is intended for browse, not general display:

```

    "indexTitle": {
      "type": "string",
      "description": "Title normalized for browsing and searching; based on the title with articles removed"
    },
```

app/models/folio/bib_data.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def author
```
cbeer, line 16: Do we need to filter this list at all? I'm trying to find the symphony equivalent, but it'd surprise me if we show every contributor (but maybe FOLIO means something different?)

app/models/folio/bib_data.rb
```diff
@@ -0,0 +1,71 @@
...
+    def finding_aid?
+      finding_aid.present?
+    end
+
+    def view_url
+      [Settings.searchworks_api, 'view', request.item_id].join('/')
+    end
+
+    def instance_id
+      @instance_id ||= folio_client.resolve_to_instance_id(hrid: "a#{request.item_id}")
```
cbeer, line 58: I think we only want to append the `a` with a numeric hrid for backwards compatibility (and nicer looking URLs). We'll have other hrids with other prefixes (e.g. `L0123`) that shouldn't have that prefix.

app/models/folio/holdings.rb
```diff
@@ -0,0 +1,58 @@
...
+    # @return [Array<OpenStruct>] a list of every holding in the requested library/location
+    def all
+      @all ||= items_in_location.map do |item|
+        item.with_status(@request.item_status(item.barcode))
+      end
+    end
+
+    def single_checked_out_item?
+      all.one? &&
+        all.first.current_location.try(:code) == 'CHECKEDOUT'
```
cbeer, line 33: Where is `current_location` defined? Should we use the `checked_out?` accessor you have below?

app/models/folio/item.rb
```diff
@@ -0,0 +1,29 @@
...
+                         type:,
+                         callnumber:,
+                         public_note:,
+                         permanent_location:,
+                         temporary_location:,
+                         request_status:)
+    end
+
+    def self.from_hash(dyn)
+      new(barcode: dyn.fetch('barcode'),
```
cbeer, line 20: I feel like we've found examples of items that don't necessarily have all these fields. Is there a reason to be so strict?

app/models/folio/item_with_status.rb
cbeer: Do we really want to raise an exception and break the page? Some placeholder value seems like it'd let us test things  before #1510 is done.

cbeer (changes requested):

cbeer:
I'm getting this exception when I try to run the app:

> undefined method `current_location' for #<data Folio::ItemWithStatus barcode="36105235919805", status="Available", request_status=#<ItemStatus:0x000000010b6082e8 @request=#<Request id: nil, type: nil, needed_date: nil, status: nil, created_at: nil, updated_at: nil, origin: "MEDIA-MTXT", destination: nil, origin_location: "MM-STACKS", item_id: "14437506", user_id: nil, data: {"request_status_data"=>{}}, item_title: nil, barcodes: [], estimated_delivery: nil, approval_status: 0, via_borrow_direct: false>, @id="36105235919805">, type="multimedia", public_note=nil, permanent_location="MEDIA-CAGE", temporary_location=nil, callnumber="ZDVD 47270 BLU-RAY">

app/models/folio/item_with_status.rb
jcoyne: Made a fake due date now.

app/models/folio/bib_data.rb
jcoyne: I have no idea how any of that works. 🤷

app/models/folio/bib_data.rb
jcoyne: The title field also displays the authors. For example: "Coffee / Nebiyu Assefa and Joanna Brown" whereas the `indexTitle` is the actual item title (e.g. "Coffee").

app/models/folio/bib_data.rb
jcoyne: I don't know the answer to this. Seems like a question for @saseestone

app/models/folio/bib_data.rb
jcoyne: But the request coming from searchworks, doesn't use the `a` prefix.  So I'm not sure how we expect to find the record in Folio if we don't append it at some point.  I'm open to suggestions.

app/models/folio/holdings.rb
jcoyne: Yes, I think this was a bad rebase prior to https://github.com/sul-dlss/sul-requests/commit/1f25a5573f0ecf0bf654e554d1204dd2e45901e4#diff-b9a80913142bd85afc2319ad52bd074b495c84e7e2e9acf11dba1eac8305fd07L36

app/models/folio/bib_data.rb
cbeer: Like I said, I think we need to check if it's a numeric catkey. If it is, toss on the `a` prefix, but if it already has a letter, don't.

app/models/folio/item.rb
jcoyne: I don't like handling `nil` if there is no reason to. That's just added complexity for no reason.  Can you show examples where the API does not have these keys? I don't think I've seen any yet.

cbeer: https://github.com/sul-dlss/SearchWorks/issues/3198 fixed an issue with some similar looking code. Maybe the APIs do a better job keeping the data consistent, but I kinda doubt it.

jcoyne:
```
select count(*) from sul_mod_inventory_storage.item where lower(jsonb ->> 'barcode'::text) is null;
 count
-------
   173
```

Grrrr! Outliers.

cbeer: Welcome to libraries.

jcoyne: https://folio-test.stanford.edu/inventory/view/a77ee8aa-ecc2-410d-a37a-69419677aeac/fef99dba-15f4-495c-b391-e501afdaed6d/a9030d19-5bbe-4976-ba31-ef8840f4145d

app/models/folio/bib_data.rb
cbeer: But `indexTitle` has some other significant side-effects; for `a14439363`, the leading article gets dropped and proper nouns get down-cased:
> "title": "A history of Persia : from the beginning of the nineteenth century to the year 1858 / Robert Grant Watson ; with an introduction by Ali Ansari",

vs 

> "indexTitle": "History of persia : from the beginning of the nineteenth century to the year 1858",

app/models/folio/bib_data.rb
cbeer: I think this is a parity question. Searchworks is pulling the only first author from some very specific fields:

```
document['author_person_display']&.first ||
        document['author_corp_display']&.first ||
        document['author_meeting_display']&.first,
 ```

I see contributors have a `primary` attribute that might be useful for filtering without needing to replicate the MARC 100/110/111 behavior, but we'd need to do some due diligence to find out if it's similar.

File: app/models/folio/bib_data.rb (hunk 6 above)
Comment (jcoyne) [reply app/models/folio/bib_data.rb:58]: Okay, I don't think I understood what you were saying originally.

Comment (jcoyne) [reply app/models/folio/bib_data.rb]: Fixed now.

File: app/models/folio/bib_data.rb (hunk 3 above)
Comment (jcoyne) [reply app/models/folio/bib_data.rb:13]: Alissa states that "an edit to the 24x in SRS will trigger and automatic update to the instance."

File: app/models/folio/bib_data.rb (hunk 4 above)
Comment (jcoyne) [reply app/models/folio/bib_data.rb]: Switched this back to `title`. I'm splitting on `' / '` to remove the author part now.

File: app/models/folio/bib_data.rb (hunk 5 above)
Comment (jcoyne) [reply app/models/folio/bib_data.rb:17]: Filtered by "primary" now.

Comment (corylown) [reply app/models/folio/bib_data.rb:17]: Filtering by `primary` seems to get us the right contributors, at least the few I've spot-checked.

File: app/models/folio/bib_data.rb [hunk 9]
```diff
@@ -0,0 +1,79 @@
...
+
+    private
+
+    # Append "a" to the item_id unless it already starts with a letter (e.g. "in00000063826")
+    def hrid
+      request.item_id.start_with?(/\d/) ? "a#{request.item_id}" : request.item_id
+    end
+
+    def json
+      @json = folio_client.find_instance(instance_id:)
```
corylown, line 72:
```suggestion
      @json ||= folio_client.find_instance(instance_id:)
```

corylown (changes requested):

corylown (approved):

PR #1479, opened 2023-05-24
Remove unused method barcoded_holdings

jcoyne:


cbeer:
Instead of merging this, can you target this to the branch for #1458? This feature would support SAL-NEWARK paging, if it ever returns.

jcoyne:
@cbeer yes, but there's a SyntaxError on that branch. Do you want to fix that first?

jcoyne:
@cbeer this now targets that branch.

PR #1483, opened 2023-05-24
Add a unique index for the users.sunetid

jcoyne:


thatbudakguy (approved):

cbeer:
I'm a little worried it's not this straightforward, but maybe someone else can reason about it better.

- sunetid is nullable, and some databases don't like unique indexes on nullable columns.
- It seems possible a given sunetid might have different barcodes over time, and I think sunetids are eventually recycled
- requests isn't really in We don't really care about

PR #1490, opened 2023-05-25
Extract Searchworks::Holdings to its own class

jcoyne:
In Searchworks the bib data and holdings data come from the same request, but these are separate requests in Folio, so separating the two models now makes it easier to migrate


app/models/searchworks/holdings.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Searchworks
+  ###
+  #  ReqestedHoldings winnows down the entire
```
thatbudakguy, line 5: copypasta? this is no longer called `RequestedHoldings`

thatbudakguy (approved):

app/models/searchworks/holdings.rb
jcoyne: Fixed.

PR #1498, opened 2023-05-31
Remove a useless test

jcoyne:
This test was just testing the test code.

### DIFF
--- a/spec/models/item_status_spec.rb
+++ b/spec/models/item_status_spec.rb
@@ -2,13 +2,7 @@
 
 require 'rails_helper'
 
-def update_symphony_data_and_save(request, symphony_data)
-  r = Request.find(request.id)
-  r.symphony_response_data = symphony_data
-  r.save!
-end
-
-describe ItemStatus do
+RSpec.describe ItemStatus do
   subject { described_class.new(request, barcode) }
 
   let(:request) { create(:mediated_page_with_single_holding) }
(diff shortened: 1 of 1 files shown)


spec/models/item_status_spec.rb
```diff
@@ -50,21 +44,6 @@ def update_symphony_data_and_save(request, symphony_data)
       subject.approve!('jstanford')
     end
 
-    context 'persisting data' do
-      let(:request) { create(:mediated_page) }
-
-      it 'reloads the record to ensure that any serialized attributes are updated' do
```
jcoyne, line 56: The update was happening on line 8 ☝️.  So this test doesn't actually test anything about persisting.

cbeer: It looks like this came in with https://github.com/sul-dlss/sul-requests/commit/484ff426774572b2c52a7447c603c2b17370a34f; isn't this testing the side-effect of `#approve!`?

jcoyne: I think the authors misunderstood that

```
https://github.com/sul-dlss/sul-requests/pull/1498/files#diff-bf70ad4cafb7b3c0e98b0d99cb10fae48dfa55c85ddb3f248ba4faa4b1dc7044L60-L62
```
Is equivalent to writing:
```ruby
     r = Request.find(request.id)
     r.symphony_response_data = response
     r.save!
     expect(Request.ils_job_class).to receive(:perform_now).with(request.id, { barcode: })
```

jcoyne:
Superseded by #1500

PR #1501, opened 2023-06-02
Encapsulate hold item status

jcoyne:
This will allow the data structure to change when we use folio, but have the same interface

### DIFF
--- a/app/models/searchworks/holding_item.rb
+++ b/app/models/searchworks/holding_item.rb
@@ -3,6 +3,9 @@
 module Searchworks
   # A model on the Searchworks holding JSON data
   class HoldingItem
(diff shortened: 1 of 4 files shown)


app/models/searchworks/holding_item.rb
cbeer, line 1: Is this class tested anywhere? I'm guessing a in feature spec?

jcoyne: It doesn't have any unit tests, but it is tested via many feature specs such as `spec/features/item_selector_spec.rb`, `spec/features/create_scan_spec.rb`, `spec/features/create_aeon_page_request_spec.rb`, `spec/features/create_hold_recall_spec.rb`, etc.

PR #1533, opened 2023-06-27
Use pseudo patrons for Folio Page/MediatedPage requests

jcoyne:


config/settings.yml
```diff
@@ -348,12 +367,14 @@ libraries:
...
     reading_room_url: https://library.stanford.edu/spc/using-our-collections
     hold_pseudopatron: HOLD@SP
+    folio_hold_pseudopatron: 5eb559e2-e986-4158-bffd-16bb6e312ca1
     hours:
       library_slug: spc
       location_slug: field-reading-room
   TANNER:
     label: Philosophy Library (Tanner)
     hold_pseudopatron: HOLD@TA
+    folio_hold_pseudopatron: 5eb559e2-e986-4158-bffd-16bb6e312ca1
```
marlo-longley, line 377: Since these are all the same key/value (`folio_hold_pseudopatron: 5eb559e2-e986-4158-bffd-16bb6e312ca1`) is there a way to set it once, and reference a variable from then on? This could easier to read and easier to quickly understand that these all use the same pseudopatron, rather than trying to read these UUIDs multiple times.

jcoyne: It's not going to be all the same, Sarah just hadn't created them all just yet.

jcoyne: These UUIDs have now been updated.

thatbudakguy: Could we just store pseudopatron barcodes instead? That's what https://github.com/sul-dlss/folio-graphql/pull/88 assumes we will do when we migrate the config to FOLIO, and it's what we've done with other values that could otherwise be UUIDs (like location codes) so it's easier to edit them and see what they mean.

jcoyne: @thatbudakguy it's possible yes, but this would introduce another API call to resolve the user_id.  API calls are not reliable, so handling errors introduces unnecessary complexity.  Are we sure we want to add additional failure modes so that we can avoid looking at UUIDs?

jcoyne: In post-standup today we determined that this optimization could wait for later

marlo-longley:
We decided to merge this and revisit once work is merged that allows us to pull data about pseudopatrons directly from FOLIO (once we can use our graphql endpoint in sul-requests). At that point we will be referencing patron barcodes rather than the UUIDs we see here.

marlo-longley (approved):

PR #1562, opened 2023-06-30
Combine tests that use the same scenario setup

jcoyne:
This makes the suite faster since we don't have to repeat the identical test setup

### DIFF
--- a/spec/mailers/request_status_mailer_spec.rb
+++ b/spec/mailers/request_status_mailer_spec.rb
@@ -78,19 +78,22 @@ RSpec.describe RequestStatusMailer do
     end
 
     describe '#request_status_for_scan' do
+      subject(:body) { mail.body.to_s }
+
       let(:mailer_method) { :request_status_for_scan }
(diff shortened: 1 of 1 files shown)


thatbudakguy (commented):
does this really make a difference? i kind of prefer the opposite; single assertions per spec makes it easy to tie the failure to what was supposed to happen. tests are a place where being redundant and even slow is mostly fine for me, within reason.

jcoyne:
Yes, I think that fast tests help keep developers productive and prevent you from loosing focus on the problem at hand. For me, having a fast test suite is critical. Rspec points you directly at the line that failed, so I'm unclear as to how it would be confusing as to which assertion is problematic.

cbeer:
Would chained expectations (`expect().to include().and include()`) or `aggregate_failures` give us the best of both worlds? Consolidating the test setup seems good, but it's also nice if the tests give us the full picture of what went wrong 🤷‍♂️

thatbudakguy:
I like the chained expectations idea. That might also make the `Rspec/MultipleExpectations` cop happy.

thatbudakguy (approved):

PR #1565, opened 2023-07-03
Add ability for folio job to log its params and display them for debugging

### LINKED ISSUE #1560: [sul-requests/dev] ArgumentError: missing keyword: :logger
## Backtrace

    line 42 of [PROJECT_ROOT]/app/jobs/submit_folio_request_job.rb: initialize
    line 21 of [PROJECT_ROOT]/app/models/concerns/ils_request.rb: new
    line 21 of [PROJECT_ROOT]/app/models/concerns/ils_request.rb: ils_request_command

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/97292103)

jcoyne:
The SubmitFolioRequestJob::Command does not have a method called request_params. We can ignore this as it's on the debug page.

Fixes #1560

### DIFF
--- a/app/jobs/submit_folio_request_job.rb
+++ b/app/jobs/submit_folio_request_job.rb
@@ -49,10 +49,16 @@ class SubmitFolioRequestJob < ApplicationJob
     def execute!
       requested_items = barcodes.map do |barcode|
         item_id = folio_client.get_item(barcode)['id']
+        create_log(barcode:, item_id:)
         response = place_hold(item_id:)
         { barcode:, msgcode: '209', response: }
       end
       { requested_items: }
+    end
+
+    # Called by the debug views
+    def request_params
+      request.folio_command_logs.map(&:as_json)
     end
 
     private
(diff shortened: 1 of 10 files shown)


cbeer:
We'll probably want to implement something similar for FOLIO. This is useful debug information for us, and it seems a shame to just throw it away.

jcoyne:
@cbeer in Symphony, there is a single API call, but in FOLIO there are three separate calls (get_item, get_service_point, create_item_hold) so it doesn't make sense to cram all this together in the same way that we were doing for Symphony. I think it would be okay to re-add this if we have some design for what we want, but It seems like a higher priority to prevent this error right now until we can discuss what sort of design we need for this in the future.

cbeer:
At least recording the parameters we're sending to `create_item_hold` seems like a good + sufficient start to me. We'll get some information, and from that can uncover what else we might want from FOLIO.

jcoyne:
@cbeer that would mean this display function would require kicking off 2 API requests.  Is that something we want to do in this display page?

cbeer:
get_item and get_service_point are the API requests? That's not ideal, but I think we'll really benefit from having this information. If it's a problem, we already have a pattern for stashing data on the `Request` object, and we could stash that stuff if we need to?

jcoyne:
@cbeer What if this method is called before the job is run in the queue?  Do we then run these API calls or are we only going to show the values after it's been run in the queue?  I want to ensure we avoid a race condition with regard to writing in the JSON field as it writes all the json, which means two threads operating on the same record could cause data loss.

cbeer:
As a first pass, I think either behavior would be fine.

I'd point to https://stanfordlib.slack.com/archives/C0A7J65ME/p1688170260067899 as the most common scenario where this debug information is useful ("Someone tried to place a hold, and we don't know what's going on or what state it's in")

cbeer:
Sure, we can see how this goes. It strikes me as a little overkill compared to stashing stuff in `Request#data`, but maybe we'll get some benefit from this approach.

PR #1567, opened 2023-07-03
Delegate to requests to HoldRecall for FOLIO in-process items

### LINKED ISSUE #1557: Inprocess item should not show request and scan option
URL: https://searchworks-folio-dev.stanford.edu/view/14600413

What should show for an inprocess item?

thatbudakguy:
If there is a single item and it's being processed, don't offer
the user the option to page or scan and instead treat it as a
hold recall.

### DIFF
--- a/app/models/folio/holdings.rb
+++ b/app/models/folio/holdings.rb
@@ -32,6 +32,10 @@ module Folio
       all.one? && all.first.checked_out?
     end
 
+    def single_in_process_item?
+      all.one? && all.first.processing?
+    end
+
     private
 
     def items_and_holdings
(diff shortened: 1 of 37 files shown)


thatbudakguy:
Not sure if, or where, this should be tested...couldn't find a spec that seemed like it contained tests similar to this.

cbeer:
Our implementation should also cover the other hold-recall use cases (e.g. missing and on-order). These were covered in the `hold_recallable_location` rule for Symphony.

I'd maybe look at extending RequestAbilities to know something about statuses?

thatbudakguy:
> I'd maybe look at extending RequestAbilities to know something about statuses?

Would you be OK with me doing that in a separate PR from this? I have work from #1523 that creates a seam for a FOLIO-specific LocationRules implementation, so I could start with that.

cbeer:
In addition to or instead of this PR? I hope we can avoid a `single_in_process_item?` method before things get even more out of control 😬

thatbudakguy:
looks like #1556 will improve the situation in tests too, so I'm going to close this and work on a different path.

PR #1570, opened 2023-07-05
Remove feature flag for estimate_delivery

jcoyne:
This makes it easier to follow the logic and we don't have to maintain two branches when we only use one


cbeer:
I'm not sure we want to remove this functionality just yet. 

Although it was initially added as part of the pandemic reopening where we didn't want to make any promises about delivery dates, in the near-term it seems plausible we'd consider using it during the  Symphony -> FOLIO migration downtime too.

PR #1572, opened 2023-07-05
Move resolve_to_instance stub into the stub_folio_holdings method

jcoyne:


cbeer:
#resolve_to_instance was removed in #1589

PR #1574, opened 2023-07-05
Remove spec that tests nothing

jcoyne:


cbeer:
I wonder if this was a rebase gone wrong in #1476? It seems like we could just as easily add an appropriate expectation, right?

jcoyne:
@cbeer yes, I think a rebase messed up, but we already removed it once and then the rebase readded it: https://github.com/sul-dlss/sul-requests/pull/1546/files

PR #1581, opened 2023-07-06
Delegate hold-recallability to ILS-specific implementations

### LINKED ISSUE #1557: Inprocess item should not show request and scan option

thatbudakguy:
Currently, a request becomes a hold/recall when any of three conditions are met:
1. a barcode is provided (meaning a single item is requested)
2. the holding is in a special hold/recallable origin location (in symphony this covers things like `CHECKEDOUT`)
3. there is only one item in the holding, and that item is checked out

This PR expands the third condition to cover all the possibilities we aren't handling — namely, if _every_ item in the holding is allowed to be hold/recalled (checked out, in process, etc). It does this by delegating `hold_recallable?` down to the item level for each ILS's implementation. In Symphony, `CHECKEDOUT`, `INPROCESS`, etc. are item locations. In FOLIO, they are item statuses.

Fixes #1557

### DIFF
--- a/app/components/request_and_pickup_button_component.rb
+++ b/app/components/request_and_pickup_button_component.rb
@@ -23,7 +23,7 @@ class RequestAndPickupButtonComponent < ViewComponent::Base
   end
 
   def single_checked_out_item?
-    @current_request.holdings_object.all.one? && @current_request.holdings_object.all.first.checked_out?
+    @current_request.holdings_object.one? && @current_request.holdings_object.all?(&:checked_out?)
   end
 
   def single_library_value
--- a/app/components/request_and_pickup_button_component.rb
+++ b/app/components/request_and_pickup_button_component.rb
@@ -23,7 +23,7 @@ class RequestAndPickupButtonComponent < ViewComponent::Base
   end
 
   def single_checked_out_item?
-    @current_request.holdings_object.single_checked_out_item?
+    @current_request.holdings_object.all.one? && @current_request.holdings_object.all.first.checked_out?
   end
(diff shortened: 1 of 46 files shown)


app/components/request_and_pickup_button_component.rb
```diff
@@ -23,7 +23,7 @@ def default_pickup_library
   end
 
   def single_checked_out_item?
-    @current_request.holdings_object.single_checked_out_item?
+    @current_request.holdings_object.all.one? && @current_request.holdings_object.all.first.checked_out?
```
cbeer, line 26: Should we make Holdings an Enumerable so we can use `one?` or `all?` directly on that object? `all.first` and `all.one?` read a little funky.

thatbudakguy: yeah, I had the same thought. will do

thatbudakguy: sigh. doing this means you have to mock `each` in every test, because these things are all just regular `double`s...

thatbudakguy: i guess i could change all the `all:` to `each:` ...

thatbudakguy: @cbeer is there a smart way to do this I'm missing?

thatbudakguy: added some TODOs about it.

PR #1583, opened 2023-07-07
Implement FOLIO Patron proxy methods

### LINKED ISSUE #1552: Handle proxy patrons in FOLIO
Currently `Folio::Patron#proxy?` is just a stub.

cbeer:
fixes #1552

It's not ideal that we're adding a patron comment instead of using FOLIO's `proxyUserId`, but the mod-patron hold API doesn't (currently?) support proxies. Alternatively, we could use `circulation/requests`, but we'd have to implement FOLIO's hold type logic (recall, hold, or page) ourselves.


jcoyne:
@cbeer rubocop doesn't approve! 👮‍♂️

PR #1593, opened 2023-07-10
Make RequestAbilities#location_rule private

cbeer:
This creates a more convenient seam for a FOLIO implementation, which may not have the same kinds of location rules


jcoyne:
@cbeer please run `rubocop -a`.

jcoyne (approved):

Issue #1588, opened 2023-07-08
Implement Folio::ItemWithStatus methods


cbeer:
- [ ] status_class #1628
- [ ] status_text
- [x] due_date
https://github.com/sul-dlss/sul-requests/blob/c863fd05000feb4e7c78f9bc64cd1d8a325503d5/app/models/folio/item_with_status.rb#L72-L74

PR #1595, opened 2023-07-10
Fetch due dates from graphql

cbeer:
Part of https://github.com/sul-dlss/sul-requests/issues/1588


jcoyne:
@cbeer seems like a `dueDate`/`due_date` mismatch.

marlo-longley (approved):

PR #1600, opened 2023-07-11
Allow blocked Folio patrons to make requests

### LINKED ISSUE #1527: Place requests for users with blocks on behalf of the user instead of a pseudopatron
moved/copied from #1451.

in Symphony, we try to make a request, and when it fails (because the user hasn't paid fines, etc.) we instead make the request on behalf of a pseudopatron and let the staff sort it out later (manually changing the requestor back to the actual user).

in FOLIO, we can eagerly request the patron's blocked status, and if they are blocked from paging but the item would otherwise be pageable, we (think we) can create the request on behalf of the actual user (not a pseudopatron), which will save the staff some time.

the rationale here is that we want to eagerly create the request regardless because it's nice to our patrons, and so in that scenario it's better to save time and make it on behalf of a "real" person if possible.

we need to investigate whether attempting to place a request when the patron has a block errors, and if so make sure that it goes through anyway (the staff will see the block and can sort it out on their own).

ideally, we want to _always_ make the request regardless of patron status or blocks. so far, it seems like "inactive" status stops the request, but "blocked" does not.

### LINKED ISSUE #1558: User is blocked, but request placed with pseudo.
Request is being placed for HOLD@XX psuedo rather than the user account
https://searchworks-folio-dev.stanford.edu/view/14600413

jcoyne:
Fixes #1527
Fixes #1558

### DIFF
--- a/spec/models/folio/patron_spec.rb
+++ b/spec/models/folio/patron_spec.rb
@@ -14,26 +14,26 @@ RSpec.describe Folio::Patron do
       'username' => 'jcoyne85',
       'id' => '562a5cb0-e998-4ea2-80aa-34ac2b536238',
       'externalSystemId' => '09658748',
-      'barcode' => '2555716958',
+      'barcode' => '5555555555',
       'active' => true,
       'patronGroup' => '985acbb9-f7a7-4f44-9b34-458c02a78fbc',
       'departments' => [],
       'proxyFor' => [],
       'personal' =>
-        { 'lastName' => 'Coyne',
-          'firstName' => 'Justin',
-          'middleName' => 'Michael',
-          'email' => 'foliotesting@lists.stanford.edu',
-          'phone' => '(612) 868-2411',
-          'addresses' =>
-          [{ 'countryId' => 'US',
-             'addressLine1' => '4621 Ewing Ave S',
-             'city' => 'Minneapolis',
-             'region' => 'Minnesota',
-             'postalCode' => '55410-1745',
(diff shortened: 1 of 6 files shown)


spec/models/folio/patron_spec.rb
```diff
@@ -50,30 +50,65 @@
...
+              'preferredContactTypeId' => '002' },
+          'enrollmentDate' => '2022-09-01T00:00:00.000+00:00',
+          'createdDate' => '2023-05-20T00:13:20.324+00:00',
+          'updatedDate' => '2023-05-20T00:13:20.324+00:00',
+          'metadata' =>
+            { 'createdDate' => '2023-02-10T02:14:56.209+00:00',
+              'createdByUserId' => '3e2ed889-52f2-45ce-8a30-8767266f07d2',
+              'updatedDate' => '2023-05-20T00:13:20.249+00:00',
+              'updatedByUserId' => '9c77fffb-9d40-422c-8742-4c0f79957888' },
+          'customFields' => { 'affiliation' => 'staff', 'proximitychipid' => '0577826' }
```
thatbudakguy, line 107: not sure what `proximitychipid` is exactly, but are you sure you want it public?

thatbudakguy (approved):

spec/models/folio/patron_spec.rb
thatbudakguy: same for address, barcode, etc.

jcoyne: Thanks, redacted.

thatbudakguy (approved):

PR #1604, opened 2023-07-11
Don't validate when saving request after transmitting

jcoyne:
By transmitting the request to the ILS, the request is no longer valid as the item is already requested.

### DIFF
--- a/app/jobs/submit_folio_request_job.rb
+++ b/app/jobs/submit_folio_request_job.rb
@@ -18,7 +18,7 @@ class SubmitFolioRequestJob < ApplicationJob
 
     logger.debug("FOLIO response: #{response}")
     request.merge_ils_response_data(FolioResponse.new(response.with_indifferent_access))
-    request.save
+    request.save(validate: false) # By placing this request in the ILS, the item is no longer in a requestable state, so avoid validation.
     request.send_approval_status!
     logger.info("Completed SubmitFolioRequestJob for request #{request_id}")
   end


app/jobs/submit_folio_request_job.rb
```diff
@@ -18,7 +18,7 @@ def perform(request_id, options = {})
 
     logger.debug("FOLIO response: #{response}")
     request.merge_ils_response_data(FolioResponse.new(response.with_indifferent_access))
-    request.save
+    request.save!
```
cbeer, line 21: I thought `save!` raises exceptions?

jcoyne: Sorry, please reload. I forgot to save the file before committing.

app/jobs/submit_folio_request_job.rb
```diff
@@ -18,7 +18,7 @@ def perform(request_id, options = {})
 
     logger.debug("FOLIO response: #{response}")
     request.merge_ils_response_data(FolioResponse.new(response.with_indifferent_access))
-    request.save
+    request.save(validate: false) # By placing this request in the ILS, the item is no longer in a requestable state, so avoid validation.
```
cbeer, line 21: I can't think of any reason submitting the job should invalidate the request... so I guess this is probably fine? If that ever changes, I think there's a way we could instead relax the type validators for already-persisted requests but that seems... complicated.

cbeer (approved):

PR #1611, opened 2023-07-14
Searchworks availability_class is changing

jcoyne:
This allows sul-requests to handle either 'page' or 'deliver-from-offsite' with the same style


cbeer:
If there's no urgency beyond cleaning up a misleading label, I'd ask we try to ship #1610 first to avoid some nasty rebasing.

PR #1638, opened 2023-07-26
Add folio service point display

hudajkhan:
What this pull request does:

For FOLIO:

- This code will display service point codes and discovery display labels in the drop down for pickup destinations.  Furthermore, the FOLIO request abilities model employs the following logic to determine which service points to display and which to set as pre-selected in the dropdown:
-- Location restrictions supersede any other rules.  If the holdings effective location details have specific service points listed under "pageServicePoints", those are the service points that will be displayed as options for pickup.
-- When there are no location restrictions, the options will be all service points which have "isDefaultPickup" set to true in their details.  In addition, if the origin library for the request has a service point which has "pickupLocation" set to true, then that service point is also added to the list.
-- For default selections, EAST ASIA is selected if the origin location matches specific patterns.  Otherwise, the code checks for the campus code listed under effective location for the holding.  If there is a service point which lists that campus code under "isDefaultForCampus", then that service point will be the default selected option in the list.
- For paging schedules, since the destination is now a service point code, but library hours and other related logic uses library codes, the paging schedule controller maps from the service point destination to the library code.  A future pull request could further refine this so that service points are used throughout the process instead of relying on library codes.  
- Additionally, when the service point is "SCAN", it is just forwarded as SCAN

For Symphony:
- The code maintains the same logic as before, with the main difference that where methods and views referred to pickup_libraries, the code now uses pickup_destinations, with one branch executing if the FOLIO ILS flag is set and another branch executing for Symphony.

For tests:

- Added some logic to determine if the Settings flag for FOLIO is set or not, since with FOLIO, we would use service point destination codes like "GREEN-LOAN" and with Symphony we would use library codes like "GREEN".  If there is a more elegant way to make this distinction, that would be a welcome suggestion. 

How to evaluate:
- When you bring up the requests page, the dropdown should show service point codes instead of library codes.  Selecting the destination should kick off "SubmitFolioRequestJob" which should generate an email with the information for the request.  Delivery estimates should be displayed for the corresponding library.


app/jobs/submit_folio_request_job.rb
```diff
@@ -99,10 +99,14 @@ def pickup_location_id
...
 
     def get_service_point_code(destination)
-      return destination || Settings.Folio.default_service_point if Settings.ils.bib_model == 'Folio::Instance'
+      # Check if comparable service point code exists, otherwise return default
+      if Settings.ils.bib_model == 'Folio::Instance'
+        code = valid_service_point_code(destination) ? destination : Settings.folio.default_service_point
+      else
+        code = Settings.libraries[destination].folio_pickup_service_point_code
+        code ||= Settings.libraries['GREEN'].folio_pickup_service_point_code
+      end
```
jcoyne:
```suggestion
      code = if Settings.ils.bib_model == 'Folio::Instance'
        valid_service_point_code?(destination) ? destination : Settings.folio.default_service_point
      else
        Settings.libraries[destination].folio_pickup_service_point_code ||
          Settings.libraries['GREEN'].folio_pickup_service_point_code
      end
```

app/jobs/submit_folio_request_job.rb
```diff
@@ -192,6 +196,11 @@ def get_service_point_id(service_point_code)
     def get_library_for_service_point(service_point_id)
       Folio::Types.instance.get_type('locations').find { |location| location['primaryServicePoint'] == service_point_id }['libraryId']
     end
+
+    # Does this service point code exist
+    def valid_service_point_code(service_point_code)
```
jcoyne:
```suggestion
    def valid_service_point_code?(service_point_code)
```

jcoyne (changes requested):

app/jobs/submit_folio_request_job.rb
hudajkhan: For here, I realized since this is already the FOLIO submit job, I don't need any options for Symphony, so I removed the check altogether

.gitignore
```diff
@@ -33,3 +33,5 @@ yarn-debug.log*
 
 /app/assets/builds/*
 !/app/assets/builds/.keep
+
+**/.DS_Store
```
jcoyne, line 37: **suggestion:**, can you remove this from this PR and put it in your global `.gitignore` instead? Here are some directions https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

app/helpers/pickup_libraries_helper.rb
```diff
@@ -20,35 +22,70 @@ def label_for_pickup_libraries_dropdown(pickup_libraries)
 
   private
 
-  def select_for_multiple_libraries(form, pickup_libraries)
-    return unless pickup_libraries.length > 1
+  # Symphony
```
jcoyne, line 25: **suggestion**: this comment isn't clear. Can you expand it a bit?

app/helpers/pickup_libraries_helper.rb
```diff
@@ -20,35 +22,70 @@ def label_for_pickup_libraries_dropdown(pickup_libraries)
...
+          #{destination_label || pickup_destination}
         </div>
-        #{form.hidden_field :destination, value: library}
+        #{form.hidden_field :destination, value: pickup_destination}
       </div>
     HTML
   end
 
+  # Get the label, if it exists, for the pickup destination
+  def get_destination_label(pickup_destination)
```
jcoyne, line 57:
```suggestion
  def destination_label(pickup_destination)
```
I think this is a bit more conventional. See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AccessorMethodName

app/helpers/pickup_libraries_helper.rb
```diff
@@ -20,35 +22,70 @@ def label_for_pickup_libraries_dropdown(pickup_libraries)
...
+    destination_label = get_destination_label(pickup_destination)
     <<-HTML
       <div class='form-group'>
         <div class='#{label_column_class} control-label'>
-          #{label_for_pickup_libraries_dropdown([])}
+          #{label_for_pickup_destinations_dropdown([])}
         </div>
         <div class='#{content_column_class} input-like-text'>
-          #{Settings.libraries[library]&.label || library}
+          #{destination_label || pickup_destination}
```
jcoyne, line 49: **suggestion** Can we modify `get_destination_label` to return the argument you pass to it if it doesn't have a better label? That would eliminate this condition.

app/helpers/pickup_libraries_helper.rb
```diff
@@ -3,15 +3,17 @@
 ##
 # Helpers for formatting and displaying pickup libraries
 module PickupLibrariesHelper
-  def select_for_pickup_libraries(form)
-    pickup_libraries = form.object.pickup_libraries
-    return unless pickup_libraries.present?
+  include Folio::TypesUtils
+  # Pickup libraries for Symphony
+  def select_for_pickup_destinations(form)
```
jcoyne, line 8: **suggestion** Can this be made into a viewComponent? ViewComponents use proper encapsulation, so they are easier to test and reason about than helpers.

app/jobs/submit_folio_request_job.rb
```diff
@@ -93,13 +94,16 @@ def place_title_hold
...
-        code = Settings.libraries[request.destination].folio_pickup_service_point_code
-        code ||= Settings.libraries['GREEN'].folio_pickup_service_point_code
-
-        Folio::Types.instance.service_points.values.find { |v| v.code == code }&.id
+        code = get_service_point_code(request.destination)
+        get_service_point_id(code)
       end
     end
 
+    def get_service_point_code(destination)
```
jcoyne, line 102: **suggestion** please avoid prefixing methods with `get_` (see rubocop AccessorMethodName)

jcoyne (changes requested):

app/services/folio/types_utils.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Folio
+  # Map between service points and libraries and identify service points and libraries by code or id
+  module TypesUtils
```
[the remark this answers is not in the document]
hudajkhan: Integrating into Folio::Types as suggested

app/models/folio/pickup_destination.rb
```diff
@@ -0,0 +1,27 @@
...
+  class PickupDestination
+    # Get the default destination
+    def self.default_destination
+      Settings.folio.default_service_point
+    end
+
+    # @param code: service point or library code
+    def initialize(destination_code)
+      @code = destination_code
+      @type = 'ServicePoint'
```
jcoyne, line 14: I don't see `@type` used. Do we need it?

app/models/symphony/pickup_destination.rb
```diff
@@ -0,0 +1,27 @@
...
+  class PickupDestination
+    # Get the default destination
+    def self.default_destination
+      Settings.default_pickup_library
+    end
+
+    # @param code: service point or library code
+    def initialize(destination_code)
+      @code = destination_code
+      @type = 'Library'
```
jcoyne, line 14: I don't see this used. Do we need it?

app/models/symphony/pickup_destination.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module Symphony
+  # Encapsulates pickup destination logic for Symphony
+  class PickupDestination
+    # Get the default destination
+    def self.default_destination
+      Settings.default_pickup_library
+    end
+
+    # @param code: service point or library code
```
jcoyne, line 11:
```suggestion
    # @param [String] destination_code service point or library code
```

jcoyne: Isn't it just library code?

app/services/folio/types.rb
```diff
@@ -86,6 +88,68 @@ def get_type(type)
...
+
+    # Find the library id for the location with which this service point is associated
+    def library_for_service_point(service_point_id)
+      loc = get_type('locations').find { |location| location['primaryServicePoint'] == service_point_id }
+      loc.present? && loc.key?('libraryId') ? loc['libraryId'] : nil
+    end
+
+    # Check if valid service point
+    def valid_service_point_code?(service_point_code)
+      service_points.values.find { |v| v.code == service_point_code }.present?
```
jcoyne, line 119:
```suggestion
      service_points.values.any? { |v| v.code == service_point_code }
```

app/services/folio/types.rb
```diff
@@ -86,6 +88,68 @@ def get_type(type)
       JSON.parse(file.read) if file.exist?
     end
 
+    # Mapping and functions for finding specific information
+    # For FOLIO, destination is specified as service point
+    # Convert service point to library for scheduling and library hours
+    def map_to_library(service_point_code)
```
jcoyne, line 94: Can we change the name of this method to `map_to_library_code` since it returns a code and not a Library object?

jcoyne (changes requested):

app/services/folio/types.rb
hudajkhan: yes

app/helpers/pickup_libraries_helper.rb
hudajkhan: yes

app/models/symphony/pickup_destination.rb
hudajkhan: yes

app/models/symphony/pickup_destination.rb
hudajkhan: If you're referring to @type - it didn't get used so I removed it.

app/models/folio/pickup_destination.rb
hudajkhan: I removed it

.github/workflows/ruby.yml
```diff
@@ -74,6 +75,7 @@ jobs:
       run: bundle exec rake
       env:
         SETTINGS__ILS__BIB_MODEL: SearchworksItem
+        SETTINGS__ILS__PICKUP_DESTINATION_CLASS: Symphony::PickupDestination
```
jcoyne, line 78: I don't think this is necessary as it is set here: https://github.com/sul-dlss/sul-requests/pull/1638/files#diff-e769bbb8c1ba3711c5403b424ed9c218ffafba7f1890ee394717196f28ff4540R86

hudajkhan: Ok, I'll remove that one and try it. I wasn't sure what was required to get the tests to pass, and the FOLIO part was required.

app/helpers/pickup_libraries_helper.rb
hudajkhan: Since this has been changed since, I think it was just to indicate the following line needed to be executed for the Symphony condition, not the FOLIO one.

app/helpers/pickup_libraries_helper.rb
hudajkhan: I may leave this to a future pull request.

app/models/folio/request_abilities.rb
```diff
@@ -60,47 +60,39 @@ def pageable?
...
-      end.map(&:to_s)
+      location_restricted_service_point_codes
     end
 
     # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-    def default_pickup_library
-      service_points = Folio::Types.instance.service_points.select do |_k, v|
+    def default_pickup_destination
+      return 'EAST-ASIA' if request.origin_location&.match?(Regexp.union(/^EAL-SETS$/, /^EAL-STKS-/))
+
```
cbeer, line 73:
```suggestion
```

This was removed in https://github.com/sul-dlss/sul-requests/pull/1648

app/jobs/submit_folio_request_job.rb
```diff
@@ -93,13 +93,16 @@ def place_title_hold
...
-
-        Folio::Types.instance.service_points.values.find { |v| v.code == code }&.id
+        code = service_point_code(request.destination)
+        Folio::Types.instance.service_point_id(code)
       end
     end
 
+    def service_point_code(destination)
+      # Check if comparable service point code exists, otherwise return default
+      Folio::Types.instance.valid_service_point_code?(destination) ? destination : Settings.folio.default_service_point
```
cbeer, line 103: At least for the transition period, I think we probably want to fall back on the `Settings.libraries` lookup (at least to handle the backlog of requests that come in after Symphony is done and FOLIO isn't ready)

```suggestion
      Folio::Types.instance.valid_service_point_code?(destination) ? destination :  Settings.libraries[destination]&.folio_pickup_service_point_code || Settings.folio.default_service_point
```

hudajkhan: Sounds good.

app/models/folio/request_abilities.rb
hudajkhan: Noted!

app/models/concerns/request_validations.rb
```diff
@@ -18,11 +18,15 @@ module RequestValidations
   protected
 
   def destination_is_a_pickup_library
-    return if pickup_libraries.include?(destination)
+    return if check_destination?(destination)
 
     errors.add(:destination, 'is not a valid pickup library')
   end
 
+  def check_destination?(destination)
```
marlo-longley, line 26: Very minor but if you only use this method once in the codebase, potentially it could be more readable if you just directly put the check into line 21 instead of writing a method.

hudajkhan: will do!

PR #1645, opened 2023-07-28
Look for items with a permanent location

### LINKED ISSUE #1642: [sul-requests/folio-dev] NoMethodError: undefined method `callnumber' for nil:NilClass
## Backtrace

    line 9 of [PROJECT_ROOT]/app/views/application/_item_selector.html.erb: block in _app_views_application__item_selector_html_erb__3458773957260064665_40240
    line 2 of [PROJECT_ROOT]/app/views/application/_item_selector.html.erb: _app_views_application__item_selector_html_erb__3458773957260064665_40240
    line 6 of [PROJECT_ROOT]/app/views/requests/_form.html.erb: block in _app_views_requests__form_html_erb__1189377718974039696_40160

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/98561049)

jcoyne:
This provides parity with how we find relevant items in Searchworks. Fixes #1642

### DIFF
--- a/app/models/folio/item.rb
+++ b/app/models/folio/item.rb
@@ -46,10 +46,12 @@ module Folio
   #       See https://github.com/sul-dlss/searchworks_traject_indexer/blob/02192452815de3861dcfafb289e1be8e575cb000/lib/traject/config/sirsi_config.rb#L2379
   # NOTE, barcode and callnumber may be nil. see instance_hrid: 'in00000063826'
   class Item
-    attr_reader :barcode, :status, :type, :callnumber, :public_note, :effective_location, :material_type, :loan_type
+    attr_reader :barcode, :status, :type, :callnumber, :public_note, :effective_location, :permanent_location_code, :material_type,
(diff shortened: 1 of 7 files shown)


app/models/folio/item.rb
```diff
@@ -123,6 +127,7 @@ def self.from_hash(dyn)
                        dyn['chronology']].filter_map(&:presence).join(' '),
           public_note: dyn.fetch('notes').find { |note| note.dig('itemNoteType', 'name') == 'Public' }&.fetch('note'),
           effective_location: Location.from_hash(dyn.fetch('effectiveLocation')),
+          permanent_location_code: dyn.fetch('permanentLocation').fetch('code'),
```
cbeer, line 130: We also need to fall back on the holding's effective Location; we're no longer guaranteed an item-level permanent location.

jcoyne: I believe the fallback is complete.

PR #1665, opened 2023-08-03
[HOLD behind a feature flag until 8/18 at noon] Add a migration alert to hold-recall and page emails

### LINKED ISSUE #1576: Change hold/recall confirmation email to inform patrons of processing delay during shutoff
We need to inform patrons that their hold/recalls opened during the changeover period will be delayed in processing until the switch to FOLIO.

### LINKED ISSUE #1579: Handle scan requests via email to staff during Symphony shutoff
We still want to send scans to ILLiad, but instead of placing a hold, we want to send an email to a staff mailing list so that they can pick up the request, contact the patron, and handle it from there.

The staff email to use (regardless of material/location) is sulcirchelp@stanford.edu.

jcoyne:
Fixes #1576

### DIFF
--- a/app/views/request_status_mailer/request_status_for_holdrecall.html.erb
+++ b/app/views/request_status_mailer/request_status_for_holdrecall.html.erb
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
   <body>
+    <%= render 'migration_message', borrow_direct: true if Settings.features.migration %>
     <%= render 'pickup_confirmation' %>
 
     <p>
--- a/app/views/request_status_mailer/request_status_for_page.html.erb
+++ b/app/views/request_status_mailer/request_status_for_page.html.erb
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html>
   <body>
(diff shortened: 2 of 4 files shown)


app/views/request_status_mailer/request_status_for_holdrecall.html.erb
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
   <body>
+    <%= render 'migration_message', borrow_direct: true %>
```
cbeer, line 4:
```suggestion
    <%= render 'migration_message', borrow_direct: true if Settings.features.migration %>
```

Can we toss a feature flag on this and ship it now?

app/views/request_status_mailer/request_status_for_page.html.erb
```diff
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html>
   <body>
+    <%= render 'migration_message', borrow_direct: false %>
```
cbeer, line 4:
```suggestion
    <%= render 'migration_message', borrow_direct: false if Settings.features.migration %>
```

PR #1670, opened 2023-08-04
Add User#barcode to provide the patron barcode information even when …

cbeer:
…we can't talk to the ILS

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -29,6 +29,12 @@ class User < ActiveRecord::Base
     self.library_id = card_number[/\d{5}(\d+)/, 1]
   end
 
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -29,6 +29,11 @@ def sucard_number=(card_number)
     self.library_id = card_number[/\d{5}(\d+)/, 1]
   end
 
+  # Prefer the patron barcode from the ILS, but fall back to the library ID as-provided
```
jcoyne, line 32: Can we add to the comment to say "so that the system can function when the ILS is offline"?

jcoyne (approved):

PR #1678, opened 2023-08-07
Fall back on LDAP-provided univ id data when available

### LINKED ISSUE #1671: Submit Reshare requests even when the ILS is unavailable
We should be able to submit reshare requests for eligible patrons when the ILS is unavailable. One step in submitting the reshare request is validating the patron's eligibility; we're currently doing that by checking whether the ILS has a `university_id` on file.

## Backtrace

    line 157 of [PROJECT_ROOT]/app/jobs/submit_symphony_request_job.rb: patron_barcode
    line 209 of [PROJECT_ROOT]/app/jobs/submit_symphony_request_job.rb: generic_request
    line 220 of [PROJECT_ROOT]/app/jobs/submit_symphony_request_job.rb: block in place_hold_params

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/98842632)

cbeer:
We don't currently have access to the `suUnivID` field [1], so this PR is a little optimistic that we'll have that in place.  We got the OK to fall back on the library id if we don't get it.

Fixes #1671 

[1] we previously were blessed with access, but lost access with the migration to shibboleth and have to start the whole process over.

### DIFF
--- a/app/jobs/submit_iplc_listener_job.rb
+++ b/app/jobs/submit_iplc_listener_job.rb
@@ -85,7 +85,7 @@ class SubmitIplcListenerJob < ApplicationJob
     end
 
     def iplc_patron_id
-      request.user.patron.university_id
+      request.user.university_id
     end
 
     # Symphony location codes can be converted to ReShare location codes by:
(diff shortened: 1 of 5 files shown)


jcoyne (approved):

jcoyne:
@cbeer needs a rubocop fix, then feel free to merge.

PR #1683, opened 2023-08-08
Handle SAL3 coordinate libraries as aliases to SAL3

### LINKED ISSUE #1661: Handle paging schedules for coordinate libraries
Coordinate libraries have SAL3 locations in FOLIO:

e.g.
HILA-SAL3-STACKS
LANE-SAL3
LANE-SAL3X
BUS-SAL3-STACKS
BUS-SAL3-PAGE-BU
SPEC-SAL3-MSS

Currently, all paging schedules are based around the library the material is being transferred to/from. We should use the SAL3 origin paging schedule for these types of locations.

Use the origin location's `pagingSchedule` information in `details` to pick the right origin paging schedule (e.g. an origin location with the `pagingSchedule` of `SAL3` means we should ignore the location's library code and use the paging schedule code instead.

jcoyne:
Fixes #1661

### DIFF
--- a/app/models/paging_schedule.rb
+++ b/app/models/paging_schedule.rb
@@ -4,25 +4,29 @@
 #  PagingScheduleEstimate is a module that takes a requested item and can
 #  read the paging schedule configuration and estimate a date of arrival.
 ###
-module PagingSchedule
-  mattr_accessor :schedule do
-    []
-  end
+class PagingSchedule
+  class_attribute :schedule, default: []
+
+  SAL3_COORDINATES = %w[
+    HILA-SAL3-STACKS
+    LANE-SAL3
+    LANE-SAL3X
+    BUS-SAL3-STACKS
+    BUS-SAL3-PAGE-BU
+    SPEC-SAL3-MSS
+  ].freeze
 
   class << self
     def configure(&)
       instance_eval(&)
     end
 
-    def when_paging(to:, from:, before: nil, after: nil, &block)
-      schedule << Scheduler.new(to:, from:, before:, after:, &block)
+    def for(request)
+      new(request).schedule_or_default
     end
 
-    def for(request)
-      schedule_or_default = schedule_for_request(request) || default_schedule(request)
-      raise ScheduleNotFound unless schedule_or_default.present?
-
-      schedule_or_default
+    def when_paging(to:, from:, before: nil, after: nil, &block)
+      schedule << Scheduler.new(to:, from:, before:, after:, &block)
     end
 
     def worst_case_delivery_day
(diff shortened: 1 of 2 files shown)


app/models/paging_schedule.rb
```diff
@@ -4,25 +4,29 @@
...
 #  read the paging schedule configuration and estimate a date of arrival.
 ###
-module PagingSchedule
-  mattr_accessor :schedule do
-    []
-  end
+class PagingSchedule
+  class_attribute :schedule, default: []
+
+  SAL3_COORDINATES = %w[
```
cbeer, line 10: I'm confused why we're hard-coding this list. As noted on the ticket, there's an attribute in the location details we should be referencing.

app/models/paging_schedule.rb
```diff
@@ -31,26 +35,44 @@ def worst_case_delivery_day
...
+  end
+
+  private
+
+  attr_reader :request
+
+  delegate :destination, :created_at, to: :request, prefix: 'target'
+
+  def target_origin
+    @target_origin ||= SAL3_COORDINATES.include?(request.origin) ? 'SAL3' : request.origin
```
cbeer, line 58: Isn't `request.origin` the library code?

jcoyne: Yes. I'm not sure I see what you're hinting at though.

app/models/paging_schedule.rb
jcoyne: I'm not sure what you mean by "location details" can you help point me in the right direction?

cbeer: FOLIO locations have details. We've used them extensively for paging and scanning information for location-specific behaviors.

E.g. https://github.com/sul-dlss/sul-requests/blob/main/config/folio/locations.json#L312 (linked to for convenience. In practice, I think we should pull live data from graphql like we are all the other details)

We consume this as part of the graphql query:
https://github.com/sul-dlss/sul-requests/blob/main/app/services/folio_graphql_client.rb#L115-L125

The schema is tracked our graphql repo:
https://github.com/sul-dlss/folio-graphql/blob/main/src/schema.graphql#L4058-L4066

Last week, you merged the PR that registered the `pagingSchedule` attribute linked to this ticket:
https://github.com/sul-dlss/folio-graphql/pull/101

We need to populate FOLIO with this data (per the note on #1661) and consume it in this application.

app/models/paging_schedule.rb
cbeer: `LANE-SAL3` and `LANE-SAL3X` are location codes. I'm not sure why we'd have a request with the location code in that attribute.

jcoyne: Oh, I see my mistake here.

PR #1690, opened 2023-08-09
Use location details to determine the effect paging library

cbeer:


jcoyne:
Screenshot 2023-08-09 at 11 33 43 AM
It doesn't look like those details are set.  Am I looking at the right stuff?

PR #1692, opened 2023-08-09
Use fetch rather than jquery ajax

jcoyne:
jQuery is unnecessary indirection now that fetch is supported in all browsers.

### DIFF
--- a/spec/features/create_mediated_page_request_spec.rb
+++ b/spec/features/create_mediated_page_request_spec.rb
@@ -164,7 +164,7 @@ RSpec.describe 'Creating a mediated page request' do
       min_date = date_input['min']
       date_input.set(min_date)
     else
-      wait_for_ajax
+      sleep 1 # The fetch request is about to update #request_needed_date, so give it time to do so.
       min_date = find_by_id('request_needed_date', visible: :all)['min']
       page.execute_script("$('#request_needed_date').prop('value', '#{min_date}')")
     end
--- a/app/javascript/paging_schedule_updater.js
+++ b/app/javascript/paging_schedule_updater.js
@@ -50,15 +50,19 @@ var pagingScheduleUpdater = (function() {
 
     updatePagingSchedule: function(container, destination, schedulerText) {
       var _this = this;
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/paging_schedule_updater.js
```diff
@@ -48,18 +48,23 @@ var pagingScheduleUpdater = (function() {
...
+        ).then((data) => _this.updateSchedulerText(schedulerText, data)
+        ).catch(() => {
+            _this.updateSchedulerText(schedulerText, {
+              'text': 'No estimate available'
+            })
+        })
     },
 
     updateSchedulerText: function(schedulerText, data) {
+      console.log("updateSchedulerText", data)
```
thatbudakguy, line 67: probably don't want this here

spec/features/create_mediated_page_request_spec.rb
```diff
@@ -164,7 +164,7 @@ def fill_in_required_fields
       min_date = date_input['min']
       date_input.set(min_date)
     else
-      wait_for_ajax
+      sleep 1 # The fetch request is about to update #request_needed_date, so give it time to do so.
```
thatbudakguy, line 167: i went looking for info about `wait_for_ajax` and found this [somewhat interesting discussion on `wait_for_pending_requests` in capybara](https://github.com/teamcapybara/capybara/issues/2106). tl;dr there isn't currently (and won't be) a method like `wait_for_ajax`, so our options are:
1. do the above (just `sleep`)
2. add an actual class or some other change to the UI so that we know when the change has completed, and then use `.has_css?` which will wait until the change appears

the second option seems like the "nicer" one, but I guess I'm ok with this as a stopgap.

jcoyne: @thatbudakguy yes it's really a design question.  How does the user know that something changed?   They obviously aren't going to be looking in the developer console and saying "Hey, it looks like those XHRs finished."  That's the equivalent of this `wait_for_ajax` method.

thatbudakguy (approved):

PR #1694, opened 2023-08-10
Stop eagerly retrieveing dueDate for items

### LINKED ISSUE #1676: Investigate slow performance of page requests with lots of items
E.g.:
https://requests-folio-dev.stanford.edu/requests/new?item_id=L284&origin=LANE-MED&origin_location=LANE-SAL3
https://requests-folio-dev.stanford.edu/requests/new?item_id=L13&origin=LANE-MED&origin_location=LANE-COR
https://requests-folio-dev.stanford.edu/requests/new?item_id=L17&origin=LANE-MED&origin_location=LANE-COR
https://requests-folio-dev.stanford.edu/requests/new?item_id=3248266&origin=SAL3&origin_location=PAGE-MP

Hopefully the -prod environment will speed up these requests, but we should do some due diligence and:

- see where the performance bottlenecks are,
- see if there are "easy" fixes we can make immediately, and
- create tickets for larger application re-work (e.g. async loading of holdings information)

jcoyne:
Eagerly loading dueDate for every item was causing the query from the graphql server to take about 35 seconds for hrid: 'L284'.  Now the query time is under 17.5s.

Ref #1676

### DIFF
--- a/app/helpers/requests_helper.rb
+++ b/app/helpers/requests_helper.rb
@@ -47,7 +47,7 @@ module RequestsHelper
       else
         tag.span class: 'status pull-right availability unavailable' do
           # This content is lazily loaded, because the query to get due_date makes the initial query for items take a long time.
-          turbo_frame_tag "due_date_#{holding.id}", src: due_date_path(holding.id)
+          turbo_frame_tag "due_date_#{holding.id}"
         end
       end
     else
--- a/app/helpers/requests_helper.rb
+++ b/app/helpers/requests_helper.rb
@@ -37,10 +37,18 @@ module RequestsHelper
     current_request.origin
   end
 
+  # rubocop:disable Metrics/MethodLength
   def label_for_item_selector_holding(holding)
     if holding.checked_out?
-      content_tag :span, class: 'status pull-right availability unavailable' do
-        "Due #{holding.due_date}"
+      if Settings.ils.bib_model == 'SearchworksItem'
+        content_tag :span, class: 'status pull-right availability unavailable' do
+          "Due #{holding.due_date}"
+        end
+      else
+        tag.span class: 'status pull-right availability unavailable' do
+          # This content is lazily loaded, because the query to get due_date makes the initial query for items take a long time.
+          turbo_frame_tag "due_date_#{holding.id}", src: due_date_path(holding.id)
+        end
       end
     else
       content_tag :span, class: "status pull-right availability #{holding.status_class}" do
@@ -48,6 +56,7 @@ module RequestsHelper
       end
     end
   end
+  # rubocop:enable Metrics/MethodLength
 
   def request_level_request_status(request = current_request)
     if request.ils_response.usererr_code
--- a/app/services/folio_graphql_client.rb
+++ b/app/services/folio_graphql_client.rb
@@ -41,6 +41,8 @@ class FolioGraphqlClient
   end
 
   # rubocop:disable Metrics/MethodLength
+  # NOTE: While it would be tempting to get dueDate for each item in this query,
+  # we found that it adversely impacts response time when there are a lot of items (see hrid: L284)
   def instance(hrid:)
     data = post_json('/', json:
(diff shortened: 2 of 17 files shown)


cbeer:
Isn't this data used for https://github.com/sul-dlss/sul-requests/blob/main/app/helpers/requests_helper.rb#L43, rendered in the item selector list?

app/helpers/requests_helper.rb
```diff
@@ -37,17 +37,26 @@ def i18n_library_title_key
...
-      content_tag :span, class: 'status pull-right availability unavailable' do
-        "Due #{holding.due_date}"
+      if Settings.ils.bib_model == 'SearchworksItem'
+        content_tag :span, class: 'status pull-right availability unavailable' do
+          "Due #{holding.due_date}"
+        end
+      else
+        tag.span class: 'status pull-right availability unavailable' do
+          # This content is lazily loaded, because the query to get due_date makes the initial query for items take a long time.
+          turbo_frame_tag "due_date_#{holding.id}", src: due_date_path(holding.id)
```
cbeer, line 50: Should we be concerned about triggering hundreds or thousands of separate requests on our problem records? I thought turbo eagerly loaded frames by default?

jcoyne: It will, but this is only when they are checked out.  Do we have many instances with lots of items and we expect them all to be checked out at once?

cbeer: Our big offenders (from SAL3) at the moment:
https://searchworks.stanford.edu/view/9984192 has 49
https://searchworks.stanford.edu/view/487166 has 27
https://searchworks.stanford.edu/view/10034758 has 24

There's a pretty good smattering of 10-20 checked out items too.

I guess given the number of API calls the FOLIO UI itself makes, it's probably fine??

cbeer: I tested this with 9984192. It takes a long time for the due dates to come in (I see requests made immediately, but nothing updates for ~10-15s). I'm surprised how much longer this is compared to the RTAC response we make to load due dates on https://searchworks-folio-dev.stanford.edu/view/9984192 (~0.5s)

jcoyne: I'm not sure what you want me to do.  I think this helps us draw the initial page faster, which was the intention of #1676, right?

cbeer: I don't think the behavior demonstrated above is acceptable. I think we could look at why it takes so long for turbo to load the due date information, look at using RTAC to asynchronously load the data (as we're doing in Searchworks and seems fast enough), or consider some option.

jcoyne: Updated to do just a single request for all items. This gets all the dates in 6s rather than 2-3 seconds per row.

cbeer: I'm surprised it still takes 6s. Searchworks does it in <1s. Maybe @corylown can help with using RTAC.

jcoyne: @cbeer might it be because we're proxying through the graphql server and not hitting the folio server directly?

app/services/folio_graphql_client.rb
```diff
@@ -152,6 +154,31 @@ def instance(hrid:)
     data&.dig('data', 'instances', 0)
   end
 
+  def due_date(instance_id:)
+    data = post_json('/', json:
+      {
+        query:
+          <<~GQL
+            query DueDate {
+              availability(id: "#{instance_id}") {
```
cbeer, line 163: Is RTAC fast enough that we can just do this in the original query?

jcoyne: It's still adds significant slowness, especially in the worst cases.

jcoyne:
The RTAC call is now 0.9s on https://requests-folio-dev.stanford.edu/pages/new?item_id=9984192&origin=SAL3&origin_location=PAGE-GR

PR #1704, opened 2023-08-11
Update permanent location logic to match other apps

### LINKED ISSUE #1701: [sul-requests/folio-dev] NoMethodError: undefined method `permanent_location' for nil:NilClass
## Backtrace

    line 58 of [PROJECT_ROOT]/app/models/request.rb: paging_origin_library
    line 1 of [PROJECT_ROOT]/app/views/requests/_destination_selector.html.erb: _app_views_requests__destination_selector_html_erb__3832463460399578679_38900
    line 2 of [PROJECT_ROOT]/app/views/requests/_destination_and_needed_date.html.erb: _app_views_requests__destination_and_needed_date_html_erb__3694913886403541007_38880

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/99398741)

cbeer:
Fixes #1701

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -55,7 +55,7 @@ class Request < ActiveRecord::Base
     return origin unless bib_data.is_a? Folio::Instance # TODO: remove this line post Folio migration
 
     # items are already limited to a single permanent location, so we can just grab the first one
-    bib_data.items.first.permanent_location.details['pagingSchedule'] || origin
+    bib_data.items.first&.permanent_location&.details&.dig('pagingSchedule') || origin
   end
 
   def library_location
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

app/models/request.rb
```diff
@@ -55,7 +55,7 @@ def paging_origin_library
     return origin unless bib_data.is_a? Folio::Instance # TODO: remove this line post Folio migration
 
     # items are already limited to a single permanent location, so we can just grab the first one
-    bib_data.items.first.permanent_location.details['pagingSchedule'] || origin
+    bib_data.items.first&.permanent_location&.details&.dig('pagingSchedule') || origin
```
jcoyne, line 58: I understand the first `&.`, but is there a situation where the other two are needed? Don't we always have a details object for a permanentLocation?

cbeer: Isn't this just how `&.` works?

jcoyne: okay, so it's chaining for when there are no items rather than specifically protecting against those other structures being nil?

PR #1708, opened 2023-08-11
[refactor] Use a more descriptive method name

jcoyne:


app/services/folio/types.rb
```diff
@@ -46,42 +46,42 @@ def circulation_rules
...
+      @libraries ||= load_cache(type: 'libraries').index_by { |p| p['id'] }
     end
 
     def locations
-      @locations ||= get_type('locations').index_by { |p| p['id'] }
+      @locations ||= load_cache(type: 'locations').index_by { |p| p['id'] }
     end
 
-    def get_type(type)
+    def load_cache(type:)
```
cbeer, line 84: I personally like the current name more than `load_cache`. I also don't know why we want a keyword arg here.

marlo-longley (approved):

app/services/folio/types.rb
marlo-longley: Oops. I thought it looked OK but happy to defer / discuss.

jcoyne: `load_cache` tells me what the method is doing. It's populating a cache from the files on disk.  `get_type` on the other hand is two very ambigous words. "get" from where?  "type" of what?

cbeer: I don't think being cached is the important part of this. We're just trying to load FOLIO type data.

PR #1709, opened 2023-08-11
Replace turbolinks with @hotwired/turbo

jcoyne:


thatbudakguy (approved):
tried a few requests and everything looks right!

PR #1714, opened 2023-08-11
Extract logic about querying service points to a new class

jcoyne:


cbeer:
Is this going to make a nasty conflict with #1638?

cbeer:
I'd selfishly like to get #1725 in first; we've done similar refactoring, but I think it'll be easier to rebase these changes on that instead of the other way around.

cbeer:
I've merged this into https://github.com/sul-dlss/sul-requests/pull/1725.

PR #1727, opened 2023-08-15
Extract a DestinationSelectorComponent

jcoyne:
From https://github.com/sul-dlss/sul-requests/pull/1638\#discussion_r1293674040


app/components/destination_selector_component.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <div class='form-group'>
+      <div class='<%= helpers.label_column_class %> control-label'>
+        Will be delivered to
+      </div>
+      <div class='<%= helpers.content_column_class %> input-like-text'>
+        <%= destination_label(pickup_destinations.first) %>
+      </div>
+      <%= form.hidden_field :destination, value: pickup_destinations.first %>
+    </div>
+  <% elsif pickup_destinations.many? # Symphony %>
```
cbeer, line 12:
```suggestion
  <% elsif pickup_destinations.many? %>
```

I'm not sure what this comment means? It seems like this scenario is possible with FOLIO too, isn't it?

jcoyne: I didn't understand this comment either and I remarked upon that here: https://github.com/sul-dlss/sul-requests/pull/1638#discussion_r1291268738. In this case I'm just copying what was already there.

jcoyne: Fixed in #1730

app/views/shared/_request_status_information.html.erb
```diff
@@ -9,7 +9,7 @@
   <% destination_name = LibraryLocation.library_name_by_code(current_request.destination) %>
   <% if destination_name.present? %>
     <dt>
-      <%= label_for_pickup_destinations_dropdown(current_request.pickup_destinations) %>
+      Will be delivered to
```
cbeer, line 12: Is this (and the test changes) an intentional change from the current behavior? I think the subtle intent is to indicate whether the patron has any choice about the delivery location, and it seems like we've lost that?

cbeer (dismissed):
See code comment

app/views/shared/_request_status_information.html.erb
jcoyne: This seemed weird that the status display would display two different labels "Will be delivered to" and "Delivered to" depending on whether or not there was a choice.  The function that was used here was intended as a label for a dropdown, but this is not a dropdown.  


Do you want to retain this behavior? I can revert this part, but it just seems wrong.

cbeer: I would err on the side of keeping the current behavior. It was important at the time and I think there's a UX argument to be made about using consistent labels between the form and the email.

If we want to change it, I think it's important to do so transparently and not in a refactoring PR.

jcoyne:
@cbeer reverted the change on the summary page.

PR #1750, opened 2023-08-21
Disable honeybadger notifications for Symphony in migration mode

cbeer:
Taking inspiration from https://github.com/sul-dlss/mylibrary/pull/917, I think these honeybadger notices are all expected and just noise at this point 🤷‍♂️


jcoyne:
Rubocop errors prevented CI from passing.

Issue #1770, opened 2023-08-29
[sul-requests/prod] NoMethodError: undefined method `library_id' for nil:NilClass


cbeer:
## Backtrace

    line 39 of [PROJECT_ROOT]/app/models/folio/service_point.rb: library_id
    line 29 of [PROJECT_ROOT]/app/models/folio/service_point.rb: library
    line 30 of [PROJECT_ROOT]/app/models/folio/pickup_destination.rb: library_code

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/99760209)

hudajkhan:
In the case of RWC, the FOLIO pickup destination class tries to look up library code.  RWC is a service point with no associated location or library.  
https://github.com/sul-dlss/sul-requests/blob/main/app/models/folio/pickup_destination.rb#L30 calls 
https://github.com/sul-dlss/sul-requests/blob/main/app/models/folio/service_point.rb#L28 to get the library code
which calls https://github.com/sul-dlss/sul-requests/blob/main/app/models/folio/service_point.rb#L28 to get the library id from the location associated with this service point.

Since there is no location, the object is nil, and trying to get the library_id from a nil object returns an error.

hudajkhan:
Relates to https://github.com/sul-dlss/sul-requests/tree/handleNoLibrary

PR #1774, opened 2023-08-29
handling service points without associated library or location

hudajkhan:
Relates to https://github.com/sul-dlss/sul-requests/issues/1770 . 

For a service point like RWC, which has no associated library or location, the service point code which looks up library was running into an error because it was trying to get library_id on the location and the location returned was nil. This update ensures we don't run into this error.

### DIFF
--- a/app/models/folio/service_point.rb
+++ b/app/models/folio/service_point.rb
@@ -39,11 +39,7 @@ module Folio
 
     def library_id
       # Not every service point, e.g .RWC, has an associated location or library
-      location = Folio::Types.locations.find_by(primary_service_point_id: id)
-
-      return if location.nil?
-
-      @library_id ||= location.library_id
+      @library_id ||= Folio::Types.locations.find_by(primary_service_point_id: id)&.library_id
     end
   end
 end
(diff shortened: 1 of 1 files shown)


app/models/folio/service_point.rb
```diff
@@ -36,7 +38,12 @@ def pickup_location?
...
 
     def library_id
-      @library_id ||= Folio::Types.locations.find_by(primary_service_point_id: id).library_id
+      # Not every service point, e.g .RWC, has an associated location or library
+      location = Folio::Types.locations.find_by(primary_service_point_id: id)
+
+      return if location.nil?
+
+      @library_id ||= location.library_id
     end
```
thatbudakguy, line 47: i think doing this will get rid of the memoization, since the `find_by` call will always happen...maybe?
it might work to do:
```ruby
@library_id ||= do
    location = Folio::Types.locations.find_by(primary_service_point_id: id)
    location&.library_id
end
```

hudajkhan: Used Folio::Types.locations.find_by(primary_service_point_id: id)&.library_id
(Chris also had that comment in slack)

PR #1780, opened 2023-08-30
Excise Symphony/Searchworks models

cbeer:


hudajkhan (approved):

hudajkhan:
Ran the code locally and put in a regular request and I could see that worked fine.  Many of the changes constitute the deletion of Symphony-specific models or tests.

PR #1816, opened 2023-09-05
adding beep to the circulation check failure screen

hudajkhan:
- Adds sound from original site to assets audios directory
- Adds audio error sound autoplay when the user places an incorrect request (e.g. barcode does not exist) or when the item is not checked out ("failure" screen).


hudajkhan:
Audio autoplay policies may have changed in Chromium browsers where autoplay may not happen.  Factors may include if the user has already played video/audio in the browser, etc. I followed some of the suggestions here (https://www.tutorialspoint.com/how-to-autoplay-audio-on-chrome) - i.e. adding preload etc.  When I tried this out locally on my Chrome browser, it worked.

hudajkhan:
For additional reference: https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide .  We should be ok b/c this is the policy "As a general rule, you can assume that media will be allowed to autoplay only if at least one of the following is true:

    The audio is muted or its volume is set to 0
    The user has interacted with the site (by clicking, tapping, pressing keys, etc.)
    If the site has been allowlisted; this may happen either automatically if the browser determines that the user engages with media frequently, or manually through preferences or other user interface features
    If the autoplay [Permissions Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Permissions_Policy) is used to grant autoplay support to an [<iframe>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) and its document."
Since the user is clicking/submitting, we should be ok with the autoplay.

PR #1818, opened 2023-09-05
Use the permanent location to determine request abilities properties

cbeer:


jcoyne:
Can you say more about why do we need this change?

cbeer:
If an item is supposed to be in e.g. SPEC (because of the permanent location), it should stay e.g. Aeon pageable even if the item is temporarily out for repair.

jcoyne:
Thank you for the details.

PR #1830, opened 2023-09-08
Folio code

jcoyne:


config/settings/test.yml
```diff
@@ -2,12 +2,12 @@ super_admin_groups: ['FAKE-TEST-SUPER-ADMIN-GROUP']
...
 origin_admin_groups:
-  FAKE-ORIGIN-LIBRARY: ['FAKE-ORIGIN-LIBRARY-TEST-LDAP-GROUP']
+  EDUCATION: ['FAKE-ORIGIN-LIBRARY-TEST-LDAP-GROUP']
   ART: ['ART-TEST-LDAP-GROUP']
 
 origin_location_admin_groups:
-  FAKE-ORIGIN-LOCATION: ['FAKE-ORIGIN-LOCATION-TEST-LDAP-GROUP']
-  PAGE-MP: ['PAGE-MP-TEST-LDAP-GROUP']
+  SAL3-PAGE-MP: ['FAKE-ORIGIN-LOCATION-TEST-LDAP-GROUP']
+  EAR-PAGE-MP: ['PAGE-MP-TEST-LDAP-GROUP']
```
jcoyne, line 10: This change needs to be made in shared configs on the servers too.

config/locales/en.yml
```diff
@@ -124,8 +124,27 @@ en:
   mediated_pages:
     header:
       default: 'Request on-site access'
-      HOPKINS: 'Request delivery to campus library'
-      PAGE-MP: 'Request delivery to campus library'
+      HOPKINS: 'Request delivery to campus library' # Deprecated
+      PAGE-MP: 'Request delivery to campus library' # Deprecated
+      EAR-PAGE-MP: 'Request delivery to campus library'
+      SAL3-PAGE-MP: 'Request delivery to campus library'
+      MAR-ATLASES: 'Request delivery to campus library'
```
cbeer, line 131: I hope we can avoid playing new location whack-a-mole if/when Hopkins adds more locations; is there any way we can continue to use the library code?

config/settings.yml
```diff
@@ -28,22 +28,22 @@ scan_pilot_groups:
...
-  HOOVER: ['sul:requests-hoover-library']
-  RUMSEYMAP: ['sul:requests-rumsey']
+  MARINE-BIO: ['sul:requests-hopkins']
+  HILA: ['sul:requests-hoover-archive']
+  RUMSEY-MAP: ['sul:requests-rumsey']
   SPEC-COLL: ['sul:requests-spec']
 
 origin_location_admin_groups:
-  PAGE-MP: ['sul:requests-branner']
+  EAR-PAGE-MP: ['sul:requests-branner']
```
cbeer, line 37:
```suggestion
  SAL3-PAGE-MP: ['sul:requests-branner']
```

I thought EAR-PAGE-MP was the "prepare items to go to SAL3", not the "items are at SAL3" location. I made up this location code, so you should  verify wit with real-world items.

app/controllers/concerns/request_strong_params.rb
```diff
@@ -4,25 +4,40 @@
...
+  # rubocop:disable Metrics/AbcSize
   def new_params
-    params.require(:origin)
-    params.require(:item_id)
-    params.require(:origin_location)
+    if params[:origin] || params[:origin_location]
+      # If you have one of these you must provide both
+      params.require(:origin) # Legacy symphony library code
+      params.require(:origin_location) # Legacy symphony location code
+      params[:location] = FolioLocationMap.folio_code_for(library_code: params[:origin], home_location: params[:origin_location])
```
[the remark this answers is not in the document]
jcoyne: No. Do we still need to consider that?  Isn't our SW index built off folio data, where all the conversions have been made already?

config/settings.yml
jcoyne: I don't understand what you mean.  There are two PAGE-MP locations: SAL3- and EAR-.  My read here was that since we were asking for an admin group called "requests-branner" we'd want to use EAR-.

app/models/request.rb
```diff
@@ -51,17 +52,22 @@ class Request < ActiveRecord::Base
     self.item_title ||= bib_data&.title
   end
 
+  after_find do
+    # This converts stored records so that they have a location set.
+    self.location ||= FolioLocationMap.folio_code_for(library_code: origin, home_location: origin_location)
+  end
```
thatbudakguy, line 58: could we just do this with a migration?

app/helpers/requests_helper.rb
```diff
@@ -23,19 +25,21 @@ def i18n_location_title_key
...
+        current_request.location.presence
       else
-        current_request.origin_location.presence
+        current_request.origin_location.presence # Deprecated path
       end
+    elsif current_request.location.present?
+      current_request.location.presence
     elsif (origin_location = current_request.origin_location).present?
-      origin_location
+      origin_location # Deprecated path
```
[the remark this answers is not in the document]
jcoyne: Yes, precisely.

app/models/request.rb
jcoyne: Yes, but I don't like doing data migrations in automated database migrations.  It's possible we get through 90 pct of the records and find something that blocks the migration (perhaps a location that isn't in the migration table.), so I think we should do that apart from this code change.

config/settings.yml
jcoyne: I've changed all instances to SAL3-PAGE-MP, as I guess that is the correct one?

config/locales/en.yml
jcoyne: Okay, I reverted this.  I thought this way was less complex. 🤷

app/models/request.rb
thatbudakguy: is the general process that you just delete this line after some time has passed? how do we know that everything has been updated?

jcoyne: Yes, after you do `Request.find_each { |r| r.save! }`, then all the data is migrated, then we can remove this line.

PR #1834, opened 2023-09-08
Import the ldap authz mapping from shared configs

jcoyne:
This allows us to test the code as configured.  This means less variation is required between the managed/tested code and the code/config that is deployed


thatbudakguy:
Do we know that there wasn't an opsec-type reason for not exposing these? Are we sure it's fine to have them be public?

jcoyne:
@thatbudakguy not that I know of. We have ldap groups listed in other public repos: https://github.com/sul-dlss/argo/blob/fab16388fc3517cc875d3304157270f9fd524733/app/models/user.rb#L9-L12C13

thatbudakguy (approved):

Issue #1776, opened 2023-08-29
Show + handle temporary locations


cbeer:
This item has 2 items that are on course-reserve. The items on course reserve should not be pageable (because their current circ rule doesn't allow it)

https://requests.stanford.edu/pages/new?item_id=13956859&origin=LAW&origin_location=STACKS-1

I believe the current implementation is basically “look at what the item status has to say about its availability”. The app  groups items by their permanent location (because searchworks does it).. and usually permanent location is good enough to figure out the circ rules.

However:
if an item has an effective location/material type/etc that doesn’t allow paging, holds or recalls (under the circ policy…), we need to:
- show the temporary location (just/kinda like we do in searchworks)
- disable the checkbox for the item
- ??

thatbudakguy:
Should the UI for this look similar to how it does for checked out items, except that you can't actually check the box? Or do we not need the ability to expand the status by clicking on it?

![Screenshot 2023-08-30 at 11 10 42](https://github.com/sul-dlss/sul-requests/assets/4924494/5b61f7a7-cf8b-4e60-b753-59b22fada7ae)

PR #1837, opened 2023-09-08
Refactor requestability to delegate to the item

thatbudakguy:
This PR fixes a couple of inconsistencies (typos, issues with fixtures) and does one main refactor: making `Folio::Item` the source of truth for requestability via various methods. This means that:
- `RequestAbilities` can use the items in the request to figure out whether the overall request is viable
- `SubmitFolioRequestJob` can use the items in the request to figure out how to submit the job
- The item selector view can use the items to figure out which ones should be requestable/checkable

As a result, the tests for `SubmitFolioRequestJob` were significantly expanded and check for more things (like whether the request is actually the right type, and who it is for).

This came out of/is pre-work for #1776, because we will need to do things like enable/disable request checkboxes and set the status text/icons on a per-item basis using this logic.

### DIFF
--- a/app/models/folio/item.rb
+++ b/app/models/folio/item.rb
@@ -101,10 +101,12 @@ module Folio
       [availability_class, circ_class].compact.join(' ')
     end
 
-    def status_text
-      if !circulates?
-        'In-library use only'
-      elsif status == STATUS_AVAILABLE
+    def status_text(request)
+      if !requestable && temporary_location.present?
+        temporary_location.discovery_display_name
+      elsif !circulates?
+        'In-library use'
+      elsif status == STATUS_AVAILABLE && requestable_with?(request)
         'Available'
       else
         'Unavailable'
@@ -151,7 +153,7 @@ module Folio
                                end) || (Location.from_hash(dyn.dig('holdingsRecord', 'effectiveLocation')) if dyn.dig('holdingsRecord',
                                                                                                                       'effectiveLocation')),
           material_type: MaterialType.new(id: dyn.dig('materialType', 'id'), name: dyn.dig('materialType', 'name')),
-          loan_type: LoanType.new(id: dyn.fetch('tempooraryLoanTypeId', dyn.fetch('permanentLoanTypeId'))))
+          loan_type: LoanType.new(id: dyn.fetch('temporaryLoanTypeId', dyn.fetch('permanentLoanTypeId'))))
     end
     # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
--- a/app/models/folio/item.rb
+++ b/app/models/folio/item.rb
@@ -101,12 +101,10 @@ module Folio
       [availability_class, circ_class].compact.join(' ')
     end
 
-    def status_text(request)
-      if !requestable && temporary_location.present?
-        temporary_location.discovery_display_name
-      elsif !circulates?
-        'In-library use'
-      elsif status == STATUS_AVAILABLE && requestable_with?(request)
+    def status_text
+      if !circulates?
+        'In-library use only'
+      elsif status == STATUS_AVAILABLE && requestable?
         'Available'
       else
         'Unavailable'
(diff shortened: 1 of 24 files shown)


app/models/folio/item.rb
```diff
@@ -132,12 +135,47 @@ def paged?
     end
 
     def hold_recallable?
-      HOLD_RECALL_STATUSES.include?(status)
+      recallable? || holdable?
     end
 
-    # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+    def recallable?
+      HOLD_RECALL_STATUSES.include?(status) && allowed_request_types.include?('Recall')
```
jcoyne, line 142: allowed_request_types could be nil, so you can't send `include?` to it without checking for nil.

app/models/folio/item.rb
```diff
@@ -132,12 +135,47 @@ def paged?
...
+    end
+
+    def holdable?
+      HOLD_RECALL_STATUSES.include?(status) && allowed_request_types.include?('Hold')
```
jcoyne, line 146: allowed_request_types could be nil, so you can't send `include?` to it without checking for nil.

app/models/folio/item.rb
```diff
@@ -132,12 +135,47 @@ def paged?
...
+    end
+
+    def pageable?
+      PAGEABLE_STATUSES.include?(status) && allowed_request_types.include?('Page')
```
jcoyne, line 150: allowed_request_types could be nil, so you can't send `include?` to it without checking for nil.

jcoyne (changes requested):

app/models/folio/item.rb
thatbudakguy: Added a check to make it `[]` if nil

jcoyne (approved):

cbeer (changes requested):
I think (someone) should take this for a spin locally to do some UAT before we merge this. (Using Request changes to block merge for now). Some scenarios:

- [ ] On-campus paging (e.g. from BUSINESS)
- [x] Location-restricted paging (e.g. SAL3-PAGE-SP)
- [x] Hold/recall from a non-pageable location
- [ ] Hold/recall from a pageable location
- [ ] Hold/recall for an in-process item
- [ ] Title-level hold

cbeer:
I'm seeing a new stack trace paging a non-existing item:

> undefined method `items' for nil:NilClass

(Coming from Request#paging_origin_library).

I don't think we need to handle that scenario gracefully, but maybe we can do better than a NoMethodError.

cbeer:
I'm also getting a `PagesController::UnpageableItemError` error paging a checked out item from BUS-SAL3-STACKS (e.g. https://searchworks-folio-test.stanford.edu/view/10000041). It might be a circ rule thing, but I think this should either be a sent down the hold/recall path or use the paging form (but submit a hold-recall).

cbeer:
https://searchworks.stanford.edu/view/13956859 is now in a weird state too; maybe the problem is we're showing a request button, but it also gives a `PagesController::UnpageableItemError` (although I thought Missing was perhaps hold/recall-able?)

thatbudakguy:
> I'm seeing a new stack trace paging a non-existing item:

I added a pathway where this raises a 404.

> I'm also getting a PagesController::UnpageableItemError error paging a checked out item from BUS-SAL3-STACKS (e.g. https://searchworks-folio-test.stanford.edu/view/10000041).

This thing is `Aged to lost`, which FOLIO [doesn't let you request at all](https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/RequestTypeItemStatusWhiteList.java#L71-L94). For now, it seems like we need to suppress the request button in SearchWorks.

> https://searchworks.stanford.edu/view/13956859 is now in a weird state too; maybe the problem is we're showing a request button, but it also gives a PagesController::UnpageableItemError (although I thought Missing was perhaps hold/recall-able?)

Added a check so that in this case (no barcode sent, one holdable but not pageable item, two non-requestable items) we fall back to assuming the request is a Page and let the items figure it out after the user selects one.

cbeer (approved):

PR #1846, opened 2023-09-14
Item selector UX

thatbudakguy:
This brings in a bunch of changes to make the UX for the item selector more predictable.

- disable checkboxes for items that aren't requestable
- if the item would be a hold/recall, make it a warning-y color and show a warning icon because we don't know when you'll be able to get it
- if the item would be a hold/recall, show its current status so you know if it's e.g. missing, waiting for someone else to pick it up, in transit, etc.
- if the item isn't requestable because it's in a temporary location (e.g. put on course reserve), show that location's display name so that you know that it's on reserve

Example:

Screenshot 2023-09-14 at 17 13 54

### DIFF
--- a/app/models/folio/item.rb
+++ b/app/models/folio/item.rb
@@ -116,6 +116,8 @@ module Folio
         'In-library use only'
       elsif status == STATUS_AVAILABLE && requestable?
         'Available'
+      elsif hold_recallable?
+        status
       else
         'Unavailable'
       end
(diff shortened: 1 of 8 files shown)


app/models/folio/item.rb
```diff
@@ -110,10 +109,14 @@ def status_class
     end
 
     def status_text
-      if !circulates?
+      if !requestable? && temporary_location.present?
+        temporary_location.discovery_display_name
+      elsif !circulates?
```
cbeer, line 114: I'm not sure "Unavailable" is quite the right text here (the item status "Aged to lost" is the most likely item status that results in it not being requestable, which is.. worse), but 🤷‍♂️ 

```suggestion
      return temporary_location.discovery_display_name || 'Unavailable' if !requestable?

      if !circulates?
```

cbeer, line 114:
```suggestion
      return temporary_location&.discovery_display_name || 'Unavailable' if !requestable?

      if !circulates?
```

PR #1854, opened 2023-09-19
Send the whole exception to honeybadger for better stack traces and debugging

cbeer:


hudajkhan:
Based on this: https://docs.honeybadger.io/lib/ruby/getting-started/reporting-errors/, this should work and hopefully provide more info than is currently being reported.

hudajkhan (approved):

PR #1865, opened 2023-09-21
Prune/rename Symphony library codes from settings where no longer needed

hudajkhan:
What this pull request does:

Changes to the Settings file:
- For any library codes in the Settings file that have FOLIO codes but which also have additional information needed from Settings, this pull request removes the label information as the name should come from FOLIO.  The extra information not in FOLIO corresponds to the following properties : instructions, hold_pseudopatron, iplc_pickup_location_code, contact_info, hours, and reading_room_url. 
- For library codes that have been renamed, but which still need some of the extra info, the older Symphony code version has been removed and any info needed is maintained under the new FOLIO library code mapping.  These libraries are: MARINE-BIO (renamed from Hopkins), LANE (renamed from LANE-MED), MEDIA-CENTER (renamed from MEDIA-MTXT)and RUMSEY-MAP (renamed from RUMSEYMAP).
- Library codes with no equivalent FOLIO codes which are still needed remain as they were.  These are: RWC (which is not a library in FOLIO but which still needs its "folio_service_pickup_point" information as there is no other way to retrieve the related service point) and "SCAN".
- Library codes that are no longer used at all by Requests have been removed entirely.  These are CHEMCHENG, LATHROP, BIOLOGY, and MATH-CS.  Besides Lathrop, the other three codes no longer correspond to existing physical buildings.

Changes to code related to using FOLIO library data to retrieve the library name
- The FOLIO library model has been updated to also read in the "name" field from libraries.json and to return that info in an attribute. 
- LibraryLocation has been updated to look first for the FOLIO information, and if it doesn't exist, resort to the label in Settings. 
- The info model and request status views have been updated to use LibraryLocation to get the name for the library as opposed to relying on the label from Settings.

### DIFF
--- a/app/models/library_location.rb
+++ b/app/models/library_location.rb
@@ -24,18 +24,20 @@ class LibraryLocation
   end
 
   class << self
+    # Most library codes used in requests will be FOLIO codes
+    # Settings will be used for the RWC code which has a service point but no FOLIO library associated
     def library_name_by_code(code)
-      all_libraries[code]&.label
+      Folio::Types.libraries.find_by(code:)&.name || Settings.libraries[code]&.label
     end
 
     # This is a super-clunky way to convert data from RailsConfig to something
     # Enumerable, so we can use e.g. #select
-    def all_libraries
+    def all_settings_libraries
       Settings.libraries.map.to_h.with_indifferent_access
     end
 
     def pageable_libraries
-      all_libraries.select { |_, v| v.pageable }
+      all_settings_libraries.select { |_, v| v.pageable }
     end
   end
 end
--- a/app/models/folio/library.rb
+++ b/app/models/folio/library.rb
@@ -3,15 +3,16 @@
 module Folio
   # Model for FOLIO's library data
   class Library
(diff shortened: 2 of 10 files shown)


hudajkhan:
Also, tests have been updated to pass in a name for libraries.  Of note, pageable libraries still require Settings to produce their list.

app/models/library_location.rb
```diff
@@ -24,18 +24,20 @@ def folio_location_code
   end
 
   class << self
+    # Most library codes used in requests will be FOLIO codes
+    # Certain libraries may need fallback info in Settings
```
jcoyne, line 28: Which ones in particular?  Why?

hudajkhan: I'll add the example to the comment.

jcoyne (approved):

hudajkhan:
> @hudajkhan would you mind condensing this into a single commit, or do you think that preserving the history here would provide some benefit to readers in the future?

Done. Thanks.

jcoyne (approved):

PR #1876, opened 2023-09-28
Remove SymphonyResponse

jcoyne:


cbeer:
Won't this break the status + mediation pages for pre-FOLIO requests?

jcoyne:
@cbeer I don't think so. It still loads the symphony data from the database.

PR #1879, opened 2023-09-28
Make :name argument optional in Folio::Library

### LINKED ISSUE #1877: [sul-requests/stage] ArgumentError: missing keyword: :name
## Backtrace

    line 12 of [PROJECT_ROOT]/app/models/folio/library.rb: initialize
    line 11 of [PROJECT_ROOT]/app/models/folio/location.rb: new
    line 11 of [PROJECT_ROOT]/app/models/folio/location.rb: from_hash

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/100689554)

corylown:
Fixes #1877

When we instantiate `Folio::Library` with data from the FOLIO API (`FolioClient.new.find_instance_by(hrid: 'a3978751')`) the library information on the item contains only an `:id` and a `:code` so `:name` needs to be optional.

### DIFF
--- a/app/models/folio/library.rb
+++ b/app/models/folio/library.rb
@@ -9,7 +9,7 @@ module Folio
       new(id: dyn.fetch('id'), code: dyn.fetch('code'), name: dyn.fetch('name'))
     end
 
-    def initialize(id:, code:, name:)
+    def initialize(id:, code:, name: nil)
       @id = id
       @code = code
       @name = name


jcoyne:
@corylown what does that mean? How is it at a nameless library?

corylown:
@jcoyne the library data we get back from FOLIO that's included in an instance record's item metadata contains only an id and a code: `"library"=> {"id"=>"f6b5519e-88d9-413e-924d-9ed96255f72e", "code"=>"GREEN"}`. I think that's all that's needed in this case to validate the request type.

You can trace the calls from here: https://github.com/sul-dlss/sul-requests/blob/main/app/models/folio/instance.rb#L12 where the app passes instance record data down to Folio::Item, Folio::Location, then Folio::Library.

Issue #1889, opened 2023-10-02
Direct all hold/recall requests to ILLiad instead of ReShare or FOLIO


saseestone:
Access Services has been working with SUL branches and coordinates on a plan to direct all hold/recall requests to the SUL ILLiad environment for fulfillment. They are ready to move forward. 

ILLiad will determine the best course of action for the request (fulfillment via BorrowDirect, UC agreements, other ILL agreements, or placing a hold FOLIO) 

There are some preliminary docs that have been prepared to map SearchWorks fields with ILLiad fields: https://docs.google.com/spreadsheets/d/1bvMuOL4xDjAlXl-QjpsTWHf8Lc3icB7O3ED3FDmKvTQ/edit#gid=0

Work will also need to be done in My Library. An older ticket on that effort: https://github.com/sul-dlss/mylibrary/issues/300

saseestone:
@thatbudakguy We haven't talked about how the request dashboard will look when requests are directed to ILLiad.  Can you confirm:

- Hold/recall requests will continue to be added to the request dashboard
- Will the status page for the request include information on the ILLiad response?

thatbudakguy:
I will double-check, but to my knowledge nothing around this has changed, so it should all "just work". The functionality to show the ILLiad response will still be there, and the ILLiad response data should correctly get added when a Hold/Recall is sent to ILLiad.

PR #1898, opened 2023-10-05
Refactor to move ILLiad param generation into Request subclasses

thatbudakguy:
This creates a seam for having different request types send their
data to ILLiad (not just scans) and expands the test coverage for
existing scan requests to ILLiad, in prep for #1889.

### DIFF
--- a/app/models/illiad_request.rb
+++ b/app/models/illiad_request.rb
@@ -4,54 +4,24 @@
 #  Class to handle creation of ILLiad OpenURL request
 ###
 class IlliadRequest
-  def initialize(scan)
-    @scan = scan
+  def initialize(request)
+    @request = request
   end
-
-  # rubocop:disable Metrics/MethodLength
-  def illiad_transaction_request
-    {
-      ProcessType: 'Borrowing',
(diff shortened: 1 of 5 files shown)


app/models/illiad_request.rb
```diff
@@ -4,54 +4,24 @@
...
-      Username: @scan.user.sunetid
-    }.to_json
-  end
-  # rubocop:enable Metrics/MethodLength
-
   def request!
-    faraday_conn_w_req_headers.post('ILLiadWebPlatform/Transaction/', illiad_transaction_request)
+    faraday_conn_w_req_headers.post(
+      'ILLiadWebPlatform/Transaction/',
+      @request.illiad_request_params.merge(default_params).to_json
```
cbeer, line 14:
```suggestion
      @request.illiad_request_params.reverse_merge(default_params).to_json
```

If they're default params, presumably the request-specific params should win?

thatbudakguy: Thank you! I'll just merge the other direction; got the order wrong

PR #1905, opened 2023-10-09
Use `LibraryLocation` and other helpers to figure out request origin information

cbeer:
Restores most of the updates from #1865

### DIFF
--- a/app/models/library_location.rb
+++ b/app/models/library_location.rb
@@ -26,7 +26,7 @@ class LibraryLocation
 
   class << self
     def library_name_by_code(code)
-      all_libraries[code]&.label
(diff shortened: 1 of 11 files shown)


app/models/library_location.rb
```diff
@@ -26,7 +26,7 @@ def folio_location_code
 
   class << self
     def library_name_by_code(code)
-      all_libraries[code]&.label
+      Folio::Types.libraries.find_by(code:)&.name(fallback_value: nil) || all_libraries[code]&.label
```
jcoyne, line 29: Can `all_libraries[code]&.label` be the fallback value here?

cbeer: No? `Folio::Types.libraries.find_by(code:)` could be nil.

jcoyne (approved):

PR #1913, opened 2023-10-12
Send Hold/Recall requests to ILLiad

### LINKED ISSUE #1889: Direct all hold/recall requests to ILLiad instead of ReShare or FOLIO

thatbudakguy:
Closes #1889 

- Support sending Hold/Recall requests to ILLiad
- Convert the Scan request job into a generic ILLiad request job
- Extract an Illiadable concern for mixing into Requests
- Make HoldRecall Illiadable and create feature flag

shared_configs PRs to turn it on:
https://github.com/sul-dlss/shared_configs/pull/2087
https://github.com/sul-dlss/shared_configs/pull/2088
https://github.com/sul-dlss/shared_configs/pull/2089


app/models/concerns/illiadable.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+# Mixin for requests that should be sent to ILLiad for initial handling
+module Illiadable
+  def submit!
```
jcoyne, line 5: Using modules that override methods on the class they are mixed into, isn't the easiet to read.  Is there a reason this is needed?  I see that HoldRecall#submit! already calls SubmitIlliadRequestJob can we modify Scan#submit to call SubmitIlliadRequestJob directly?

even better might be to convert this module to a new class (e.g. `IlliadRequest`) and have a method on the "Illiadable" requests like `#illiad_request`.  I think there is value in separating the concept of "The user made a request" and "We made this ILL request to fulfull the users request"

jcoyne, line 5: Using modules that override methods on the class they are mixed into, isn't the easiet to read. Is there a reason this is needed? I see that HoldRecall#submit! already calls SubmitIlliadRequestJob can we modify Scan#submit to call SubmitIlliadRequestJob directly?

even better might be to convert this module to a new class (e.g. IlliadRequest) and have a method on the "Illiadable" requests like #illiad_request. I think there is value in separating the concept of "The user made a request" and "We made this ILL request to fulfull the users request"

jcoyne (changes requested):

app/models/concerns/illiadable.rb
thatbudakguy: I'm fine just not including this method in the mixin – I figured if you mixed in Illiadable it'd be because you want to submit the request via ILLiad, but in the case of scans we _also_ want to do something else, so it's fine to just be explicit instead.

thatbudakguy: Updated to remove the method and make `Scan#submit!` explicitly call the illiad job

jcoyne (approved):

PR #1915, opened 2023-10-13
Add real ldap groups to the settings

jcoyne:
We always use these groups, so this makes it easier to approximate production from a development/test machine


marlo-longley (approved):

marlo-longley:
I find this much more straightforward and we (you) did a similar think in library_hours that also helped.

PR #1918, opened 2023-10-13
Consolidate request tests

jcoyne:
These tests repeat the same test setup and just have different assertions about parts of the body. We can make the test faster and more cohesive by just expecting the whole of the body

### DIFF
--- a/spec/models/illiad_request_spec.rb
+++ b/spec/models/illiad_request_spec.rb
@@ -47,91 +47,29 @@ RSpec.describe IlliadRequest do
     subject.request!
   end
 
-  it 'POSTs to the illiad api url' do
-    expect(a_request(:post, 'https://illiad.stanford.edu/ILLiadWebPlatform/Transaction/')).to have_been_made
-  end
-
   describe 'request headers' do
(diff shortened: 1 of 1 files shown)


thatbudakguy:
I still don't understand what the issue with this is. Each assertion is a specific statement about the body and how a single field in it should look. If you change it to a single assertion, you have to visually compare the two large request bodies to guess at what went wrong, instead of immediately knowing from the test setup.

There is actual knowledge encoded in the assertions that is valuable to spell out, like the fact that for Scans, the barcode should be included twice in two different fields (but not for Hold/Recalls). I don't think this specificity has any actual cost. The tests still run fast. It was helpful to me to write them in this way while I developed the features, and it's still helpful.

jcoyne:
I can appreciate that these tests were helpful when developing this feature, as someone who needed to make changes to this code it was overwhelming to me that I had to wade through 24 tests that were all testing the same method/behavior.  I think there is a very real cost to the complexity for maintainers to dig through all of this. It took me about 20 minutes to realize that all these tests are just testing 1 behavior.

PR #1923, opened 2023-10-16
Test that submitting a scan returns a success page

### LINKED ISSUE #1922: Scans no longer succeeds for the user.
This is due to a regression in https://github.com/sul-dlss/sul-requests/issues/1889 where `Scan#submit!` no longer returns true unless `scan_destination.patron_barcode` returns a present value.   

This return value is checked in `RequestsController#create` and then it tells the user that `There was a problem creating your request.`.  

One underlying cause of this bug being introduced is there is not feature test for creating a Scan.

jcoyne:
Removed the view test since this is now covered by the request test. 
Fixes #1922


app/models/requests/scan.rb
```diff
@@ -21,13 +21,14 @@ def send_approval_status!
...
-    # This ensures that only scan rules with a destination get sent to the ILS.
-    # We no longer want to send SAL3 requests to the ILS as this is handled by the ILLiad integration.
-    # SAL1/2 requests still go to the ILS at this time.
-    send_to_ils_later! if scan_destination&.dig(:patron_barcode).present?
+    SubmitIlliadRequestJob.perform_later(id).tap do
+      # This ensures that only scan rules with a destination get sent to the ILS.
+      # We no longer want to send SAL3 requests to the ILS as this is handled by the ILLiad integration.
+      # SAL1/2 requests still go to the ILS at this time.
+      send_to_ils_later! if scan_destination&.dig(:patron_barcode).present?
+    end
```
thatbudakguy, line 31: was this bug introduced because something expects `submit!` to return whatever the output of `perform` methods is? i don't really understand the implications of this change.

cbeer, line 31: I wonder if this captures the intent better (make sure we enqueue all the background jobs we need to):

```suggestion
    result = SubmitIlliadRequestJob.perform_later(id)
    return result if !result || scan_destination&.dig(:patron_barcode).blank?

    # This ensures that only scan rules with a destination get sent to the ILS.
    # We no longer want to send SAL3 requests to the ILS as this is handled by the ILLiad integration.
    # SAL1/2 requests still go to the ILS at this time.
    send_to_ils_later!
```

thatbudakguy: this is a little clearer imo, but the comment should go with the new `return` line, right? also `return result if !result` is kind of...weird and might require some more explanation

spec/features/create_scan_spec.rb
```diff
@@ -34,11 +34,25 @@
       stub_current_user(create(:scan_eligible_user))
     end
 
-    it 'displays a copyright restrictions notice in a collapse' do
+    it 'displays a copyright restrictions notice in a collapse', js: true do
```
cbeer, line 37:
```suggestion
    it 'displays a copyright restrictions notice in a collapse', :js do
```

PR #1948, opened 2023-10-30
Select the OCLC-M identifier and strip its value prefix before sending it to ILLiad

### LINKED ISSUE #1933: Strip OCoLC prefix from OCLC numbers sent to ILLiad
per @saseestone, our requests sent to ILLiad include the OCLC#, but FOLIO often has it prefixed with `OCoLC-M` or similar before the actual numbers start. this causes problems because other apps need to consume the number without the prefix.

the best place to change it is probably in the accessor on the FOLIO instance:
https://github.com/sul-dlss/sul-requests/blob/e1276a5e579215e19e8a2f6df145b09a6cb4bf27/app/models/folio/instance.rb#L79-L81

check at least the indexer, because we almost certainly do this there, and there may be other prefix styles for OCLC numbers that we need to handle.

cbeer:
Fixes #1933

### DIFF
--- a/app/models/folio/instance.rb
+++ b/app/models/folio/instance.rb
@@ -77,7 +77,11 @@ module Folio
     end
 
     def oclcn
-      @oclcn.join('; ')
+      return if @oclcn.blank?
+
+      preferred_value = @oclcn.find { |value| value.start_with?('(OCoLC-M)') } || @oclcn.first
+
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

jcoyne:
@cbeer feel free to merge after fixing up rubocop errors.

PR #1988, opened 2024-01-22
Clean-up Folio migration cruft

### LINKED ISSUE #1982: Clean up temporary FOLIO migration code
Mostly from https://github.com/sul-dlss/sul-requests/pull/1674

E.g. some of the needed_date handling for the `Page` should be unnecessary.

mjgiarlo:
This change reverts changes made to support the ILS migration from Symphony to Folio

Fixes #1982

Includes:
* Remove `needed_date` concern from `Page` model
* Remove migration feature flag

### DIFF
--- a/app/controllers/mediated_pages_controller.rb
+++ b/app/controllers/mediated_pages_controller.rb
@@ -23,10 +23,7 @@ class MediatedPagesController < RequestsController
   end
 
   def validate_request_type
-    # TODO: Determine if the logic after the || should stick around
-    return if current_request.mediateable? || (Settings.features.migration && current_request.pageable? && params[:action] == 'update')
-
-    raise UnmediateableItemError
+    raise UnmediateableItemError unless current_request.mediateable?
   end
 
   class UnmediateableItemError < StandardError
--- a/app/controllers/mediated_pages_controller.rb
+++ b/app/controllers/mediated_pages_controller.rb
@@ -23,6 +23,7 @@ class MediatedPagesController < RequestsController
   end
 
   def validate_request_type
+    # TODO: Determine if the logic after the || should stick around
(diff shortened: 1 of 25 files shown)


app/controllers/mediated_pages_controller.rb
```diff
@@ -23,7 +23,8 @@ def update_params
   end
 
   def validate_request_type
-    return if current_request.mediateable? || (Settings.features.migration && current_request.pageable? && params[:action] == 'update')
+    # TODO: Determine if the logic after the || should stick around
```
mjgiarlo, line 26: Bringing attention to this question for reviewers.

app/controllers/mediated_pages_controller.rb
```diff
@@ -23,7 +23,7 @@ def update_params
   end
 
   def validate_request_type
-    return if current_request.mediateable? || (Settings.features.migration && current_request.pageable? && params[:action] == 'update')
+    return if current_request.mediateable? || (current_request.pageable? && params[:action] == 'update')
```
jcoyne, line 26: Can't we pull out this second clause?  Wasn't `Settings.features.migration` being nil causing the rest of this to not be needed?

jcoyne (changes requested):

app/controllers/mediated_pages_controller.rb
mjgiarlo: I had originally thought not, but tracing through the `blame` farther than I did at first, yes, I do believe the second clause was added to support the migration work. Change forthcoming. Thanks!

mjgiarlo: Done.

jcoyne (approved):

PR #1995, opened 2024-01-29
[DO NOT MERGE] Rename "Library ID" to "University ID"

mjgiarlo:
Connects to sul-dlss/SearchWorks#3773

This commit helps us move the access portfolio beyond using library barcode numbers, leaning instead on the university-assigned and -managed "university ID."


mjgiarlo:
@corylown Do you know if this is still valuable to keep around or was it superseded?

PR #1997, opened 2024-01-31
FOLIO items may not have barcodes; fall back on the item UUID if there isn't a barcode

cbeer:
Fixes SW-4281

Irina adds:
> We found 323 items without a barcode where an item is attached to a holding with multiple items (the request does go through if there is only one item without a barcode).
> Most of them fall into two categories:
> 1.Items that are part of the acquisitions workflow where an item record is created when the order is placed, but the barcode is added only when we receive the piece. 
> 2. Items created by EAL staff for “East Asia Big Sets” SAL location (e.g., https://searchworks.stanford.edu/view/11612016).
Both categories need to be requestable.

### DIFF
--- a/app/models/folio/item.rb
+++ b/app/models/folio/item.rb
@@ -60,7 +60,7 @@ module Folio
                    due_date: nil, id: nil, holdings_record_id: nil, suppressed_from_discovery: false)
       @id = id
       @holdings_record_id = holdings_record_id
-      @barcode = barcode
+      @barcode = barcode.presence || id
       @status = status
       @type = type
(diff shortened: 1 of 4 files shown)


saseestone:
Other examples from Irina (from the original list of 323):

- https://searchworks.stanford.edu/view/in00000040213
- https://searchworks.stanford.edu/view/11612032
- https://searchworks.stanford.edu/view/4648482

I've also uploaded the report to the SW JIRA issue: https://jirasul.stanford.edu/jira/browse/SW-4281

corylown (approved):

PR #2004, opened 2024-02-06
Preserve parameters during interstitial redirect

### LINKED ISSUE #1934: Fix request parameter rewriting in rails 7.1.x
It seems that in rails 7.1 our HTTP request parameters are being rewritten during the delegation process, where e.g.:
> https://requests.stanford.edu/requests/new?a=b&c=d

should become
> https://requests.stanford.edu/hold_recalls/new?a=b&c=d

but instead becomes (note ampersand)
> https://requests.stanford.edu/hold_recalls/new?a=b&amp;c=d

this causes the parameters to be parsed incorrectly by the receiving controller:
```
Started GET “/hold_recalls/create?request[barcodes][36105231817524]=1&amp;request[destination]=GREEN-LOAN&amp;request[item_id]=in00000001563&amp;request[needed_date]=2024-10-23&amp;request[origin]=SAL3&amp;request[origin_location]=STACKS”
Parameters: {“request”=>{“barcodes”=>{“36105231817524”=>“1"}}, “amp;request”=>{“destination”=>“GREEN-LOAN”, “item_id”=>“in00000001563”, “needed_date”=>“2024-10-23”, “origin”=>“SAL3”, “origin_location”=>“STACKS”}}
```

this behavior originates here:
https://github.com/sul-dlss/sul-requests/blob/e1276a5e579215e19e8a2f6df145b09a6cb4bf27/app/controllers/requests_controller.rb#L148-L153

taylor-steve:
Fixes #1934. 

This seems to occur in both Rails 7.0 and 7.1:
https://app.honeybadger.io/projects/49963/faults/101591418/01HDEW0GXPZ486GB2K9862T6WT
https://app.honeybadger.io/projects/49963/faults/101723263/01HE0QQMPMVWHHTNX50KW496GQ

As suggested in #1934 it appears this was happening when going through `bounce_request_through_sso`, when the interstitial show ERB was escaping the URL in the script tag. This led to the parameter rewriting behavior. All previous tests passed because none had parameters that would be escaped.

As for safety, the only uses of interstitial use internally created `redirect_to` paths. This is my first time dealing with this in rails, if there is a better way, I am happy to hear it.

### DIFF
--- a/app/views/interstitial/show.html.erb
+++ b/app/views/interstitial/show.html.erb
@@ -4,5 +4,5 @@
 <script type='text/javascript'>
   var div = document.getElementById('redirectNote');
   div.innerHTML = 'Processing...';
-  window.location.href = '<%= @redirect_to %>';
+  window.location.href = '<%= @redirect_to.html_safe %>';
 </script>
--- a/app/views/interstitial/show.html.erb
+++ b/app/views/interstitial/show.html.erb
@@ -4,6 +4,8 @@
 <script type='text/javascript'>
   var div = document.getElementById('redirectNote');
   div.innerHTML = 'Processing...';
+
+  <%# Decode ampersands only, see https://github.com/sul-dlss/sul-requests/issues/1934 %>
   const redirectUrl = '<%= @redirect_to %>'.replaceAll('&amp;', '&');
   window.location.href = redirectUrl;
 </script>
--- a/app/views/interstitial/show.html.erb
+++ b/app/views/interstitial/show.html.erb
@@ -4,5 +4,6 @@
 <script type='text/javascript'>
   var div = document.getElementById('redirectNote');
   div.innerHTML = 'Processing...';
-  window.location.href = '<%= @redirect_to.html_safe %>';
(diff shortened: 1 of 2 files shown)


corylown (changes requested):
@taylor-steve I believe the issue has been fixed upstream in Rails (I am still looking for what exactly got fixed between 7.1.1 and 7.1.3.2). I can duplicate the problem on requests-dev with this request https://requests-dev.stanford.edu/requests/new?item_id=in00000001563&origin=SAL3&origin_location=SAL3-STACKS with Rails 7.1.1 but 7.1.3.2 works fine. I think we should close this PR and open a new PR that updates Rails to latest version of 7.1 to fix https://github.com/sul-dlss/sul-requests/issues/1934

Amendment: the issue appears to be in Rack and not Rails.

corylown:
Here's a draft PR to update Rails to 7.1.3: https://github.com/sul-dlss/sul-requests/pull/2022

corylown (approved):

Issue #196, opened 2015-06-26
Item selector styling


jvine:
* [x] Overall container styling
```
.item-selector {  
  overflow: auto;
  padding-right: 0;
  margin-bottom: 1em;
  padding-right: 0;
  max-height: 300px;
}
```
* [x] Remove margin-right from .search.form-control 
* [x] Remove margin-right from .sort-headings
* [x] Add btn-default to the call number and status header buttons 
* [x] Change the width split to 75%/25%
* [x] Make the call number button active by default
* [x] Make the icon tiny and floated right
```
  font-size: 10px;
  float: right;
  margin-top: 4px;
  margin-right: 5px;
```

* [x] Override all the border radiuses to 0px
 - .item-selector .input-group .form-control 
 - .input-group-addon
 - .sort-headings .btn

* [x] .input-group-addon gets same colour as .btn-default (can you just add the class, maybe?)
```
  color: #3f3c30;
  background-color: #f2f1eb;
  border-color: #e3dfd5;
```

* [x] Form control border (note border-top override)
```
.item-selector .input-group .form-control {
  border-radius: 0px !important;
  border-color: #e3dfd5;
  border-top: 0; 
}
```

* [x] When an item is selected, can we give its row a background colour? [#faf8d2]

With search box:
![screen shot 2015-06-25 at 5 04 47 pm](https://cloud.githubusercontent.com/assets/6945691/8368499/0000e418-1b60-11e5-9634-a265ef2299ff.png)

Without search box:
![screen shot 2015-06-25 at 5 32 20 pm](https://cloud.githubusercontent.com/assets/6945691/8368512/277d8960-1b60-11e5-9316-ebee037a9729.png)

PR #2013, opened 2024-02-13
update shelving order to match searchworks

### LINKED ISSUE #2002: Order items appearing in volume picker same (or similar) to how they display in SearchWorks record view
When displaying the "volume picker" feature, which allows the user to indicate which volume or specific copy they want to request, we should be ordering the items as they appear in the SearchWorks record view.

@dbranchini has a good example record for when the sort order changes.

dnoneill:
Update sort order to sort desc if there is a date, sort asc if there is a volume number without date.

Assumptions I made that might not be right
1. There are no enumerations with 4 digits in a row that are not a year
2. We want the header sort to match what is on sw.

Problem records/not exact matches
1. https://searchworks.stanford.edu/view/3439715: SAL3 sorting is slightly off due to the call number `AP4 .N65 V.149:NO.5540-5550 2020:OCT-202` missing a date and only including it in the notes. We might be able to fix this but I wasn't sure if it was possible due to the fact that the note indicating this wasn't in the results I was getting from dev. It might also be somewhat dangerous if the note is in a weird format. Not all are about missing dates.

**Weirdness**
I used an index value to keep sort matching what is in searchworks. Unfortunately in order to allow sort on that number you also have to allow filtering. That means if you search for example 6 on this record: http://localhost:3000/pages/new?item_id=72765&origin=SAL3&origin_location=SAL3-STACKS you will get two results.

Tested pages
-----
http://localhost:3000/pages/new?item_id=6273253&origin=SAL&origin_location=SAL-ND-PAGE-EA (same on both)

http://localhost:3000/pages/new?item_id=9273087&origin=SAL3&origin_location=SAL3-STACKS (same on both)

http://localhost:3000/pages/new?item_id=5490833&origin=MUSIC&origin_location=MUS-STACKS
Prod
5490833 - current
This branch
5490833 - local

http://localhost:3000/pages/new?item_id=72770&origin=SAL3&origin_location=SAL3-STACKS (this one should be no change. Looks exactly the same on prod and this branch).

http://localhost:3000/pages/new?item_id=72765&origin=SAL3&origin_location=SAL3-STACKS (this one should be no change. Looks exactly the same on prod and this branch).
72765 - local

http://localhost:3000/pages/new?item_id=3439715&origin=SAL&origin_location=SAL-PAGE
Prod
3439715 - current
This branch
3439715 - locally

http://localhost:3000/pages/new?item_id=2403853&origin=SAL3&origin_location=SAL3-STACKS

Prod
2403853 - current

This branch
2403853 - local

http://localhost:3000/pages/new?item_id=372380&origin=SAL3&origin_location=SAL3-STACKS
prod
372380 - current

this branch
372380 - local

closes #2002

### DIFF
--- a/app/javascript/item_selector/item_selector_filtering.js
+++ b/app/javascript/item_selector/item_selector_filtering.js
@@ -4,7 +4,7 @@ import itemSelector from '../item_selector'
 var itemSelectorFiltering = (function() {
   var listOptions = {
     page: 10000,
-    valueNames: ['callnumber', 'status']
+    valueNames: ['callnumber', 'status', 'index']
   };
 
   return $.extend({}, itemSelector, {
@@ -25,7 +25,7 @@ var itemSelectorFiltering = (function() {
       // List.js dynamically injects behavior, so we need to check
       // if the list is sortable before setting the default sort
       if($.isFunction(list.sort)) {
-        list.sort('callnumber', { order: 'asc'});
+        list.sort('index', { order: 'asc'});
       }
     },
 
--- a/app/assets/stylesheets/modules/item-selector.scss
+++ b/app/assets/stylesheets/modules/item-selector.scss
@@ -86,3 +86,6 @@
   }
 }
 
+.index {
+  display: none;
+}
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,6 +56,49 @@ module ApplicationHelper
     params.except(:controller, :action).to_unsafe_h
   end
 
+  def sort_holdings(holdings_object)
+    holdings_object.sort_by { |item| digit_match(item) }
+  end
+
+  def enumeration_month(enumeration)
+    monthconversion = { 'jan' => 1, 'feb' => 2, 'mar' => 3, 'apr' => 4, 'may' => 5, 'june' => 6,
+                        'july' => 7, 'aug' => 8, 'sept' => 9, 'oct' => 10, 'nov' => 11, 'dec' => 12 }
+    hasmonth = enumeration.downcase.scan(/(#{monthconversion.keys.join('|')})/)
+
+    return 0 if hasmonth.empty?
+
+    monthconversion[hasmonth.last[0]]
+  end
+
+  def enumeration_year_volume(enumeration_numbers, defaultvalues, enumeration)
+    enumeration_numbers.each do |number|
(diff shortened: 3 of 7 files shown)


cbeer:
I think there's another edge case around volume sorting that a simple reverse sort doesn't fix. Take a look at this multivolume work:
https://searchworks.stanford.edu/view/72765

Searchworks has the volumes in ascending order.

dnoneill:
@cbeer Do you know why https://github.com/sul-dlss/sul-requests/blob/d701d6bed0f30debf744be5157ad00fd08e62a97/app/javascript/item_selector/item_selector_filtering.js#L28 this exists? It ends up reordering everything even if I get it right in rails. Can I disable it or is that going to cause problems somewhere else?

app/javascript/item_selector/item_selector_filtering.js
dnoneill, line 1: could use feedback on why we need this. I have a guess

hudajkhan: Relates to work done as part of https://github.com/sul-dlss/sul-requests/issues/196

hudajkhan: The question: Do we wish to maintain the feature to sort call numbers and status values lexically? Tagging @dbranchini for the question

dnoneill: I double checked. This does not get rid of that if you click on the call number or status heading.

dnoneill: Okay, figured out that the line specifically triggered the call number header to display as toggled. I have updated the sorting to work on call numbers to match what is in sw.

dbranchini:
This should not hold up the release of this work. We can always add another ticket to "fix" or update the order on the SearchWorks side.

hudajkhan:
I tried an example out locally and looked at the code. Thank you for the screenshots! they are fantastic.  My only comment is around the filtering/searching function which is also using the index value (since "list options" is used for both sorting and filtering).  It does see like there could be an option to use a custom search method to have the list not include the index value when filtering.  That said, we can create a new issue that will handle that instead of forcing it into this one, so I'll go ahead and approve and merge, and perhaps @dnoneill can make a new issue if there is a way to set up custom filtering code. Thanks!

hudajkhan (approved):

PR #2014, opened 2024-02-14
Rename objects that hold FOLIO data

marlo-longley:
This made it a lot easier for me to understand the code, because these objects aren't actually requests. I found the naming misleading.


marlo-longley:
@jcoyne done !

PR #2030, opened 2024-03-06
update requests to allow bound-withs

### LINKED ISSUE #1446: Title level: display explanatory info about Bound-with parent
So that a user isn't confused when they ultimately receive the parent item. The user should expect to receive the parent item with their requested child item bound inside.

See https://github.com/sul-dlss/SearchWorks/issues/3200 for more context

Could test this out by sending a query param, even without editing the button from SearchWorks.

This is the view:

Screenshot 2024-01-31 at 5 07 04 PM

dnoneill:
closes #1446 

**I was not able to test actually creating a request locally. I am pretty sure based on the code that call number isn't something get sent for single items but I could not confirm that**
What this PR does:
1. updates instance graphql query to query `holdingsRecord` for `boundWithItem` records for those holdings. This gets parent_bound_withs for the record. It also adds `boundWithHoldingsPerItem` to the item query that is used both by the `boundWithItem` and `instance`.
2. Updates folio instance model to get items not only from the response but also from the boundWithItem field. This can cause an increase in items (see http://localhost:3000/pages/new?item_id=387559&origin=SAL3&origin_location=SAL3-STACKS). This example adds the [A 1.34:87 bound with A 1.34:81. 36105019614499 (item id)](https://searchworks.stanford.edu/view/10564224) 
3. This ticket also adds a list of bound-withs for an item if they have them.
4. Moves call number

Note: after discussion with Darcy the call number/title display is in one format for maintenance considerations.

1. http://localhost:3000/hold_recalls/new?item_id=6290245&origin=GREEN&origin_location=GRE-STACKS
Screenshot 2024-03-06 at 11 39 24 AM

4. http://localhost:3000/hold_recalls/new?barcode=36105029307795&item_id=6290236&origin=GREEN&origin_location=GRE-STACKS
Screenshot 2024-03-06 at 11 38 51 AM

5. http://localhost:3000/pages/new?item_id=2969234&origin=SAL3&origin_location=SAL3-STACKS
Screenshot 2024-03-06 at 11 18 24 AM

6. http://localhost:3000/pages/new?item_id=10564224&origin=SAL3&origin_location=SAL3-STACKS

Screenshot 2024-03-06 at 12 25 54 PM

### DIFF
--- a/app/assets/stylesheets/modules/forms.scss
+++ b/app/assets/stylesheets/modules/forms.scss
@@ -65,6 +65,11 @@ hr {
   white-space: nowrap;
 }
 
+.single-item-callnumber {
+  margin: -1em 0 0.5em;
+  color: $color-cardinal-red;
+}
+
 .go-back-link {
   line-height: 3;
 }
@@ -73,11 +78,6 @@ hr {
   color: $dark-gray;
   margin-bottom: 0;
  }
-
-.form-control-static {
-  min-height: 0;
-  padding-bottom: 0;
-}
 
 .login-send-group {
   margin-top: 25px;
--- a/app/assets/stylesheets/modules/item-selector.scss
+++ b/app/assets/stylesheets/modules/item-selector.scss
@@ -89,3 +89,28 @@
 .index {
   display: none;
 }
+
+.bound-withs {
+  max-height: 14em;
+  background-color: $fog-light;
+  overflow: scroll;
+  padding: 8px;
+  margin-bottom: 15px;
+  border-radius: 4px;
+
+  ul {
+    margin: 0px;
+  }
+
+  .bound-with-item {
+    padding: 4px 0px;
+  }
+
+  .heading {
+    padding: 8px 0px;
+  }
+
+  li:not(:only-child) {
+    border-top: 1px solid $grayish-yellow;
+  }
+}
--- a/app/models/folio/instance.rb
+++ b/app/models/folio/instance.rb
@@ -14,6 +14,46 @@ module Folio
 
       Folio::Instance.from_dynamic(data)
     end
+
+    def self.filter_parent_bound_withs(parent_bound_withs, hrid)
+      return [] unless parent_bound_withs
+
+      parent_bound_withs.flatten.compact.each do |bound_with_record|
+        @filter_parent_bound_withs = bound_with_record.dig('instance', 'items').each.map do |item|
+          boundwithhrids = item['boundWithHoldingsPerItem'].filter { |bw| bw.dig('instance', 'hrid') == hrid }.flatten
+          update_item_fields(item, boundwithhrids.first) if boundwithhrids.present?
+        end
+      end
+      @filter_parent_bound_withs.compact
+    end
+
+    def self.update_item_fields(item, holdings_record)
+      item['effectiveCallNumberComponents']['callNumber'] = holdings_record['callNumber']
+      item['volume'] = ''
+      item['enumeration'] = ''
+      item
+    end
+
+    def self.items(json)
+      hrid = json.fetch('hrid', '')
+      parent_bound_withs = parent_bound_withs(json)
+      return json.fetch('items', []).map { |item| Folio::Item.from_hash(item) } unless parent_bound_withs.present? && hrid
+
+      parent_bound_withs = parent_bound_withs.pluck('boundWithItem')
+      items = filter_parent_bound_withs(parent_bound_withs, hrid)
+      items += json['items'] if json['items']
+      items.compact.map { |item| Folio::Item.from_hash(item) }
+    end
+
+    def self.parent_bound_withs(json)
+      json.fetch('holdingsRecords', []).filter { |holding| holding['boundWithItem'] }
+    end
+
+    def self.child_bound_withs(json)
+      return unless json['items']
+
+      json['items'].pluck('boundWithHoldingsPerItem').flatten.compact
+    end
     # rubocop:enable Style/OptionalBooleanParameter
 
     # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
@@ -35,14 +75,17 @@ module Folio
         end,
         electronic_access: json.fetch('electronicAccess', []),
         edition: json.fetch('editions', []),
-        items: json.fetch('items', []).map { |item| Folio::Item.from_hash(item) }
+        items: items(json),
+        parent_bound_withs: parent_bound_withs(json),
+        child_bound_withs: child_bound_withs(json)
       )
     end
+
     # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
     # rubocop:disable Metrics/MethodLength, Metrics/ParameterLists
     def initialize(id:, hrid: '', title: '', contributors: [], pub_date: nil, pub_place: nil, publisher: nil, format: nil,
-                   isbn: [], oclcn: [], electronic_access: [], edition: [], items: [])
+                   isbn: [], oclcn: [], electronic_access: [], edition: [], items: [], parent_bound_withs: [], child_bound_withs: [])
       @id = id
       @hrid = hrid
       @title = title
@@ -56,6 +99,8 @@ module Folio
       @electronic_access = electronic_access
       @edition = edition
       @items = items
+      @parent_bound_withs = parent_bound_withs
+      @child_bound_withs = child_bound_withs
     end
     # rubocop:enable Metrics/MethodLength, Metrics/ParameterLists
 
@@ -70,7 +115,7 @@ module Folio
       contributor&.fetch('name')
     end
 
-    attr_reader :pub_date, :pub_place, :publisher, :format
+    attr_reader :pub_date, :pub_place, :publisher, :format, :parent_bound_withs, :child_bound_withs
 
     def isbn
       @isbn.first
--- a/app/views/application/_item_selector.html.erb
+++ b/app/views/application/_item_selector.html.erb
@@ -2,20 +2,23 @@
   <%= f.fields_for :barcodes do |barcode| %>
     <% if f.object.barcode_present? || f.object.all_holdings.one? %>
       <% single_item = f.object.requested_holdings.first || f.object.all_holdings.first %>
-      <div class='form-group'>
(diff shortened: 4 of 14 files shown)


app/assets/stylesheets/modules/forms.scss
```diff
@@ -65,6 +65,11 @@ hr {
   white-space: nowrap;
 }
 
+.single-item-callnumber {
+  margin: -1em 0 0.5em;
+  color: #8c1515;
```
jcoyne, line 70: Please make this a variable and put it in https://github.com/sul-dlss/sul-requests/blob/main/app/assets/stylesheets/sul-variables.scss#L1

app/assets/stylesheets/modules/item-selector.scss
```diff
@@ -86,3 +86,34 @@
   }
 }
 
+.bound-withs {
+  max-height: 200px;
+  background-color: #F4F4F4;
```
jcoyne, line 91: Please make this a variable and put it in https://github.com/sul-dlss/sul-requests/blob/main/app/assets/stylesheets/sul-variables.scss#L1

app/assets/stylesheets/modules/item-selector.scss
```diff
@@ -86,3 +86,34 @@
...
+  overflow: scroll;
+  padding: 8px;
+  margin-bottom: 15px;
+  border-radius: 4px;
+
+  .bound-with-item span{
```
jcoyne, line 97: Can we just use a `div` instead?

app/assets/stylesheets/modules/item-selector.scss
```diff
@@ -86,3 +86,34 @@
...
+    padding: 4px 0px;
+  }
+
+  .heading {
+    font-weight: 500;
+    margin-bottom: 10px;
+  }
+
+  li:not(:only-child) {
+    border-top: 1px solid #D5D5D4;
```
jcoyne, line 111: Please make this color a variable and put it in https://github.com/sul-dlss/sul-requests/blob/main/app/assets/stylesheets/sul-variables.scss#L1

app/models/folio/instance.rb
```diff
@@ -14,6 +14,44 @@ def self.fetch(request, _live_lookup = true)
...
+    end
+
+    def self.update_item_fields(item, holdings_record)
+      item['effectiveCallNumberComponents']['callNumber'] = holdings_record['callNumber']
+      item['volume'] = ''
+      item['enumeration'] = ''
+      item
+    end
+
+    def self.get_items(json)
```
jcoyne, line 37: Using `get_` is sort of frowned on in ruby. Can we use `items(json)` instead?  The "getting" is implied.

app/models/folio/instance.rb
```diff
@@ -14,6 +14,44 @@ def self.fetch(request, _live_lookup = true)
...
+      return json.fetch('items', []).map { |item| Folio::Item.from_hash(item) } unless parent_bound_withs.present? && hrid
+
+      parent_bound_withs = parent_bound_withs.pluck('boundWithItem')
+      items = filter_parent_bound_withs(parent_bound_withs, hrid)
+      items += json['items'] if json['items']
+      items.compact.map { |item| Folio::Item.from_hash(item) }
+    end
+
+    def self.parent_bound_withs(json)
+      json.fetch('holdingsRecords', [])&.filter { |holding| holding['boundWithItem'] }
```
jcoyne, line 49: Is the safe navigation necessary here?

app/models/folio/instance.rb
```diff
@@ -14,6 +14,44 @@ def self.fetch(request, _live_lookup = true)
...
+    end
+
+    def self.child_bound_withs(json)
+      json['items']&.pluck('boundWithHoldingsPerItem')&.flatten&.compact
```
jcoyne, line 53: If you do `return unless json['items']`, then you don't need all of these safe navigation operators.

app/views/application/_item_selector.html.erb
```diff
@@ -2,20 +2,26 @@
...
+      <div class="single-item-callnumber">
+        Call number: <%= single_item.callnumber %>
       </div>
       <%= barcode.hidden_field single_item.barcode.present? ? single_item.barcode : 'NO_BARCODE', value: 1 %>
+      <% if f.object.bound_with? %>
+        <div class="bound-withs">
+            <% is_child = f.object.parent_bound_withs.present? %>
+            <div class="heading">This item is bound<% if is_child %> and shelved<% end %> with</div>
+            <% if is_child %>
+              <%= render partial: 'item_selector_bound_withs', locals: { bound_withs: f.object.parent_bound_withs, is_child: } %>
```
jcoyne, line 14:
```suggestion
              <%= render 'item_selector_bound_withs', bound_withs: f.object.parent_bound_withs, is_child: %>
```

app/views/application/_item_selector.html.erb
```diff
@@ -2,20 +2,26 @@
...
+            <% else %>
+              <%= render partial: 'item_selector_bound_withs', locals: { bound_withs: f.object.child_bound_withs, is_child: } %>
```
jcoyne, line 16:
```suggestion
              <%= render 'item_selector_bound_withs',  bound_withs: f.object.child_bound_withs, is_child: %>
```

app/views/application/_item_selector.html.erb
```diff
@@ -2,20 +2,26 @@
...
+            <% end%>
```
jcoyne, line 17:
```suggestion
            <% end %>
```

app/views/application/_item_selector.html.erb
```diff
@@ -2,20 +2,26 @@
...
+        Call number: <%= single_item.callnumber %>
       </div>
       <%= barcode.hidden_field single_item.barcode.present? ? single_item.barcode : 'NO_BARCODE', value: 1 %>
+      <% if f.object.bound_with? %>
+        <div class="bound-withs">
+            <% is_child = f.object.parent_bound_withs.present? %>
+            <div class="heading">This item is bound<% if is_child %> and shelved<% end %> with</div>
+            <% if is_child %>
+              <%= render partial: 'item_selector_bound_withs', locals: { bound_withs: f.object.parent_bound_withs, is_child: } %>
+            <% else %>
```
jcoyne, line 15: This branch looks identical to the one above.

app/assets/stylesheets/modules/item-selector.scss
```diff
@@ -86,3 +86,34 @@
   }
 }
 
+.bound-withs {
+  max-height: 200px;
```
jcoyne, line 90: Can we use relative units (rem or em) where appropriate?  This makes it easier to style for responsiveness to different screen sizes.

app/views/application/_item_selector_bound_withs.html.erb
```diff
@@ -0,0 +1,10 @@
+<ul>
```
jcoyne, line 1:
```suggestion
<ul  class="list-unstyled">
```

app/assets/stylesheets/modules/item-selector.scss
```diff
@@ -86,3 +86,34 @@
...
+  .heading {
+    font-weight: 500;
+    margin-bottom: 10px;
+  }
+
+  li:not(:only-child) {
+    border-top: 1px solid #D5D5D4;
+  }
+
+  ul {
```
jcoyne, line 114: Can we use the Bootstrap  class="list-unstyled" instead?

jcoyne (changes requested):

app/models/folio/instance.rb
dnoneill: I was thinking holdingsRecord can return nil but I am not sure that exists so no.

app/views/application/_item_selector.html.erb
dnoneill: I am not sure I understand what you are saying here.

dnoneill: I am not sure I understand what you are saying.

jcoyne (approved):

cbeer:
This is great. I've tested it out locally with some bound-with children + bound-with parents. It's great we're able to preserve the bound-with child information through mylibrary by associating the request with the bound-with child. I think we should verify Caiasoft is ok with that (Darsi is following up with them), but I don't think that testing needs to hold up this PR.

app/models/folio/instance.rb
```diff
@@ -14,6 +14,46 @@ def self.fetch(request, _live_lookup = true)
...
+    def self.filter_parent_bound_withs(parent_bound_withs, hrid)
+      return [] unless parent_bound_withs
+
+      parent_bound_withs.flatten.compact.each do |bound_with_record|
+        @filter_parent_bound_withs ||= bound_with_record.dig('instance', 'items').each.map do |item|
+          boundwithhrids = item['boundWithHoldingsPerItem'].filter { |bw| bw.dig('instance', 'hrid') == hrid }.flatten
+          update_item_fields(item, boundwithhrids.first) if boundwithhrids.present?
+        end
+      end
+      @filter_parent_bound_withs.compact
```
cbeer, line 27:
```suggestion
      parent_bound_withs.flatten.compact.each do |bound_with_record|
        filter_parent_bound_withs ||= bound_with_record.dig('instance', 'items').each.map do |item|
          boundwithhrids = item['boundWithHoldingsPerItem'].filter { |bw| bw.dig('instance', 'hrid') == hrid }.flatten
          update_item_fields(item, boundwithhrids.first) if boundwithhrids.present?
        end
      end
      filter_parent_bound_withs.compact
```

I don't think we want to memoize these at the class level here.

dnoneill: @cbeer It breaks the application if we remove the memoization.

dnoneill:
There is one thing that isn't totally consistent across pages. If it is a new page http://localhost:3000/pages/new?item_id=14029321&origin=SAL3&origin_location=SAL3-STACKS the call number will have more spacing because of the line https://github.com/sul-dlss/sul-requests/blob/7cb6f9b6ab04a2da1f78705474ec1d34d2357fc9/app/assets/stylesheets/modules/forms.scss#L15. I don't think it is a change that will stop this from being merged but I wanted to make sure it was documented.

PR #2052, opened 2024-03-13
Use Node v20 LTS in CI

taylor-steve:


jcoyne (approved):

.github/workflows/ruby.yml
```diff
@@ -24,10 +24,10 @@ jobs:
         key: ${{ runner.os }}-gems-202103-${{ hashFiles('**/Gemfile.lock') }}
         restore-keys: |
           ${{ runner.os }}-gems-202103
-    - name: Use Node.js LTS v18
+    - name: Use Node.js LTS v20
       uses: actions/setup-node@v4
       with:
-        node-version: 'lts/hydrogen'
+        node-version: 'lts/iron'
```
jcoyne, line 30: Can we do `node-version: '20'` here?  Do we need to know the node version codename?

jcoyne: I do.

PR #2063, opened 2024-03-18
Create PatronRequestsController with a totally separate layout, stylesheet, javascript and model

### LINKED ISSUE #2065: Create a seam that isolates the new styles + javascript from the existing application
We want an initial scaffolding with:
- Bootstrap 5
- up-to-date Stanford branding (colors, fonts, etc)
- turbo

This might also involve creating a new controller + layout.

cbeer:
Fixes #2065

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -21,6 +21,10 @@ class RequestsController < ApplicationController
 
   rescue_from bib_model_class::NotFound, with: :item_not_found
 
+  before_action only: :new, if: -> { Settings.features.requests_redesign } do
+    redirect_to(new_patron_request_path(current_request))
+  end
+
   def new
     current_request.assign_attributes(new_params)
     validate_request_type
@@ -146,6 +150,7 @@ class RequestsController < ApplicationController
 
   def delegated_new_request_path(request, url_params = nil)
     url_params ||= params.except(:controller, :action).to_unsafe_h
+
     request.delegate_request!
     new_polymorphic_path(request.type.underscore, url_params)
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_08_07_203550) do
+ActiveRecord::Schema[7.1].define(version: 2024_03_18_225555) do
   create_table "admin_comments", force: :cascade do |t|
     t.string "commenter"
     t.string "comment"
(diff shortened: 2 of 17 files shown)


db/schema.rb
dnoneill, line 1: I am assuming that the inclusion of hrid, folio_instance_id and origin_location_code are to replace title? And from there we would live lookup items in the view? Can you expand on why the additional fields are being added?

cbeer: I just created fields for (what seemed like) the form fields in the designs (or fields we need to place a request in FOLIO), and added:

- `data`, to match our current catch-all serialized field. I don't know what we'd put in it, but maybe it's where we stash some FOLIO data
- `status` and `origin_location_code` to support mediated paging (it seems like we make extensive use of them now, so presumably we'll want something like them?)
- `folio_request_id` just to make a connection to the FOLIO request we're creating.. mostly for debugging, I think (but maybe useful if we let someone cancel a request in the app?)

cbeer: Just in the spirit of the rest of this seam. There's a lot of logic and weird stuff in the existing `Request` model that we probably don't need. By starting fresh now, hopefully at the end of the workcycle we'll have a clear idea of what we need going forward and can work to deprecate the old model (either by migrating data, merging the models, or something else). 

Also, if we tried to use `Request`, we'll have to keep anything we do working with e.g. the existing mediated paging forms. I'm sure it could be done, but hopefully this gives us less to review and fewer backwards compatibility concerns when we're starting development.

thatbudakguy: I'm cool with a new table also to move away from the STI structure that characterizes current requests; makes it easier to select all the requests created since/before the transition

thatbudakguy (approved):

app/controllers/requests_controller.rb
```diff
@@ -146,6 +146,8 @@ def check_if_proxy_sponsor
 
   def delegated_new_request_path(request, url_params = nil)
     url_params ||= params.except(:controller, :action).to_unsafe_h
+    return new_patron_requests_path(url_params) if Settings.features.requests_redesign
```
dnoneill, line 149: new_patron_request_path (the extra s causes a problem)

dnoneill: Also line 154 needs to be changed to 
redirect_to delegated_new_request_path(current_request) unless (delegated_request? || current_request.scannable?) && !Settings.features.requests_redesign

dnoneill (changes requested):

app/controllers/requests_controller.rb
cbeer: Thanks, good catch. I've pulled that logic all the way up to a `before_action` so we don't have to worry about any interaction with the existing code 🤞

PR #604, opened 2016-10-26
Improved item lookup for mediation

jkeck:
Closes #547 
Closes #602

PR #2077, opened 2024-03-20
Refactor `Folio::Instance.fetch` to decouple it from `Request` (and remove some legacy parameters)

cbeer:


dnoneill:
@cbeer I am inclined to merge, but I can't quite figure out what the live_lookup variable was doing? Is that legacy code or am I missing something?

cbeer:
Legacy; now that we're talking to FOLIO directly, we're effectively always getting "live" data.

It originally came in with https://github.com/sul-dlss/sul-requests/pull/604, and we kept it going to keep our method signatures the same.. but now the old stuff is all gone, so 🤷‍♂️

dnoneill (approved):

PR #2078, opened 2024-03-20
Implement SUL header

### LINKED ISSUE #2066: Add Stanford libraries branding header for new layout
Image

thatbudakguy:
Closes #2066

I didn't add the 'SearchWorks' text to the app header (it just says "Requests") since I don't think we were 100% on that yet?

The font size for the header links is also quite small in current production and in the designs (smaller than BS5's minimum font size). Should we make it that small? I left it as the default.

Most screens:

Screenshot 2024-03-20 at 14 14 55

Small screens:

Screenshot 2024-03-20 at 14 15 09

### DIFF
--- a/app/views/shared/_sul_header.html.erb
+++ b/app/views/shared/_sul_header.html.erb
@@ -1,0 +1,34 @@
+<!-- SUL header -->
+<div id="sul-header" class="bg-white text-cardinal py-1 fs-7">
+  <div class="container d-flex flex-column flex-sm-row gap-1 gap-sm-0 justify-content-between align-items-center">
+    <%= link_to 'https://library.stanford.edu' do %>
+      <%= tag.span t('.sul'), class: "visually-hidden" %>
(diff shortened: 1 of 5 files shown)


app/views/shared/_sul_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+            <%= link_to t('.logout', sunetid: current_user.sunetid), logout_path(referrer: root_path) %>
+          <% else %>
+            <%= link_to t('.login'), login_path(referrer: request.original_url) %>
+          <% end %>
+        </li>
+        <li>
+          <%= link_to t('.my_account'), "https://mylibrary.stanford.edu/" %>
+        </li>
+        <li>
+          <%= link_to t('.feedback'), "https://searchworks.stanford.edu/feedback", target: '_blank', rel: 'noopener noreferrer'  %>
```
cbeer, line 22:
```suggestion
          <%= link_to t('.feedback'), "https://searchworks.stanford.edu/feedback", target: '_blank' %>
```

Apparently _blank implies these now.

thatbudakguy:
It is, yeah. I agree!

PR #2089, opened 2024-03-21
Use Warden to control authentication instead of our home-grown implementation

### LINKED ISSUE #2070: Implement SSO/Library ID auth (e.g. MyLibrary's SessionsController)

cbeer:
This refactor inserts Warden into the authentication stack (similar to MyLibrary) now that we're planning to authenticate both through Shibboleth and via Library ID/pin. `CurrentUser` is now just a shim around the underlying `Warden::Proxy`.

In the future, `CurrentUser` and/or `User` might get refactored away, but I've left them both as-is to avoid refactoring the current requests flow.

TODO:
- [x] clean up commit history
- [x] use warden to clean up `CurrentUser` model
- [x] consider whether we really need HTML flash messages
- [x] do we need to use warden scopes to distinguish between request users and admin users?
- [x] test  this on -dev

Unresolved questions:

- [ ] is the login link in the header still only for SSO? What happens if a libid or name/email user clicks login in the middle of the requests flow?
- [ ] where do we want to send auth'ed users? does `patron_requests#new` handle both cases, or do we need a "step 2" action?
- [ ] can/should we use turbo for our forms (SSO has to be its own non-turbo thing)?

Fixes #2070


app/services/folio_client.rb
```diff
@@ -44,6 +44,8 @@ def inspect
   def login_by_sunetid(sunetid)
     response = get_json('/users', params: { query: CqlQuery.new(username: sunetid).to_query })
     response&.dig('users', 0)
+  rescue RuntimeError
+    nil
```
thatbudakguy, line 48: what raises this?

app/services/folio_client.rb
```diff
@@ -53,6 +55,34 @@ def login_by_library_id(library_id)
...
 
+  # Return the FOLIO user object given a library id (e.g. barcode)
+  # See https://s3.amazonaws.com/foliodocs/api/mod-users/p/users.html#users__userid__get
+  def login_by_library_id_and_pin(library_id, pin)
+    response = get_json('/users', params: { query: CqlQuery.new(barcode: library_id).to_query })
+    user = response&.dig('users', 0)
+
+    return unless user && validate_patron_pin(user['id'], pin)
+
+    user
```
thatbudakguy, line 66: is it not possible to `RuntimeError` here too?

thatbudakguy (approved):

app/services/folio_client.rb
cbeer: Good catch; I think maybe we can remove this now (or we'll discover a need later).

PR #2090, opened 2024-03-21
port over mylibrary feedback form to requests

### LINKED ISSUE #2069: Pull over a requests-specific feedback form
See e.g. https://github.com/sul-dlss/mylibrary/issues/571 for initial design guidance. I'm sure there will be tweaks + updates, but we should get the functional pieces in place within requests instead of linking out to searchworks.

dnoneill:
closes #2069 

Difference from mylibrary
- show_feedback_form? Replaced with current_page?(feedback_form_path)
- Doesn’t show feedback link in menu if it is on the feedback form page
- Updates the validate method to not mark an email as spam if there email has an url in it
- Uses the feedback_form/new template for all feedback form views instead of a shared/feedback_forms/form 
- Does not separate the reporting from html into a partial
- Uses root_url for mailer 
- Checks for sunset id for current_user?

Feedback page
Screenshot 2024-03-21 at 5 22 29 PM

Feedback page at home
Screenshot 2024-03-21 at 5 22 43 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -120,3 +120,5 @@ gem "jsbundling-rails", "~> 1.1"
 gem "turbo-rails", "~> 1.4"
 
 gem "propshaft", "~> 0.7.0"
+
+gem 'recaptcha'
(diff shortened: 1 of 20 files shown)


hudajkhan (approved):
The feedback form looks great! Thanks

PR #2091, opened 2024-03-22
Add handling for when the origin location is SAL3-PAGE-AS

### LINKED ISSUE #2059: Aeon interstitial page constructs bad url
Items in FOLIO location = SAL3-PAGE-AS have a location details key `pageAeonSite` with a value of `ARS`

Using that loc details, we send the request off to Aeon, but through sul-requests so the user see's the Aeon interstitial page: https://requests.stanford.edu/aeon_pages/new?item_id=11060578&origin=SAL3&origin_location=SAL3-PAGE-AS

That interstitial page has a link at the bottom for more information. 
image

For items in this location, that link loops the user back to the interstitial page. :(

When g-core looked at this in Nov 2023, Chris thought that link was being constructed based on the library of the permanent loc. That's why it tries to send you to the "SAL3 (off-campus storage) Reading Room service page".  

IF we keep this Aeon interstital page, we should use the loc details value of `ARS` to know that ARS manages this location, and construct the url (and link to the information page) based on that value. Not the perm library of the location. 

JIRA issue of original report: https://jirasul.stanford.edu/jira/browse/SW-4259

hudajkhan:
Background:
As described in the issue, when the origin location is SAL3-PAGE-AS, the interstitial page for Aeon uses the URL for the same request as the link for the reading room.

Desired behavior:
In this situation, the page should display the information for 'ARS' which is included under 'pageAeonSite' in the details for the location in Folio. 

What this pull request does:

For both the page heading and the more info description and link, this code uses a method that checks if the location details has "ARS" as the value for "pageAeonSite".  If that condition is true, we return "ARS" as the origin library code.  Otherwise we return the actual origin library code. 

Closes #2059

### DIFF
--- a/app/helpers/requests_helper.rb
+++ b/app/helpers/requests_helper.rb
@@ -99,8 +99,8 @@ module RequestsHelper
   # For the reading room information, we need to check if 'ARS' is in the location details
   # for the library. An example is SAL3, which should show the ARS reading room information
   # and so should return ARS as the library code for the reading room text block.
+  # This logic will be extended in the future to cover any location that has a pageAeonSite value.
   def aeon_reading_room_code
-
     details = Folio::Types.locations.find_by(code: current_request.origin_location).details
     details.key?('pageAeonSite') && details['pageAeonSite'] == 'ARS' ? 'ARS' : current_request.origin_library_code
   end
--- a/app/helpers/requests_helper.rb
+++ b/app/helpers/requests_helper.rb
@@ -95,4 +95,13 @@ module RequestsHelper
       user.library_id
     end
   end
+
+  # For the reading room information, we need to check if 'ARS' is in the location details
+  # for the library. An example is SAL3, which should show the ARS reading room information
+  # and so should return ARS as the library code for the reading room text block.
+  def aeon_reading_room_code
+
+    details = Folio::Types.locations.find_by(code: current_request.origin_location).details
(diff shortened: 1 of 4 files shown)


app/helpers/requests_helper.rb
```diff
@@ -95,4 +95,13 @@ def requester_info(user)
...
   end
+
+  # For the reading room information, we need to check if 'ARS' is in the location details
+  # for the library. An example is SAL3, which should show the ARS reading room information
+  # and so should return ARS as the library code for the reading room text block.
+  # This logic will be extended in the future to cover any location that has a pageAeonSite value.
+  def aeon_reading_room_code
+    details = Folio::Types.locations.find_by(code: current_request.origin_location).details
+    details.key?('pageAeonSite') && details['pageAeonSite'] == 'ARS' ? 'ARS' : current_request.origin_library_code
+  end
```
dnoneill, line 106: Is there a reason we can't move this functionality into https://github.com/sul-dlss/sul-requests/blob/8b883d5d15b9698a553cf65ed3d5bf784c6b2899/app/models/request.rb#L178

hudajkhan: For the origin library code function that is more generally applicable, I am not sure this mapping (i.e. from SAL3-PAGE-AS to ARS) applies to any other part of the application except for the reading room page.  In other words, for other situations where you would want an origin library code, this mapping doesn't apply.  I suppose we could do an aeon reading room code within the request class instead, but it may not matter for the moment because we'll be rewriting that anyway.

dnoneill (approved):

PR #2098, opened 2024-03-26
2097 recalls based on user group

### LINKED ISSUE #2097: Use patron-specific circulation rules (or FOLIO APIs?) to figure out whether to place a Hold or a Recall
E.g. https://requests.stanford.edu/hold_recalls/379379/status should have been placed as a Hold, not a Recall, because the patron's group doesn't allow Recalls.

It looks like we initially relied on FOLIO's `mod-patron` to assign the proper request type, but when we switched to using `mod-circulation` (for reasons... proxy users, perhaps?) we continued to use the generic circulation rule, not a patron-specific rule.

Here's where we're sending the request type:
https://github.com/sul-dlss/sul-requests/blob/main/app/jobs/submit_folio_request_job.rb#L92C23-L92C45

And the implementation of best_request_type: 
https://github.com/sul-dlss/sul-requests/blob/main/app/models/folio/item.rb#L179

We'll need a patron-aware version that evaluates the request type for the specific patron (and presumably opening https://github.com/sul-dlss/sul-requests/blob/main/app/services/folio/circulation_rules/policy_service.rb#L6 up to accept a specific patron group). 

If, when evaluating the specific patron group permissions, they are ineligible to hold or recall an item (but one of the "standard" patron groups can do one or the other), I think we should fall back on a regular `Hold` request until we can provide more specific messaging to the user.

dnoneill:
closes #2097 

I think a lot of this code is going to be able to be reused for our rewrite.

What it does:
Allows the patron model to check the rules for specific locations. The item model will then check if the logged in patron has recall permissions.

I used the Fee Borrower user and http://localhost:3000/hold_recalls/new?barcode=36105232863154&item_id=14720993&origin=GREEN&origin_location=GRE-STACKS for testing.

### DIFF
--- a/app/jobs/submit_folio_request_job.rb
+++ b/app/jobs/submit_folio_request_job.rb
@@ -89,7 +89,7 @@ class SubmitFolioRequestJob < ApplicationJob
 
       request_data = FolioClient::CirculationRequestData.new(
         request_level: 'Item',
-        request_type: item.best_request_type,
+        request_type: item.best_request_type(patron),
         instance_id: request.bib_data.instance_id,
         item_id: item.id,
         holdings_record_id: item.holdings_record_id,
--- a/app/models/folio/item.rb
+++ b/app/models/folio/item.rb
@@ -95,6 +95,14 @@ module Folio
       end
     end
 
+    def current_location_id
+      if effective_location && permanent_location.code == effective_location.code
+        effective_location.id
+      else
+        permanent_location.id
+      end
+    end
+
     def suppressed_from_discovery?
       @suppressed_from_discovery
     end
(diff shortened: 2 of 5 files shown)


app/jobs/submit_folio_request_job.rb
```diff
@@ -89,7 +89,7 @@ def create_item_circulation_request(barcode)
 
       request_data = FolioClient::CirculationRequestData.new(
         request_level: 'Item',
-        request_type: item.best_request_type,
+        request_type: item.best_request_type(patron),
```
cbeer, line 92: This might end up a little gnarly, but I wonder if this whole refactor would be clearer if we created `Patron#best_request_type(item)`. That should let us calculate the circ rules for the patron, instead of having to do it on the item

As a result, `Folio::CirculationRules::PolicyService.rules` should just be all the rules, and the group filtering can just stay with the instance? I think we were trying to keep the applicable rules slim for Searchworks because it could evaluate some thousands of items on a search results page... but requests might be better off using the whole list.

PR #2108, opened 2024-03-28
Remove the login link in the sul header

thatbudakguy:


dnoneill:
@thatbudakguy It might make more sense to have the login link be a logout link and just change it to that? Maybe that should be a different PR? I am not sure.

thatbudakguy:
when you're logged in, it is a logout link, right? when you aren't you won't see it

dnoneill:
@thatbudakguy Yup. Ignore me.

dnoneill (approved):

PR #2110, opened 2024-03-29
Wire up the page form

### LINKED ISSUE #2099: Build out form for single-item paging
Image

(I believe "Not needed after" is only for hold/recall requests, not ordinary paging. For this first pass, I think we can ignore this field.)

The "Earliest Available" date is calculated the same way as today -- depending on the origin location and the selected destination. Although we have existing javascript for that, maybe turbo/stimulus would get us somewhere easier?

cbeer:
Fixes #2099


hudajkhan (approved):
Tried this out locally and it works for page setup

PR #2111, opened 2024-03-29
add in user header after log in

dnoneill:


hudajkhan (approved):
Tried it out and works as expected

PR #2127, opened 2024-04-03
add logout for library users

### LINKED ISSUE #2123: Allow logout for library id or university id users
Currently only sso_users (users with sunet id) can log out.

dnoneill:
closes #2123  

Screenshot 2024-04-03 at 11 42 45 AM

### DIFF
--- a/app/views/shared/_sul_header.html.erb
+++ b/app/views/shared/_sul_header.html.erb
@@ -8,9 +8,10 @@
     <% end %>
     <nav aria-label="header menu">
       <ul class="list-unstyled d-flex gap-4 m-0">
-        <% if current_user.sso_user? %>
+        <% if current_user.sso_user? || current_user.library_id_user? %>
+          <% logout_path = current_user.site_admin? ? root_path : request.url %>
           <li>
-            <%= link_to t('.logout', sunetid: current_user.sunetid), logout_path(referrer: root_path) %>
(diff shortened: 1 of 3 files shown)


app/views/shared/_sul_header.html.erb
```diff
@@ -8,9 +8,9 @@
     <% end %>
     <nav aria-label="header menu">
       <ul class="list-unstyled d-flex gap-4 m-0">
-        <% if current_user.sso_user? %>
+        <% if current_user.sso_user? || current_user.library_id_user? %>
           <li>
-            <%= link_to t('.logout', sunetid: current_user.sunetid), logout_path(referrer: root_path) %>
+            <%= link_to t('.logout', sunetid: current_user.sunetid || current_user.library_id), logout_path(referrer: root_path) %>
```
cbeer, line 13: I don't think we want to send library id users to the root path. That page is kinda a dead-end for them:

Screenshot 2024-04-03 at 08 56 45

dnoneill: Do you have a suggestion of somewhere else? Right now we are sending sunet there as well but I am guessing because of the redirect in warden they aren't actually getting there.

cbeer: Yeah, that's tricky. For admin users, the root path makes sense.. but if it's a normal patron's placing a request, maybe it makes sense to send them back to the first request page (where they can choose to login again, maybe as a different user?)

hudajkhan (approved):
Works as described for library pin users !

PR #2138, opened 2024-04-04
add in request preferences

### LINKED ISSUE #2105: Build out HTML for ILL vs local hold/recall
Screenshot 2024-03-28 at 12 48 05

dnoneill:
closes #2105 

Add form for request/hold. Make form build based on patron rights (in screenshots, Finely Fee doesn't have recall access so they don't get the option to get as soon as possible)

Screenshot 2024-04-04 at 11 07 27 AM

Screenshot 2024-04-04 at 11 07 45 AM

### DIFF
--- a/app/models/patron_request.rb
+++ b/app/models/patron_request.rb
@@ -109,6 +109,14 @@ class PatronRequest < ApplicationRecord
     self.barcodes = [barcode]
   end
 
+  def holdable_recallable_items
+    @holdable_recallable_items ||= items_in_location.filter { |item| item.recallable?(patron) && item.holdable?(patron) }
+  end
+
+  def patron
+    @patron ||= Folio::Patron.find_by(patron_key: patron_id)
+  end
+
   private
 
   # Returns default service point codes
--- a/app/views/patron_requests/_request_preferences.html.erb
+++ b/app/views/patron_requests/_request_preferences.html.erb
@@ -1,0 +1,23 @@
+<% if f.object.holdable_recallable_items.any? %>
+<p class="text-cardinal">
+  This item is currently checked out by another patron or otherwise unavailable. 
+</p>
+<fieldset class="mb-4">
+  <legend class="fs-6">Choose your preferred request method:</legend>
+  <div class="mt-2 form-check">
+    <%= f.radio_button :fulfillment_type, 'recall', class: 'form-check-input' %>
+    <%= f.label :fulfillment_type_recall, class: 'form-check-label' do %>
+      Expedite a copy from a partner library
+      <a href="https://library.stanford.edu/services/borrow-and-request#request"><i class="bi bi-box-arrow-up-right"></i> Learn more</a>
+    <% end %>
+  </div>
+  <div class="mt-2 form-check">
+    <%= f.radio_button :fulfillment_type, 'hold', class: 'form-check-input' %>
+    <%= f.label :fulfillment_type_hold, 'Wait for a Stanford copy to become available', class: 'form-check-label' %>
+  </div>
+</fieldset>
+<div class="mb-4">
+  <%= f.label :needed_date, 'Not needed after:', class: 'required-label' %>
+  <%= f.date_field :needed_date, required: true, min: Date.today, class: 'd-block' %>
+</div>
+<% end %>
--- a/app/controllers/patron_requests_controller.rb
+++ b/app/controllers/patron_requests_controller.rb
@@ -15,6 +15,7 @@ class PatronRequestsController < ApplicationController
   end
 
   def new
+    @request.patron_id = current_user.patron&.id
     current_request.assign_attributes(new_params)
   end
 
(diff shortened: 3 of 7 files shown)


app/assets/stylesheets/application_redesign.scss
```diff
@@ -132,4 +132,14 @@ summary .bi-chevron-right::before {
 label[for="user-header-hide"] {
   margin-left: -1.5em;
   margin-top: .6em;
+}
+
+input[type="radio"] {
+  width: 22px;
+  height: 22px;
+  margin-right: 4px;
+}
```
cbeer, line 141: Can we use the standard bootstrap styling for radio buttons here?
https://getbootstrap.com/docs/5.3/forms/checks-radios/#radios

I think/hope it's close (enough).

app/assets/stylesheets/application_redesign.scss
```diff
@@ -132,4 +132,14 @@ summary .bi-chevron-right::before {
...
+
+.new_patron_request input:not([type=checkbox]):not([type=radio]):not([type=submit]) {
```
cbeer, line 143: What's this styling controlling? It's a little confusing; is there a standard bootstrap way to do this?

app/assets/stylesheets/application_redesign.scss
dnoneill: Probably close enough, though I do think the radio button outline is a bit light, not sure if that is an accessibility problem. Automated checkers don't think so. Screenshots of before and after.
Screenshot 2024-04-04 at 11 23 54 AM
Screenshot 2024-04-04 at 11 36 40 AM

app/views/patron_requests/_request_preferences.html.erb
```diff
@@ -0,0 +1,25 @@
+<% unless f.object.holdable_recallable_items.empty? %>
```
cbeer, line 1: I feel like rubocop would have some style suggestions here (but I guess doesn't run on erb files):

Could we flip this condition to an `if` instead of an `unless`?
```suggestion
<% if f.object.holdable_recallable_items.any? %>
```

app/views/patron_requests/_request_preferences.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <div class="mt-2 form-check">
+    <%= f.radio_button :fulfillment_type, 'hold', :class => 'form-check-input' %>
+    <%= f.label :fulfillment_type_hold, 'Wait for a Stanford copy to become available', :class => 'form-check-label' %>
+  </div>
+</fieldset>
+
+<div class="mb-4">
+  <%= f.label :needed_date, 'Not needed after:', :class => 'required-label' %>
+  <%= f.date_field :needed_date, :required => true, min: Date.today, :required => true, :class => 'd-block' %>
+</div>
```
cbeer, line 23:
```suggestion
    <%= f.radio_button :fulfillment_type, 'recall', class: 'form-check-input' %>
    <%= f.label :fulfillment_type_recall, class: 'form-check-label' do %>
      Expedite a copy from a partner library
      <a href="https://library.stanford.edu/services/borrow-and-request#request"><i class="bi bi-box-arrow-up-right"></i> Learn more</a>
    <% end %>
  </div>
  <div class="mt-2 form-check">
    <%= f.radio_button :fulfillment_type, 'hold', class: 'form-check-input' %>
    <%= f.label :fulfillment_type_hold, 'Wait for a Stanford copy to become available', class: 'form-check-label' %>
  </div>
</fieldset>

<div class="mb-4">
  <%= f.label :needed_date, 'Not needed after:', class: 'required-label' %>
  <%= f.date_field :needed_date, required: true, min: Date.today, class: 'd-block' %>
</div>
```

cbeer: (More rubocop-ing, although also note the extra `required` on the `needed_date` too)

app/models/patron_request.rb
```diff
@@ -55,6 +55,11 @@ def barcode=(barcode)
     self.barcodes = [barcode]
   end
 
+  def holdable_recallable_items
+    patron = Folio::Patron.find_by(patron_key: patron_id)
+    items_in_location.filter { |item| item.recallable?(patron) && item.holdable?(patron) }
+  end
+
   private
```
cbeer, line 63: What do you think about extracting `patron` and memoize the result of this calculation (I'm assuming calculating the circ rules for each item could get expensive..) ? 

```suggestion
  def holdable_recallable_items
    @holdable_recallable_items ||= items_in_location.filter { |item| item.recallable?(patron) && item.holdable?(patron) }
  end
  
  def patron
      @patron ||= Folio::Patron.find_by(patron_key: patron_id)
  end

  private
```

dnoneill: I was thinking the same thing Friday about patron. I think we are going to be needing it for when we submit so I think its a good idea.

PR #2139, opened 2024-04-04
Push guest pre-auth check down to Ability

cbeer:


hudajkhan (approved):
Appears to make sense.

PR #2141, opened 2024-04-05
adding path, test, and handling for visitor registration

### LINKED ISSUE #2072: Extend the SessionsController to support guest registration (with name + email)

hudajkhan:
Closes #2072 .

What this pull request does:

- Extend Warden with a "register_visitor" strategy.  Valid? checks for either the name or email field.  Authenticate checks that both are present.  
- Add a "register_visitor" method to sessions controller and add path to the routes.rb.  The method checks if the Warden strategy for register_visitor authenticates. If so, it will redirect to the next page.  If not, an error message will be shown.
- Update the login form so that the form action now goes to the sessions controller register_visitor action.  Add the new patron request with step parameter as the referrer parameter. 
- Update en.yml to include the error alert message if the visitor registration form does not pass. 
- Update current user to handle name and email user initialization
- Add tests for both session controller and current user

Although the fields on the form don't allow empty values for name and email, it is possible to enter just a space each.  If that happens, the authenticate method will capture that and show an error. 

Screenshot 2024-04-05 at 9 41 26 AM

### DIFF
--- a/app/views/patron_requests/login.html.erb
+++ b/app/views/patron_requests/login.html.erb
@@ -83,7 +83,8 @@
                 </a></p>
                 <details>
                     <summary class="btn btn-link p-0">Proceed as visitor <i class="bi bi-chevron-right"></i></summary>
-                    <%= form_tag new_patron_request_path(new_params.merge(step: 'select')), method: 'POST', class: 'form p-3', data: { turbo: false } do %>
+                    <%= form_tag register_visitor_url, method: 'POST', class: 'form p-3', data: { turbo: false } do %>
+                        <%= hidden_field_tag :referrer, new_patron_request_path(new_params.merge(step: 'select')) %>
                         <div class="form-group mt-2">
                             <label class="form-label required-label" for="name">Name</label>
                             <input class="form-control" id="name" name="name" required />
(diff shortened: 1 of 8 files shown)


app/views/patron_requests/login.html.erb
```diff
@@ -89,7 +89,8 @@
                 <details>
                     <summary class="btn btn-link p-0<% unless @guest_can_request %> disabled<%end %>">Proceed as visitor <i class="bi bi-chevron-right"></i></summary>
                     <% if @guest_can_request %>
-                        <%= form_tag new_patron_request_path(new_params), method: 'POST', class: 'form p-3', data: { turbo: false } do %>
+                        <%= form_tag register_visitor_url, method: 'POST', class: 'form p-3', data: { turbo: false } do %>
+                            <%= hidden_field_tag :referrer, new_patron_request_path(new_params.merge(step: 'select')) %>
                             <div class="form-group mt-2">
```
hudajkhan, line 94: Originally, the new_patron_request_path had "step=select" as a parameter.  Just want to check that we still want this parameter, since the form tag in the current code doesn't include "merge(step:'select')".  This has been moved to the referrer field because we want to submit the form to the session controller

PR #2143, opened 2024-04-05
render html page when there is no schedule

### LINKED ISSUE #2132: Fix location estimate to say no estimate for items we have no estimate for
http://localhost:3000/patron_requests/new?instance_hrid=14720993&origin_location_code=GRE-STACKS&step=select

When you load this page the estimated delievery date keeps saying loading.... because http://localhost:3000/paging_schedule/GRE-STACKS/GREEN-LOAN returns an error. We should update the estimate to provide a no estimate available if this is true.

dnoneill:
close #2132 

This fixes that is there is no schedule for the item it will not staying `Loading....`

It does change the error page a little bit for paging scheduler

Screenshot 2024-04-05 at 3 57 40 PM

Main
Screenshot 2024-04-05 at 3 58 37 PM

This branch

Screenshot 2024-04-05 at 3 58 24 PM

### DIFF
--- a/app/controllers/paging_schedule_controller.rb
+++ b/app/controllers/paging_schedule_controller.rb
@@ -5,7 +5,9 @@
 ###
 class PagingScheduleController < ApplicationController
   rescue_from PagingSchedule::ScheduleNotFound do
-    render plain: 'Schedule not found', status: :not_found
+    turbo_frame = helpers.content_tag('turbo-frame', style: 'display:none', id: 'earliestAvailable') { 'No delievery estimate' }
+    turbo_frame += helpers.content_tag('body') { 'Schedule not found' }
+    render html: turbo_frame, layout: false, status: :not_found
   end
 
   before_action only: [:show, :open] do


app/controllers/paging_schedule_controller.rb
```diff
@@ -5,7 +5,9 @@
 ###
 class PagingScheduleController < ApplicationController
   rescue_from PagingSchedule::ScheduleNotFound do
-    render plain: 'Schedule not found', status: :not_found
+    turbo_frame = helpers.content_tag('turbo-frame', style: 'display:none', id: 'earliestAvailable') { 'No delievery estimate' }
+    turbo_frame += helpers.content_tag('body') { 'Schedule not found' }
+    render html: turbo_frame, layout: false, status: :not_found
```
cbeer, line 10: When reviewing, we should make sure the current estimated delivery date widget is ok with this change. It may not be turbo-aware (I'm uncertain if we can do `respond_to` in a `rescue_from` or not).

PR #2145, opened 2024-04-05
Port the PIN reset workflow from mylibrary

### LINKED ISSUE #2073: Pull over reset pin workflow from MyLibrary
Image

E.g. from MyLibrary:
Image

thatbudakguy:
Closes #2073

- pulls over PIN reset functionality from mylibrary
- sets up redirecting back to your request after a PIN reset
- updates the UI for all flash messages to match the designs more closely
- re-implements the hide/show password toggle as a stimulus controller


app/controllers/application_controller.rb
```diff
@@ -30,4 +30,16 @@ def rescue_can_can(exception)
...
+    if params[:referrer]
+      redirect_to post_action_redirect_url
+    else
+      redirect_back fallback_location: root_url
+    end
+  end
+
+  def post_action_redirect_url
+    params[:referrer].presence || root_url
+  end
```
thatbudakguy, line 44: pulled this up here and renamed slightly to make it more generic, since it's now used in multiple controllers

PR #2151, opened 2024-04-08
Adding placeholder patron for registered visitor

### LINKED ISSUE #2149: Create placeholder patron object for guest registration
Logic in the remainder of the app will depend on being able to retrieve a patron object.

For the user associated with a guest/visitor, we want to create a Patron that uses the SUL Purchased group as the patron group.

hudajkhan:
Closes #2149 .

Background: At various points, the new requests workflow requires the existing of a patron object with specific fields (or would use the information, like name, for display).  We need to create a placeholder patron object to support the workflow for a registered visitor. 

What this pull request does:

- In the user class, the patron method returns a Patron object depending on various conditions.  This pull requests adds a condition for a name and email user i.e. registered visitor, where a placeholder Patron object is returned.  The object uses the patron group id for the sul-purchased patron group.  In addition, the id is set to a value that is added to Settings to indicate a registered visitor user.  The id is necessary b/c the page throws an error without the id.  Also, the name the user adds is set to "last name" for the patron.  The name field doesn't distinguish between first and last name, and there is no general "name" field in patron.  Using "last name" allows the resulting user header message to display the name the registered visitor entered on the request landing page.
- In the patron class, the blocks method checks if the user is a registered visitor by looking at the id.  If the user is a registered visitor, the method returns an empty array instead of trying to check the id with Folio client (which would throw an error).
- Adds tests for patron and user

### DIFF
--- a/app/models/folio/patron.rb
+++ b/app/models/folio/patron.rb
@@ -154,6 +154,10 @@ module Folio
       @policy_service ||= Folio::CirculationRules::PolicyService.new(patron_groups: [patron_group_id])
     end
 
+    def visitor_patron?
+      id == Settings.folio.visitor_placeholder_id
+    end
+
     private
 
     def patron_summary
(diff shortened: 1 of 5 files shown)


thatbudakguy (approved):
looks good! thanks for the detailed explanation.

PR #2164, opened 2024-04-11
update badge to reflect visitor, update library locations for visitors

### LINKED ISSUE #2162: Restrict pickup locations for name/email patrons
Users should NOT be able to select Law or Business. They should definitely be able to select Green, East Asia, Art, and Lane. (Sarah is confirming the rest of the SUL libraries.)

dnoneill:
closes #2162

### DIFF
--- a/app/models/patron_request.rb
+++ b/app/models/patron_request.rb
@@ -45,9 +45,12 @@ class PatronRequest < ApplicationRecord
 
   # FOLIO
   def pickup_destinations
-    return (default_pickup_service_points + additional_pickup_service_points).uniq if location_restricted_service_point_codes.empty?
+    destinations = (default_pickup_service_points + additional_pickup_service_points).uniq if location_restricted_service_point_codes.empty?
+    destinations ||= location_restricted_service_point_codes
 
-    location_restricted_service_point_codes
+    return destinations.select { |destination| Settings.allowed_visitor_pickups.include?(destination) } if patron.visitor_patron?
+
+    destinations
   end
 
   # Find service point which is default for this particular campus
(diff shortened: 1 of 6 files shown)


app/views/patron_requests/_user_header.html.erb
```diff
@@ -9,7 +9,7 @@
       <div class="d-flex align-items-center">
         <div class="h4">Welcome <%= current_user.patron.display_name %>!</div>
         <span class="badge ms-2 mb-2 rounded-4 bg-jungle-green">
-          <%= current_user.patron.patron_group['desc'] %>
+          <%= current_user.patron&.id == 'visitor' ? 'Visitor' : current_user.patron.patron_group['desc'] %>
```
cbeer, line 12: It looks like there's a method on the patron model for this?

```suggestion
          <%= current_user.patron.visitor_patron? ? 'Visitor' : current_user.patron.patron_group['desc'] %>
```

config/settings.yml
```diff
@@ -663,4 +663,8 @@ paging_schedule:
 hold_instead_of_recall:
   - 'In process'
   - 'On order'
-  - 'In transit'
\ No newline at end of file
+  - 'In transit'
+
+unallowed_visitor_pickups:
+  - 'BUS-IDESK'
+  - 'LAW'
```
cbeer, line 670: I think we probably want to allow-list the ok pickup locations? I think there's only a couple pickup locations that can also handle visitor registration, right?

app/models/patron_request.rb
```diff
@@ -45,6 +45,8 @@ def pickup_service_point
 
   # FOLIO
   def pickup_destinations
+    return Settings.allowed_visitor_pickups if patron.visitor_patron?
```
cbeer, line 48: This still needs to respect the location_restricted_service_point_codes, too, right? (e.g. PAGE-EA can only be picked up in EAST-ASIA)

dnoneill: Yup. Thanks.

PR #2174, opened 2024-04-15
PIN reset fixes

thatbudakguy:
- Update PIN reset message text
- Use the right name for the PIN reset identifier
- Use the correct subject for PIN reset emails
- Ensure the reset PIN link in emails is encoded correctly
- Enforce our requirements for PIN resets via validation

### DIFF
--- a/app/views/reset_pins_mailer/reset_pin.html.erb
+++ b/app/views/reset_pins_mailer/reset_pin.html.erb
@@ -9,7 +9,7 @@
       change the PIN for account <b><%= @patron.barcode %></b>.
     </p>
     <p style="font-size: 18pt">
-      <a href="<%= @url %>">Change my PIN</a>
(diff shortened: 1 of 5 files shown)


app/views/reset_pins_mailer/reset_pin.html.erb
```diff
@@ -9,7 +9,7 @@
       change the PIN for account <b><%= @patron.barcode %></b>.
     </p>
     <p style="font-size: 18pt">
-      <a href="<%= @url %>">Change my PIN</a>
+      <%= link_to 'Change my PIN', @url.html_safe %>
```
cbeer, line 12:
```suggestion
      <%= link_to 'Change my PIN', @url %>
```

Doesn't `link_to` just take care of that?

PR #2178, opened 2024-04-16
update ability to reflect name_email_user has a patron

dnoneill:


cbeer:
I think #2190 and/or #2180 fixed this too.

PR #2189, opened 2024-04-17
Fix up patron group description

cbeer:
I think this is copy/pasta when creating an accessor on the patron model?


dnoneill:
Yes. It was.

dnoneill (approved):

PR #2198, opened 2024-04-18
Update accordion steps to use bootstrap accordion behaviors to toggle other sections closed

cbeer:
- by adding `data-bs-parent` and Bootstrap's `Collapse` (instead of simulating a click), bootstrap can handle hiding + showing the different sections for us
- clarify (and make explicit) the placeholder sections
- pull step renumbering up to a shared method (fixing an off-by-one error with the multi-item case)

### DIFF
--- a/app/javascript/controllers/patron_request_controller.js
+++ b/app/javascript/controllers/patron_request_controller.js
@@ -36,17 +36,17 @@ export default class extends Controller {
   }
 
   nextStep(event) {
-    const step = 'patron_request[' + event.target.id + ']';
+    const step = 'patron_request[' + event.params.step + ']';
     const formdata = new FormData(this.element);
     const nextstepid = formdata.get(step.replace('-', '_'));
     if (!nextstepid) { return }
 
-    const accordionbutton = document.querySelector(`[data-bs-target='#${event.target.id}']`);
+    const accordionbutton = document.querySelector(`[data-bs-target='#${event.params.step}']`);
     accordionbutton.parentElement.classList.add('completed');
     accordionbutton.click();
     accordionbutton.setAttribute('disabled', '');
     const accordions = Array.from(document.querySelectorAll('.accordion-item:not([id*="placeholder"])'));
-    const current = accordions.findIndex( x => x.id.indexOf(event.target.id) > -1  );
+    const current = accordions.findIndex( x => x.id.indexOf(event.params.step) > -1  );
     var nextitem = accordions.at(current+1).id.split('-accordion')[0];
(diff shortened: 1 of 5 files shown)


app/javascript/controllers/patron_request_controller.js
```diff
@@ -1,82 +1,87 @@
...
 
+  removeAccordionStyling() {
+    this.element.classList.remove('accordion');
+
+    const accordion = this.accordionTargets[0];
+    accordion.classList.remove('d-none');
+    accordion.classList.remove('shadow-sm');
+    accordion.querySelector('.accordion-header').remove();
+
+    Collapse.getOrCreateInstance(this.accordionTargets[0].querySelector('.accordion-collapse')).show();
```
dnoneill, line 24: replacing with below seems to remove the accordion effect on the page.
const collpase = this.accordionTargets[0].querySelector('.accordion-collapse');
collpase.classList.remove('collapse');
collpase.classList.remove('accordion-collapse');

dnoneill (approved):
I have some questions because I am not sure I fully understand what is going on but it works well locally.

app/javascript/controllers/patron_request_controller.js
cbeer: I just pushed up [89b29a7](https://github.com/sul-dlss/sul-requests/pull/2198/commits/89b29a73dd426c142a96ad5643fdcedaf78ae703), which does all this at render-time.

dnoneill:
@cbeer can you rebase and I will merge.

app/views/patron_requests/new.html.erb
```diff
@@ -30,8 +33,8 @@
     </div>
 
     <% if f.object.items_in_location.one? %>
-      <div class="accordion-item my-4 shadow-sm" id="placeholder2-accordion">
-        <%= render 'accordion_header', title: 'Scan or pickup request', target: 'placeholder2' %>
+      <div class="accordion-item my-4 shadow-sm step-placeholder accordion-step-<%= scan_or_pickup_step ||= step_enum.next %>" id="scan-or-pickup-placeholder" data-patronRequest-forRequestType="none" data-patronRequest-target="accordion">
```
dnoneill, line 36: <%= scan_or_pickup_step ||= step_enum.next %> should be 
<%= current_step = (scan_or_pickup_step ||= step_enum.next) %>

dnoneill (changes requested):

cbeer:
Thanks, hopefully that's fixed in 0225dc1.

dnoneill (approved):

PR #2203, opened 2024-04-18
Allow the item selector to initialize in the 'in progress' state

cbeer:


dnoneill (commented):
Shouldn't this be in the accordion header partial?

cbeer:
I kept it separate on all the other accordions. I guess I can easily imagine a scenario where a couple accordions might start expanded or something 🤷‍♂️

dnoneill:
I don't think we have placeholders accordions for any of our first steps. I guess I am wondering is there a reason that expanded: current_step == 1 couldn't be moved the accordion header and not placed in multiple places and it wouldn't expand anything other than the first step?

thatbudakguy (approved):

cbeer:
This gets us to consistency with the different starting states. When we have all the panels in place and want to refactor this, that'll be a good time to decide whether the first step is always the same as the initially expanded one or if there's some variation (or other benefit to being explicit)

PR #2204, opened 2024-04-18
Submit scan requests to ILLiad

cbeer:
Fixes #2161

PR #2209, opened 2024-04-19
Form fixes

cbeer:
These are included in #2204, but I've pulled some individual commits out to this PR so they may be easier to review.


app/views/patron_requests/show.html.erb
```diff
@@ -1,20 +1,30 @@
-<h1 class="fw-semibold">We receieved your request</h1>
+<h1 class="fw-semibold">We receieved your <%= @request.request_type == 'scan' ? 'scan' : 'pickup' %> request</h1>
 
 <div class="row">
     <div class="col-8">
```
thatbudakguy, line 4:
```suggestion
    <div class="col-lg-8">
```
i think we applied this elsewhere in the last design review session

thatbudakguy (approved):

Issue #2220, opened 2024-04-22
scan + pickup form will not submit due to required fields.


dnoneill:
When a form has both scan and pickup, the form will not submit because it believes that you need to fill out the pickup portion of the form. Specifically the needed date http://localhost:3000/patron_requests/new?instance_hrid=14029321&origin_location_code=SAL3-STACKS&step=select

PR #2223, opened 2024-04-22
Remove required attribute on input fields that are hidden

thatbudakguy:
This fixes https://github.com/sul-dlss/sul-requests/issues/2220.

### DIFF
--- a/app/javascript/controllers/patron_request_controller.js
+++ b/app/javascript/controllers/patron_request_controller.js
@@ -34,9 +34,27 @@ export default class extends Controller {
     this.accordionTargets.filter(e => e.dataset.patronrequestForrequesttype).forEach(el => {
       if (el.dataset.patronrequestForrequesttype == requestType) {
         el.classList.remove('d-none');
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/patron_request_controller.js
```diff
@@ -34,8 +34,30 @@ export default class extends Controller {
...
+        this.disableRequiredInputs(el);
+      }
+    });
+  }
+
+  // Temporarily disable required inputs that are children of the given element
+  // This is used to prevent validation errors when hidden/unused inputs are required
+  // The initial required state of the inputs is preserved via a data attribute
+  disableRequiredInputs(element) {
+    element.querySelectorAll('input, select, textarea').forEach(input => {
```
dnoneill, line 49: This can be replaced with querySelectorAll('[required]') and you won't need the if statement

thatbudakguy: that's smart! i was able to do it for the other one too.

PR #2228, opened 2024-04-23
add item notes to view

### LINKED ISSUE #2184: Integrate item notes properly for item selector, broken out from #2157
Broken out from #2157

dnoneill:
closes #2184 

How it would display if graphql was working. 
Screenshot 2024-04-23 at 12 44 26 PM

Right now graphql is returning null for `itemNoteType` which is a problem because public_note can't tell it's a public note.


app/models/folio/item.rb
```diff
@@ -87,6 +87,12 @@ def home_location
       permanent_location.code
     end
 
+    def call_number
+      return '(no call number)' if callnumber.blank?
+
+      callnumber
+    end
```
cbeer, line 94: Is it weird to have both`#call_number` and `#callnumber` with two different behaviors? Is there a better name for this value (`item_label`, maybe?) or a better place for this logic (maybe a helper, if it is display-only?)

dnoneill: moved to helper

PR #2231, opened 2024-04-23
Don't allow patrons to move to the next step unless the current accor…

cbeer:
…dion's required fields are populated.

### DIFF
--- a/app/javascript/controllers/patron_request_controller.js
+++ b/app/javascript/controllers/patron_request_controller.js
@@ -82,6 +82,7 @@ export default class extends Controller {
     const accordion = event.target.closest('.accordion-item');
     const formData = new FormData(accordion.closest('form'));
 
+// Don't allow moving to the next step unless all required fields are completed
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/patron_request_controller.js
```diff
@@ -80,6 +80,12 @@ export default class extends Controller {
 
   nextStep(event) {
     const accordion = event.target.closest('.accordion-item');
+    const formData = new FormData(accordion.closest('form'));
+
+    if (Array.from(accordion.querySelectorAll('[required], [data-required]')).find(x => formData.getAll(x.name).every(x => !x))) {
+      event.preventDefault();
+      return;
+    }
```
thatbudakguy, line 88:
```suggestion
// Don't allow moving to the next step unless all required fields are completed
    if (Array.from(accordion.querySelectorAll('[required], [data-required]')).find(x => formData.getAll(x.name).every(x => !x))) {
      event.preventDefault();
      return;
    }
```

thatbudakguy (approved):

Issue #2234, opened 2024-04-23
Send recalls to ILL if there is a request queue


thatbudakguy:
Spun out of #2160. This is new functionality requested by staff (not parity).

As clarified by @saseestone, this takes effect when:
- the request in question is a recall (checked out; 'expedite a copy from a partner library' selected)
- the patron is ILB eligible
- there is at least one existing request for the item in FOLIO (request queue of length > 0)

Note that we have [logic for querying request queue lengths in folio-graphql](https://github.com/sul-dlss/folio-graphql/blob/3b931e3abc964c011d7decf7ad27ef348767b003/src/folio/circulation-api.ts#L22-L31) already to support this functionality in mylibrary, but we _don't_ yet support querying the queue length for arbitrary items. Thus, this is blocked on work in folio-graphql to add the ability to get the queue length for any item.

The seam for deciding whether an item request ultimately goes to FOLIO or ILL is in the `SubmitPatronRequestJob`:
https://github.com/sul-dlss/sul-requests/blob/ca5259dfff9f6ce7e143894e8eac268efd3f92f3/app/jobs/submit_patron_request_job.rb#L8-L9

thatbudakguy:
Some things we decided:

- when placing a title-level hold, we always go to FOLIO (never to ILLiad) because of library policies that would likely result in other libraries rejecting the request anyway (because it's likely a new item that everyone is waiting for)
- there is no way to check the queue length of title-level holds in FOLIO.
- when placing an item-level hold, we check the queue length using the API provided by mod-circulation.
- mod-circulation also provides an API to check the total number of item-level holds for an _instance_ (i.e. the sum of all its item-level holds). we don't currently use this.

PR #2238, opened 2024-04-24
Send recalls to ILL for certain item statuses

### LINKED ISSUE #2160: Send recalls to ILL for some FOLIO item statuses
For the expedited case. Only for particular FOLIO patron groups, requesting materials that are not on shelf.

We should check in FOLIO if there are other requests on the Stanford copy. If yes, direct to ILLiad. 
If no other requests, and item has a FOLIO Item status of:

- Checked out
- Missing
- Aged to lost
- Claim returned
- Declared lost
- Long Missing
- Lost and paid
- Paged

Send request to ILLiad

thatbudakguy:
Closes #2160

Note that this doesn't handle the remaining recall to ILLiad case
where there is request queue; that's #2234.


app/jobs/submit_patron_request_job.rb
```diff
@@ -21,4 +20,32 @@ def perform(patron_request) # rubocop:disable Metrics/AbcSize
...
+  #   The item can be recalled, AND
+  #     The item has a status indicating it won't be available soon, OR
+  #     The item has an existing request queue (TODO: https://github.com/sul-dlss/sul-requests/issues/2234)
+  def send_to_illiad?(patron_request, item)
+    return true if patron_request.scan?
+    return false unless patron_request.patron.ilb_eligible?
+
+    patron_request.fulfillment_type == 'recall' &&
+      item.hold_recallable?(patron_request.patron) &&
+      !item.status.in?(folio_recall_statuses)
```
cbeer, line 39: What do you think about moving the item status piece up to a guard clause too? 


```suggestion
    return false unless patron_request.patron.ilb_eligible?
    return false if item.status.in?(folio_recall_statuses)

    patron_request.fulfillment_type == 'recall' &&
      item.hold_recallable?(patron_request.patron)
```

thatbudakguy: for sure!

PR #2239, opened 2024-04-24
Hook up an undo toast when removing a selected item

cbeer:
https://github.com/sul-dlss/sul-requests/assets/111218/d9de8d08-c289-4fd4-9a17-a9af0ba4f882


dnoneill:
If you choose pick up items and choose items, then change your mind go to scan and then chose an item and then try to remove it at the next step it breaks things.

dnoneill:
https://github.com/sul-dlss/sul-requests/assets/19173991/fc56d972-1f9e-4da7-ae05-16b6646a2734

cbeer:
I feel like I'm seeing that same behavior on `main` already?

PR #2251, opened 2024-04-24
fix estimate to not display date when the item is not available

dnoneill:


dnoneill:
This also needs to be fixed in turbo. See http://localhost:3000/patron_requests/new?instance_hrid=in00000001563&origin_location_code=SAL3-STACKS

app/models/patron_request.rb
```diff
@@ -109,8 +109,12 @@ def destination_library_code
   end
 
   def earliest_delivery_estimate(scan: false)
-    paging_info = PagingSchedule.for(self, scan:).earliest_delivery_estimate
-    { 'date' => Date.parse(paging_info.to_s), 'display_date' => paging_info.to_s }
+    if items_in_location&.first&.available?
```
cbeer, line 112: I'm not sure I understand what we're trying to do here. Won't this show "No date/time estimate" if I've selected 2 items where one is available for paging and one isn't?

dnoneill: We only use earliest_delivery_estimate if there is one item.

cbeer: I think it's used a couple other places where there's multiple items:

The scan/pickup option: https://github.com/sul-dlss/sul-requests/blob/main/app/views/patron_requests/_scan_or_pickup.html.erb#L25

If you don't get a choice about pickup location: https://github.com/sul-dlss/sul-requests/blob/main/app/views/patron_requests/new.html.erb#L103

And on the status page: https://github.com/sul-dlss/sul-requests/blob/main/app/models/patron_request.rb#L16

dnoneill:
fixed with https://github.com/sul-dlss/sul-requests/pull/2277

PR #2252, opened 2024-04-24
Send recalls for items with a request queue to ILLiad

### LINKED ISSUE #2234: Send recalls to ILL if there is a request queue

thatbudakguy:
Closes #2234

### DIFF
--- a/app/jobs/submit_patron_request_job.rb
+++ b/app/jobs/submit_patron_request_job.rb
@@ -34,23 +34,28 @@ class SubmitPatronRequestJob < ApplicationJob
   #   The request is a recall, AND
   #   The item can be recalled, AND
   #     The item has a status indicating it won't be available soon, OR
-  #     The item has an existing request queue (TODO: https://github.com/sul-dlss/sul-requests/issues/2234)
+  #     The item has an existing request queue
   def send_to_illiad?(patron_request, item)
     return true if patron_request.scan?
     return false unless patron_request.patron.ilb_eligible?
-    return false if item.status.in?(folio_recall_statuses)
+    return false unless patron_request.fulfillment_type == 'recall'
+    return false unless item.hold_recallable?(patron_request.patron)
 
-    patron_request.fulfillment_type == 'recall' &&
-      item.hold_recallable?(patron_request.patron)
+    item.status.in?(illiad_recall_statuses) || item.queue_length.positive?
   end
 
-  # If an item has one of these statuses it's OK to request it via FOLIO;
-  # otherwise, try ILLiad first for a faster delivery
-  def folio_recall_statuses
+  # If an item has one of these statuses the patron won't get it anytime soon
+  # from FOLIO, so try ILLiad
+  def illiad_recall_statuses
     [
-      Folio::Item::STATUS_IN_PROCESS,
-      Folio::Item::STATUS_ON_ORDER,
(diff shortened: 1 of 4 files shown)


app/jobs/submit_patron_request_job.rb
```diff
@@ -33,14 +33,14 @@ def perform(patron_request)
...
   def send_to_illiad?(patron_request, item)
     return true if patron_request.scan?
     return false unless patron_request.patron.ilb_eligible?
-    return false if item.status.in?(folio_recall_statuses)
+    return false unless patron_request.fulfillment_type == 'recall'
+    return false unless item.hold_recallable?(patron_request.patron)
 
-    patron_request.fulfillment_type == 'recall' &&
-      item.hold_recallable?(patron_request.patron)
+    !item.status.in?(folio_recall_statuses) || item.request_queue_length.positive?
```
cbeer, line 43: This is getting harder to understand. Would it be any better if we inverted `folio_recall_statuses` to `illiad_recall_statuses`?

```
item.status.in(illiad_recall_statuses) && item_request_queue_length.zero?
```

Or maybe just distinct sections of guard clauses:

- scan always need to go to ILLiad
- the patron wants it fast (and is eligible for ILL?)
- the item will be delivered fastest by ILL

thatbudakguy: I did `folio_recall_statuses` because there were fewer and I wasn't sure we had enumerated all the possible ILLiad ones, but I can invert it.

If we did that, though, I think the final clause should still be `OR` rather than `AND` – having one of those statuses means we go to ILLiad regardless of whether there is a queue.

The "request is a recall and patron is eligible to recall" also apply to both of the latter two cases, which is why I pulled them up to guard clauses.

I think if we want this more readable we might be looking at pulling some more of it out into named methods instead of trying to restructure it, because the order is really tricky to get right.

cbeer: Whatever we can do to clarify things would be good. It's a lot of conditions and exceptions, and I'm clearly fuzzy on the request queue length part. I thought the (AND) condition was: send to do FOLIO if it's maybe imminently available (because of its status) and you'd be the first in the queue?

thatbudakguy: Updated the logic.

PR #2256, opened 2024-04-25
fix refresh after checking items

dnoneill:
How it works currently.
https://github.com/sul-dlss/sul-requests/assets/19173991/45ff305d-125f-4fd7-8808-e08fd3501aa6

This removes the checks from everything after refresh

### DIFF
--- a/app/javascript/controllers/patron_request_controller.js
+++ b/app/javascript/controllers/patron_request_controller.js
@@ -8,6 +8,7 @@ export default class extends Controller {
     if (this.accordionTargets.length == 1) {
       this.removeAccordionStyling();
     }
+    this.element.reset();
   }
 
   showItemSelector(event) {


cbeer:
Should we do the same for e.g. text fields? Would a form.reset() do this too?

dnoneill:
It does.

app/javascript/controllers/patron_request_controller.js
```diff
@@ -8,6 +8,8 @@ export default class extends Controller {
     if (this.accordionTargets.length == 1) {
       this.removeAccordionStyling();
     }
+    const form = this.accordionTargets[0].closest("form");
+    form.reset();
```
cbeer, line 12:
```suggestion
    this.element.reset()
```

PR #2273, opened 2024-04-25
add sort to patron request

### LINKED ISSUE #2185: Add dynamic sorting for item selector
Broken out from #2157

dnoneill:
closes #2185

### DIFF
--- a/app/views/patron_requests/new.html.erb
+++ b/app/views/patron_requests/new.html.erb
@@ -60,14 +60,18 @@
           <table class="table mb-0 border">
             <thead>
               <tr>
-                <th class="text-start">Call number</th>
-                <th class="text-end">Status</th>
+                <th aria-sort="ascending" class="text-start" data-sortAsc=true data-sortfield="Callnumber" data-action='click->itemselector#sort'>
+                  <button class="btn btn-link pe-0 text-decoration-none text-black">Call number</button>
+                </th>
+                <th class="text-end" data-sortfield="Status" data-sortAsc=true data-action='click->itemselector#sort'>
+                  <button class="btn btn-link pe-0 text-decoration-none text-black">Status</button>
+                </th>
               </tr>
             </thead>
             <tbody>
-              <% sort_holdings(f.object.items_in_location).each do |item| %>
+              <% sort_holdings(f.object.items_in_location).each_with_index do |item, index| %>
                 <% not_requestable = cannot?(:request, item) %>
-                <tr>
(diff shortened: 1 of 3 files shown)


dnoneill:
if this works we should open a ticket in the backlog for search

app/javascript/controllers/patron_request_controller.js
```diff
@@ -18,6 +18,21 @@ export default class extends Controller {
     }
   }
 
+  sort(event) {
+    const sortby = event.target.dataset.patronrequestSortfield;
+    const isAsc = event.target.dataset.patronrequestAsc == 'true' ? true : false;
+    event.target.dataset.patronrequestAsc = !isAsc;
```
cbeer, line 24: I think stimulus would like use to use their values to manage the current state:
https://stimulus.hotwired.dev/handbook/managing-state

I wonder if the item selector controller (or a new one?) would be a better place to manage this behavior

app/views/patron_requests/new.html.erb
```diff
@@ -60,16 +60,16 @@
           <table class="table mb-0 border">
             <thead>
               <tr>
-                <th class="text-start">Call number</th>
-                <th class="text-end">Status</th>
+                <th class="text-start" data-patronRequest-sortfield="Callnumber" data-patronRequest-asc=true data-action='click->patronRequest#sort'>Call number</th>
```
cbeer, line 63: Do we need to do something additional to make these accessible (e.g. a button, or add it to the tabindex) or add a pointer cursor?

dnoneill: I think it is okay because of this? 

`Header cells must be marked up with <th>, and data cells with <td> to make tables accessible. For more complex tables, explicit associations may be needed using scope, id, and headers attributes.`

https://www.w3.org/WAI/tutorials/tables/

cbeer: Hm, does that account for interactivity?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort suggests buttons and an `aria-sort`?

dnoneill: added button and aria-sort

cbeer: That page also says:
> The aria-sort attribute should only be added to a single table or grid header at a time. The attribute is set to inform assistive technology users which column or row is sorted.

dnoneill: Oops. Thanks for the catch. Update now pushed.

PR #2277, opened 2024-04-25
don't allow estimated delievery date if there are no items that are a…

dnoneill:
…vailable


app/models/patron_request.rb
```diff
@@ -108,9 +108,17 @@ def destination_library_code
     destination_location&.library&.code
   end
 
+  def any_items_avaliable?
+    items_in_location.any? {|item| item.available? }
```
cbeer, line 112:
```suggestion
    items_in_location.any?(&:available?)
```

Issue #2159, opened 2024-04-11
Finish populating confirmation pages for the different types of requests


saseestone:
Preliminary work was done in https://github.com/sul-dlss/sul-requests/pull/2154, but needs additional:
- additional logic to account for the different types of requests (scan vs page vs hold/recall vs mediated page)
- style refinement
- item-level information

hudajkhan:
There may be an additional design screen added for this (multiple item screen)

thatbudakguy:
Note that the confirmation page design also shows the call number next to the title in parentheses if it was a single item, which we don't currently do.

saseestone:
I think in prod, we might not be sending confirmation mailers for some request types. After conversations with ILL, I think we need to plan to send a confirmation mailer for all request types from request app. 

It may result in multiple process emails sent to the user (e.g. one from request app, another from ILLiad alerting the user that BD or ILL will be used to fulfill the request, etc.) but more seems better than a.) confusion about what app is sending what and b.) reliance on other library departments to actually send confirmations.

PR #2286, opened 2024-04-26
add in confirmation mailer

### LINKED ISSUE #2240: We need to send confirmation mailers
In testing, I've submitted several tests and haven't received a confirmation email. Are we sending them? If not, we should :) 

At first, I thought it was because the user record had a test email address (dlss-access-team...or foliotesting@...) But I've updated a couple of users to use my email address and the mailers haven't come.

dnoneill:
closes #2240 

Believe it will need to be updated when https://github.com/sul-dlss/sul-requests/issues/2159 is closed

### DIFF
--- a/spec/mailers/previews/patron_request_mailer_preview.rb
+++ b/spec/mailers/previews/patron_request_mailer_preview.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+# Preview all emails at http://localhost:3000/rails/mailers/patron_request
+class PatronRequestMailerPreview < ActionMailer::Preview
+end
--- a/app/controllers/patron_requests_controller.rb
+++ b/app/controllers/patron_requests_controller.rb
@@ -9,7 +9,9 @@ class PatronRequestsController < ApplicationController
   before_action :associate_request_with_patron, only: [:new, :create, :login]
   helper_method :current_request, :new_params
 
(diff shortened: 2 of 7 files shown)


spec/mailers/previews/patron_request_mailer_preview.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# Preview all emails at http://localhost:3000/rails/mailers/patron_request
```
hudajkhan, line 3: When I go to localhost:3000/rails/mailers, the path for patron_request seems to be http://localhost:3000/rails/mailers/patron_request_mailer .  Assume the actions would be filled in with the work on the other issue #2159

hudajkhan:
The rest looks good. When I try to see what the HTML in the email would look like, it has all the content of the confirmation page.  The layout seems suitable for an email (i.e. doesn't replicate a separate column for library resources, but just lists them in sections after request confirmation).

spec/mailers/previews/patron_request_mailer_preview.rb
hudajkhan: In the interest of alacrity, I'm going to make a comment on the other confirmation mailer issue to update the URL for this comment.

hudajkhan (approved):

PR #2297, opened 2024-04-29
allow aged to lost items to be held

### LINKED ISSUE #2296: Allow aged to lost requests

dnoneill:
closes #2296 

Used http://localhost:3000/patron_requests/new?instance_hrid=929827&origin_location_code=SAL3-STACKS for testing.

This PR also allows aged to lost items to show up in the viewer. Previously they were filtered out.


app/jobs/submit_patron_request_job.rb
```diff
@@ -36,6 +36,7 @@ def perform(patron_request)
   #     The item has an existing request queue (TODO: https://github.com/sul-dlss/sul-requests/issues/2234)
   def send_to_illiad?(patron_request, item)
     return true if patron_request.scan?
+    return true if item.status == Folio::Item::STATUS_AGED_TO_LOST
     return false unless patron_request.patron.ilb_eligible?
```
cbeer, line 40:
```suggestion
    return false unless patron_request.patron.ilb_eligible?
    return true if item.status == Folio::Item::STATUS_AGED_TO_LOST
```

maybe?

@saseestone: what should we do with aged-to-lost items for patrons that aren't eligible for ILL?

cbeer: Confirmed at standup that patrons who aren't otherwise eligible shouldn't get sent to ILLiad.

PR #2343, opened 2024-05-01
Send ILB staff a message when ILLiad requests fail

cbeer:
... for parity with the current implementation.

### DIFF
--- a/app/views/ilb_mailer/failed_ilb_notification.text.erb
+++ b/app/views/ilb_mailer/failed_ilb_notification.text.erb
@@ -1,0 +1,25 @@
+On <%= l @patron_request.created_at, format: :short %>, <%= @patron_request.patron.display_name %> <<%= @patron_request.patron.email %>> requested the following:
+
+https://searchworks.stanford.edu/view/<%= @patron_request.instance_hrid %>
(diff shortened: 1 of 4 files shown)


app/views/ilb_mailer/failed_ilb_notification.text.erb
```diff
@@ -0,0 +1,25 @@
+On <%= l @patron_request.created_at, format: :short %>, <%= @patron_request.patron.display_name %> <<%= @patron_request.patron.email %>> requested the following:
+
+https://searchworks.stanford.edu/view/<%= @patron_request.instance_hrid %>
+
+However, there was a problem with the ILLiad API request:
+
+<%= @illiad_response['Message'] %>
```
hudajkhan, line 7: I tried to hack this to force this to pass me an error message.  Is it possible that illiad_response could be nil? B/c if so, this will throw an error. Thanks.

cbeer: I'm not sure, but it can't hurt to be defensive about it 🤷‍♂️

hudajkhan (approved):

PR #2348, opened 2024-05-01
fix scan pilot to user patron_group

dnoneill:
Based on thread here: https://stanfordlib.slack.com/archives/G018TJ20XGE/p1714596615304799 and https://github.com/sul-dlss/sul-requests/issues/2160

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -148,6 +148,6 @@ class Ability
   # rubocop:enable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/MethodLength
 
   def in_scan_pilot_group?(user)
-    user.affiliation.any? { |g| Settings.scan_pilot_groups.include? g }
+    Settings.folio.scan_pilot_groups.include?(user.patron.patron_group_name)
   end
 end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -2,29 +2,6 @@
 super_admin_groups: ['sul:requests-super-admin']
 site_admin_groups: ['sul:requests-site-admin']
 
-scan_pilot_groups:
-  - 'stanford:student'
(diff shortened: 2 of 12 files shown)


config/settings.yml
```diff
@@ -3,27 +3,10 @@ super_admin_groups: ['sul:requests-super-admin']
...
-  - 'stanford:staff:emeritus'
-  - 'stanford:staff:otherteaching'
-  - 'stanford:staff:affiliate'
-  - 'stanford:affiliate:fellow'
-  - 'stanford:affiliate:visitscholarvs'
-  - 'stanford:affiliate:visitscholarvt'
+  - 'student'
+  - 'faculty'
+  - 'staff'
+  - 'affiliate'
```
cbeer, line 9: The list of groups in #2160 is a little longer:

```suggestion
  - 'faculty'
  - 'fellow'
  - 'graduate'
  - 'postdoctoral'
  - 'staff'
  - 'undergrad'
  - 'visiting scholar'
```

dnoneill: thanks

app/models/ability.rb
```diff
@@ -148,6 +148,6 @@ def initialize(user, token = nil)
   # rubocop:enable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/MethodLength
 
   def in_scan_pilot_group?(user)
-    user.affiliation.any? { |g| Settings.scan_pilot_groups.include? g }
+    Settings.folio.scan_pilot_groups.include?(user.patron.patron_group_name)
```
cbeer, line 151: Don't we need to keep checking the user affiliation for current requests, because we don't know their patron until after they submit the request?

cbeer: In the current system, users don't log in until they submit the request.

PR #2357, opened 2024-05-02
Initialize the item selector using the data from the DOM

cbeer:
Otherwise, if you go to the form with a barcode in the URL the item form is blank when it initializes, e.g.

http://127.0.0.1:3000/patron_requests/new?barcode=36105028701352&instance_hrid=4678196&origin_location_code=MUS-SCORES

### DIFF
--- a/app/javascript/controllers/item_selector_controller.js
+++ b/app/javascript/controllers/item_selector_controller.js
@@ -1,11 +1,35 @@
 import { Controller } from "@hotwired/stimulus";
 import { Toast } from 'bootstrap';
 
+function camelize(value) {
+  return value.replace(/(?:[_-])([a-z0-9])/g, (_, char) => char.toUpperCase())
+}
+
+function typecast(value) {
+  try {
+    return JSON.parse(value)
+  } catch (o_O) {
+    return value
+  }
+}
+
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/item_selector_controller.js
```diff
@@ -130,4 +154,19 @@ export default class extends Controller {
       `;
     }).join('');
   }
+
+  getStimulusParams(element) {
+    const params = {}
+    const pattern = new RegExp(`^data-${this.identifier}-(.+)-param$`, "i")
+
+    for (const { name, value } of Array.from(element.attributes)) {
+      const match = name.match(pattern)
```
thatbudakguy, line 163: would it not work to iterate over `element.dataset`?

thatbudakguy (approved):

app/javascript/controllers/item_selector_controller.js
cbeer: This is copied directly from stimulus 🤷‍♂️

thatbudakguy: oh wild. well, can't argue with that, I guess

PR #2360, opened 2024-05-02
Adapt for title request

hudajkhan:
What this pull request does:

Ensure that when there are no items, the page knows how to handle a request that will translate to a title level request on the backend.

- Allows handling of no items, by specifying that this is a title level request
- Displays status of "On order" and shows "no waitlist", with no estimated date.
- Displays "not needed after" defaulting to a year from now.  This is now a partial shared between this page and request_preferences. 
- Ensures folio_client method properly uses the instance_id to fulfill the title level hold request


app/helpers/requests_helper.rb
```diff
@@ -45,7 +45,9 @@ def label_for_item_selector_holding(holding)
     end
   end
 
-  def queue_length_display(item, prefix: nil)
+  def queue_length_display(item, prefix: nil, title_only: false)
+    return "#{prefix}On order | No waitlist" if title_only
```
dnoneill, line 49: queueTotalLength works on instance in graphql. Also you should be able to get status I think? from Instance.

hudajkhan: The length will just be zero when there are no items, and I will check about queueTotalLength working in this case.   I'll check about status too.

app/models/patron_request.rb
```diff
@@ -146,6 +151,10 @@ def earliest_delivery_estimate(scan: false)
     { 'date' => Time.zone.today, 'display_date' => 'No date/time estimate' }
   end
 
+  def title_request_expiration_date
+    Time.zone.today + 1.year
+  end
+
```
dnoneill, line 157: don't think you need this. Think you need to change earliest_delievery_estiamate to 
`if title_only? || any_items_avaliable?`

app/helpers/requests_helper.rb
hudajkhan: "Status" is also returning null. I'll check with @saseestone about that.

dnoneill: That instance probably has a request on it which is why it is returning 1 but if we don't need to check that is moot I think.

Yeah I don't know why I thought instance or maybe holdings had some kind of status but it doesn't look like it on the record you were looking at.

hudajkhan: It has a status attribute, but no value is returning, which is .. I don't know why :)

hudajkhan: There is a holdings summary attribute on holdingsRecord but that too is yielding nothing in this case for order status

app/models/patron_request.rb
hudajkhan: I think I'll update earliest_delivery_estimate to handle no items and to return a year from now, b/c otherwise they return today's date.  I can then get rid of the title_request_expiration_date method

hudajkhan: I changed the earliest delivery estimate to handle the title only condition, b/c the existing conditions would just return today's date, and we want to have a year from now as the date .

hudajkhan:
If you have suggestions for more tests to add, please feel free to let me know.

app/models/patron_request.rb
```diff
@@ -195,6 +200,8 @@ def earliest_delivery_estimate(scan: false)
     if any_items_avaliable?
       paging_info = PagingSchedule.for(self, scan:).earliest_delivery_estimate
       { 'date' => Date.parse(paging_info.to_s), 'display_date' => paging_info.to_s }
+    elsif title_only?
+      { 'date' => Time.zone.today + 1.year, 'display_date' => 'No date/time estimate' }
```
dnoneill, line 204: I don't think this actually necessary. I think you can just add a year to line after.

hudajkhan: Based on discussion, I made a partial for the not needed after date section, since it's also used in request preferences.  I moved this part of the method out as no longer needed.

app/jobs/submit_patron_request_job.rb
```diff
@@ -80,7 +80,8 @@ def place_title_hold(patron_request)
     hold_request_data = FolioClient::HoldRequestData.new(pickup_location_id: patron_request.pickup_service_point.id,
                                                          patron_comments: patron_request.request_comments,
                                                          expiration_date: patron_request.needed_date.to_time.utc.iso8601)
-    folio_response = folio_client.create_instance_hold(patron_request.patron.id, patron_request.instance_hrid, hold_request_data)
+    # Use instance_id instead of instance_hrid b/c the former is what is required, the latter will leave off the "a" at the beginning
```
dnoneill, line 83: @hudajkhan  I think we can get rid of this. create_instance_hold doesn't allow for hrid so that might be confusing without the commit history. https://github.com/sul-dlss/sul-requests/blob/562a92559c342d5466ce8801b73f35be21b89a8f/app/services/folio_client.rb#L224

dnoneill (approved):

PR #2365, opened 2024-05-02
Prevent the item selector search from submitting the form on enter

### LINKED ISSUE #2364: Prevent the item selector filter box from being submittable
Reported by Jeanette:

> If you type in a search term and hit enter the select items goes away and there is no way to bring it back without starting your request over

thatbudakguy:
This solution creates a hidden, non-interactable submit button
at the beginning of the form, which becomes its "default submit"
button according to the HTML spec. Because hitting enter will
attempt to activate this button and it's disabled, nothing happens.

If you have the final, real submit button focused and hit enter,
that will work, because the explicit target of that action is the
(non-disabled) submit button.

This ensures that no elements in the form are allowed to submit
it accidentally except the final submit button.

Fixes #2364

### DIFF
--- a/app/views/patron_requests/new.html.erb
+++ b/app/views/patron_requests/new.html.erb
@@ -50,6 +50,9 @@
       <%= f.hidden_field :aeon_barcode, name: 'ItemNumber_1', value: single_barcode %>
     <% end %>
   <% end %>
+  <%# only allow submitting on enter if the final submit button is focused %>
+  <%# see: https://stackoverflow.com/a/51507806 %>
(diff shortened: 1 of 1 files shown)


thatbudakguy:
It is admittedly weird, but it also handles all of the fields on the scan form – currently if you hit enter in any of those, it jumps back to the first step of the form for some reason.

This way, no matter how many inputs of any kind we include anywhere, you can't accidentally submit via enter. And it works without javascript!

cbeer:
😬

PR #2391, opened 2024-05-03
integrating option for people to pick their sponsors in the request

### LINKED ISSUE #2249: Ask proxies which sponsor should see the request, if any
We know when a proxy is a member of a group now (and there are designs).

hudajkhan:
Closes #2249.

What this pull request does:

- Adds methods to FolioClient to retrieve the full list of names of sponsors for this patron
- Adds methods to Patron Request to check for and return this list of sponsors
- Updates the proxy partial to include a conditional: display the proxy form that already exists if the patron has proxies.  Otherwise check if the patron has sponsors and show that form instead 
- For the sponsor form, if they select that they want to do this request for their sponsor, another set of radio buttons with the sponsor selection are displayed.  If they change their selection to just themselves, the sponsor radio buttons are unchecked the next time they pick "sponsor" instead. 
- If they pick sponsor, the scan or pickup choice section will show a warning that scans can't be sent to sponsors.
- Patron comments in the submission process indicate who the sponsor is for the request. 

Note: By policy a user can only either be proxy or sponsor, although FOLIO allows both.  The code only allows one of these options to show, checking for the option where the patron has proxies first, and only checking if they have sponsors if the first condition is not true. 

Screenshots:

Screenshot 2024-05-03 at 8 36 05 PM

Screenshot 2024-05-03 at 8 36 17 PM

Screenshot 2024-05-03 at 8 36 32 PM

Screenshot 2024-05-03 at 8 36 36 PM


app/views/patron_requests/new.html.erb
```diff
@@ -69,12 +69,12 @@
...
 
   <!-- proxy -->
-  <% if f.object.patron.proxy_group_names.any? && !f.object.aeon_page? %>
+  <% if ( f.object.patron.proxy_group_names&.any? || f.object.patron.sponsor_names&.any? ) && !f.object.aeon_page? %>
     <%= render AccordionStepComponent.new(id: 'proxy', step_index: step_enum.next, form_id: f.id) do |c| %>
       <% c.with_title.with_content('Proxy/Sponsor') %>
       <% c.with_body do %>
           <%= render 'proxy', f: %>
-      <% end %>
+      <% end %>y7
```
thatbudakguy, line 77:
```suggestion
      <% end %>
```

app/models/folio/patron.rb
```diff
@@ -159,6 +172,12 @@ def all_proxy_group_info
       @all_proxy_group_info ||= self.class.folio_client.all_proxy_group_info(id)
     end
 
+    # Get all the sponsors for which this patron is a proxy
+    def all_proxy_info
+      @all_proxy_info ||= user_info.dig('stubs', 'all_proxy_info') # used for stubbing
```
thatbudakguy, line 177: what does "used for stubbing" mean here? it doesn't look like this behavior gets used in the test added for this PR, but I'm not 100% sure

app/javascript/controllers/patron_request_controller.js
```diff
@@ -45,6 +45,29 @@ export default class extends Controller {
...
+  }
+
+  // When the user says the request is for a sponsor, display the sponsor selection list
+  toggleDisplaySponsorList(event) {
+    if(event.target.value == 'share') {
+      this.selectSponsorTarget.classList.remove('d-none')
+    } else {
+      this.selectSponsorTarget.classList.add('d-none')
+      //this.event.target.checked = false
+      this.sponsorRadioButtonTarget.checked = false
```
thatbudakguy, line 67:
```suggestion
      this.selectSponsorTarget.classList.add('d-none')
      this.sponsorRadioButtonTarget.checked = false
```

thatbudakguy (commented):
this looks good overall! let's squash it before merging

app/views/patron_requests/new.html.erb
hudajkhan: Thank you for catching that!

app/models/folio/patron.rb
thatbudakguy: ah, I see – this PR didn't change anything about this. thanks

thatbudakguy (approved):

PR #2401, opened 2024-05-06
Specifying requirement for sponsor selection 

hudajkhan:
The problem: The form was not proceeding to the next accordion step if the user selected "this request is for myself" because the sponsor selection radio button was also defined as required. 

What this pull request does:

- The sponsor selection radio button starts off as not required.  Whenever the user selects "this request is for myself", the sponsor selection radio buttons (i.e. buttons with sponsor names) are set to not required, allowing the form to continue.
- If, at any time, the user selects "this request is for a sponsor", then the sponsor radio buttons are set to required, making the user select a sponsor before being able to continue.  

Screenshots:

Screenshot 2024-05-06 at 1 18 17 PM

Screenshot 2024-05-06 at 1 18 30 PM

Screenshot 2024-05-06 at 1 18 40 PM

Screenshot 2024-05-06 at 1 18 52 PM

### DIFF
--- a/app/javascript/controllers/patron_request_controller.js
+++ b/app/javascript/controllers/patron_request_controller.js
@@ -86,12 +86,15 @@ export default class extends Controller {
   }
 
   // When the user says the request is for a sponsor, display the sponsor selection list
+  // Also require a sponsor selection only if the user selects that the request is for a sponsor. 
   toggleDisplaySponsorList(event) {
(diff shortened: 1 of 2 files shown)


hudajkhan:
Noticed more to add with respect to the fact that there are multiple sponsors and thus radio buttons possible.

PR #2426, opened 2024-05-07
use barcode param to filter items_in_location

### LINKED ISSUE #2427: Item level requests with several items
### Summary: 

Instance records with several items AND **item-level request buttons** should not go to an item selector, but instead to the request details form. Because item-level request buttons appear for locations that do not have scanning and/or paging services, these buttons only appear for items that are checked out or unavailable. That means request type is not an option, and items cannot be selected, and it should go to just the pickup request details, including referred pickup location, select processing method, and not needed after.

### Record and/or other information:
https://searchworks.stanford.edu/view/673369

Image

dnoneill:
closes #2427

### DIFF
--- a/app/models/patron_request.rb
+++ b/app/models/patron_request.rb
@@ -166,9 +166,17 @@ class PatronRequest < ApplicationRecord
 
   # Item stuff
 
+  def barcode_filtered_items
+    if barcodes.present?
+      bib_data.items.select { |x| x.barcode.in?(barcodes) || x.id.in?(barcodes) }
+    else
+      bib_data.items
+    end
+  end
+
   # @return [Array<Folio::Item>] the items in the origin location
   def items_in_location
-    @items_in_location ||= bib_data.items.select do |item|
+    @items_in_location ||= barcode_filtered_items.select do |item|
       if item.effective_location.details['searchworksTreatTemporaryLocationAsPermanentLocation'] == 'true'
         item.effective_location.code == origin_location_code
       else
--- a/spec/features/create_patron_request_spec.rb
+++ b/spec/features/create_patron_request_spec.rb
@@ -332,6 +332,12 @@ RSpec.describe 'Creating a request' do
       expect(PatronRequest.last).to have_attributes(barcodes: ['12345678'])
     end
 
+    it 'filters down to a single form when barcode in parameters' do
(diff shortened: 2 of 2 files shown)


app/models/patron_request.rb
```diff
@@ -169,7 +169,9 @@ def item_title
   # @return [Array<Folio::Item>] the items in the origin location
   def items_in_location
     @items_in_location ||= bib_data.items.select do |item|
-      if item.effective_location.details['searchworksTreatTemporaryLocationAsPermanentLocation'] == 'true'
+      if barcodes.length > 0
+        barcodes.include?(item.barcode)
```
cbeer, line 173:
```suggestion
        item.barcode.in?(barcodes) || item.id.in?(barcodes)
```

We do this in `selected_items` to handle items without barcodes. Presumably we should be consistent?

cbeer:
Could you also add an integration test to `create_patron_request_spec` for this case?

app/models/patron_request.rb
dnoneill: makes sense.

app/models/patron_request.rb
```diff
@@ -169,11 +169,19 @@ def item_title
...
+
+  # @return [Boolean] if the item is in the location/has a barcode
+  # split out due to robocop
+  def items_in_location_check(item)
+    if barcodes.present? && item.barcode.present?
+      item.barcode.in?(barcodes)
+    elsif item.effective_location.details['searchworksTreatTemporaryLocationAsPermanentLocation'] == 'true'
+      item.effective_location.code == origin_location_code
+    else
+      item.home_location == origin_location_code
```
cbeer, line 184: I'm not convinced extracting this method has added clarity or made the method more readable. If anything, it's more confusing now than before.

Maybe extracting `bib_data.items` to a method that applies the barcode filtering would be clearer (and appease rubocop?), or a method that does the filtering on top of `items_in_location` (`selectable_items`, `salient_items`?)

cbeer: I've pushed up https://github.com/sul-dlss/sul-requests/pull/2429 with one of those approaches.. and other than the name, I guess it seems like it'd be ok (and maybe less confusing than having `items_in_location` sometimes further filtered?)

dnoneill: Gotcha. Dealer's choice. I fixed it so it should be working with rubocop and tests. I will say that it might be good to have it further filtered by location. There might? be an instance where two different locations have the same barcode?

PR #2435, opened 2024-05-09
Use the sponsor's patron id when submitting requests to FOLIO

cbeer:
- Normal requests use the requesting patron's id (or the destination's pseudopatron.)
- Requests for a sponsor (shared with their proxy group) should use the patron's id with a request comment that their proxies can pick it up
- Requests by a proxy on behalf of a sponsor should be placed using the sponsor's patron id.

### DIFF
--- a/app/controllers/patron_requests_controller.rb
+++ b/app/controllers/patron_requests_controller.rb
@@ -76,7 +76,7 @@ class PatronRequestsController < ApplicationController
 
   def patron_request_params
     params.require(:patron_request).permit(:patron_email, :instance_hrid, :origin_location_code, :needed_date, :service_point_code, :proxy,
-                                           :for_sponsor_name, :for_sponsor,
+                                           :for_sponsor_id, :for_sponsor,
                                            :fulfillment_type, :request_type,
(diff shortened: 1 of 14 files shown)


hudajkhan:
@cbeer When I try this out for http://localhost:3000/patron_requests/new?instance_hrid=1485586&origin_location_code=SAL3-STACKS, I get the following error:
 ArgumentError (wrong number of arguments (given 0, expected 1)):
  
 | app/models/patron_request.rb:572:in `include?'
 | app/models/patron_request.rb:572:in `for_sponsor_id_is_valid'
 | app/controllers/patron_requests_controller.rb:27:in `create'

hudajkhan:
This might be because the sponsor radio button value is the sponsor name and not the id?

cbeer:
Thanks, give it another try.

hudajkhan:
Thanks @cbeer ! Works now.  The patron comment reads "(PROXY PICKUP OK; request placed by Piper Proxy <hudakhan@stanford.edu>)" (which I think is Piper Proxy's current email address).

hudajkhan (approved):

PR #2441, opened 2024-05-09
Re-add the login link to the header

cbeer:
We need it for admins + mediators to be able to log in, and seems fairly harmless if a name/email user clicks login while on the request page (or, at least, works like I'd expect -- they go back to the request page with their logged-in patron info 🤷‍♂️ )


dnoneill:
Would it make sense to only show the login button on non patron_request pages?

cbeer:
I don't think it matters. Like I said, the link seems to do what I'd expect in the one case where it appears 🤷‍♂️

dnoneill:
If you change the else to elsif controller_name != 'patron_requests'  then it won't show up on the login page but will show up on every other page. I am not sure it works how I would think on that page because we do have 3 other login options.

Screenshot 2024-05-10 at 10 45 13 AM

dbranchini:
Chris asked me to chime in here. Let's add it back to the header as "Staff login." Later (perhaps during unscheduled week), I'd like to consider having a library-specific footer on all apps with staff login there.

dnoneill:
@cbeer You can probably get rid of the changes made to spec

dnoneill (approved):

PR #2449, opened 2024-05-10
Remember that the user has dismissed the alert; fixes #2418

### LINKED ISSUE #2418: Once a user dismisses an alert, it should remain dismissed throughout the sessionl
### Steps to reproduce:

I dismissed the broadcast message on the request stepped form page, then clicked "Submit request" which goes to the confirmation page. 

### Actual results:

The alert appears again on the confirmation page.

### Expected results:

If we allow users to dismiss the alert, then it should persist throughout the session, and not reappear on the confirmation screen.

### Record and/or other information:

Image

cbeer:


dnoneill:
I am wondering if we should recreate what we have going for the user header or move the user header over to this.

cbeer:
I wondered about that (and built this so it'd be easy to do if we wanted), but unlike the broadcast messages, the user message only shows up on the one page 🤷‍♂️

dnoneill (approved):

PR #2470, opened 2024-05-14
Expand the patron request data column size

cbeer:


hudajkhan (approved):
Tried this out and worked without errors

PR #2481, opened 2024-05-20
Stop testing with ruby 3.2

cbeer:


jcoyne:
@cbeer the tests are still required on 3.2.

PR #2495, opened 2024-06-13
update view to warn user if they have an existing request or loan

### LINKED ISSUE #2247: Can we guard against duplicate requests?
Only for physical pickup.

Staff in ILL have reported having to cancel many requests in ILLiad because the user already has a FOLIO request on the title (i.e. a page or request for a new title) OR they have the item checked out.

In our checking of FOLIO requests for SUNet ID users, could we detect that the user already has a request for the item? If yes, it might make the most sense to display a message to the user about their previous request (or checkout). But also allow them to continue with the request.  Then if the duplication is on purpose, we're not blocking them. But we give them more information about previous requests.

dnoneill:
Screenshot 2024-06-14 at 12 18 42 PM
Screenshot 2024-06-14 at 12 10 39 PM
Screenshot 2024-06-14 at 12 09 54 PM

closes #2247

### DIFF
--- a/app/models/folio/patron.rb
+++ b/app/models/folio/patron.rb
@@ -166,6 +166,14 @@ module Folio
       user_info['active'] == false
     end
 
+    def holds
+      patron_summary['holds']
+    end
+
+    def loans
+      patron_summary['loans']
+    end
+
     private
 
     def valid_proxy_relation?(info)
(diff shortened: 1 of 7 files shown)


dnoneill:
Note this also shows up on the scan view

dbranchini:
Small issue, but any chance we can get the "On order" status on one line instead of wrapping? It should be call number and item status on one line and then the alert, if it exists, beneath it. Also, a question, for the multi-item radio buttons for scan requests, is the width of the alert in the multi-item select not as wide? It looks much smaller.

dnoneill:
Can do. No that was just my browser size was different when taking screenshots.

dnoneill:
@dbranchini Done. Was item note or bound-withs supposed to span the entire row (like this) as well or was it just this?
Screenshot 2024-06-14 at 17-09-54 Request African media barometer  Benin

dnoneill:
@dbranchini ☝️

PR #2500, opened 2024-06-18
Add temporary code for detecting FOLIO Poppy version

marlo-longley:
```
=> #<FolioClient:23960  @base_url="https://okapi-prod.stanford.edu">
irb(main):002> client.poppy?
=> false
irb(main):001> client = FolioClient.new
=> #<FolioClient:22920  @base_url="https://okapi-test.stanford.edu">
irb(main):002> client.poppy?
=> true
```

### DIFF
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -163,14 +163,32 @@ class FolioClient
     end
   end
 
+  # TODO: after Poppy launch delete this method
+  def okapi_version
+    @okapi_version ||= begin
+      response = get('/_/version')
+      check_response(response, title: 'Okapi version', context: {})
+      # response.body is a string
+      response.body
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


app/services/folio_client.rb
```diff
@@ -163,14 +163,30 @@ def as_json
     end
   end
 
+  # TODO: after Poppy launch delete this method
+  def okapi_version
+    response = get('/_/version')
+    check_response(response, title: 'Okapi version', context: {})
+    # response.body is a string
+    response.body
+  end
```
cbeer, line 172:
```suggestion
  def okapi_version
    @okapi_version ||= begin
      response = get('/_/version')
      check_response(response, title: 'Okapi version', context: {})
      # response.body is a string
      response.body
    end
  end
```

Should we memoize this so we don't keep hitting the version endpoint over and over?

PR #2521, opened 2024-07-09
Validate user input to prevent GraphQL injection attack

### LINKED ISSUE #2520: [sul-requests/prod] RuntimeError: Syntax Error: Unterminated string.
## Backtrace

    line 142 of [PROJECT_ROOT]/app/services/folio_graphql_client.rb: instance
    line 248 of [PROJECT_ROOT]/app/services/folio_client.rb: find_instance_by
    line 9 of [PROJECT_ROOT]/app/models/folio/instance.rb: fetch

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/108736137)

jcoyne:
Fixes #2520


app/controllers/patron_requests_controller.rb
```diff
@@ -39,6 +39,9 @@ def current_request
 
   def assign_new_attributes
     @patron_request.assign_attributes(**new_params)
+    # This validation prevents an injection attack in GraphQL. Ideally, this would be a model validation,
+    # but there are other validations that require first running a GraphQL query.
+    raise ActionController::BadRequest unless /\Aa?\d+\z/.match?(new_params[:instance_hrid])
```
cbeer, line 44: This regex is.. starts with an `a` or a digit? We also have hrids that start with `L`, or `in`, and probably arbitrary now in FOLIO.

jcoyne: Yes. I've updated it include those cases as well.

jcoyne: How about `\w+`?  It seems bad for people to be injecting quotes, etc.

app/controllers/patron_requests_controller.rb
```diff
@@ -39,6 +39,9 @@ def current_request
 
   def assign_new_attributes
     @patron_request.assign_attributes(**new_params)
+    # This validation prevents an injection attack in GraphQL. Ideally, this would be a model validation,
+    # but there are other validations that require first running a GraphQL query.
+    raise ActionController::BadRequest unless /\A(a|L|in)?\d+\z/.match?(new_params[:instance_hrid])
```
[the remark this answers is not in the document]
jcoyne: How about \w+? It seems bad for people to be injecting quotes, etc.

cbeer:
Obsoleted by #2524

PR #2583, opened 2024-10-28
fix library hours for service points

### LINKED ISSUE #1887: [sul-requests/prod] NoMethodError: undefined method `[]' for nil:NilClass
When you request https://searchworks.stanford.edu/view/in00000010636 and change to Spec it says Content missing

## Backtrace

    line 29 of [PROJECT_ROOT]/app/models/library_hours.rb: library_slug
    line 21 of [PROJECT_ROOT]/app/models/library_hours.rb: library_hours
    line 15 of [PROJECT_ROOT]/app/models/library_hours.rb: next_business_day

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49963/faults/100676486)

dnoneill:
closes #1887

### DIFF
--- a/app/models/paging_schedule.rb
+++ b/app/models/paging_schedule.rb
@@ -175,7 +175,8 @@ class PagingSchedule
       end
 
       def destination_library_hours_next_business_day_after_delivery
-        LibraryHours.new(to).next_business_day(origin_library_next_business_day)
+        to_code = Settings.libraries[to] ? to : Folio::Types.service_points.find_by(code: to)&.library&.code
+        LibraryHours.new(to_code).next_business_day(origin_library_next_business_day)
       end
     end
   end
--- a/spec/models/paging_schedule_spec.rb
+++ b/spec/models/paging_schedule_spec.rb
@@ -23,6 +23,13 @@ RSpec.describe PagingSchedule do
       expect(schedule.to).to eq 'GREEN'
     end
 
+    it 'maps GREEN-LOAN to GREEN returns the schedule for the provided request' do
+      schedule = described_class.for(from: nil, to: 'GREEN-LOAN', library_code: 'SAL3')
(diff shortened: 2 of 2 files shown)


spec/models/library_hours_spec.rb
```diff
@@ -22,4 +22,22 @@
       end
     end
   end
+
+  describe '#library_slug' do
```
jcoyne, line 26: Should there be a test for `location_map` for the context which you are fixing?

dnoneill: This is basically testing this. Location map returns a hash which is why I am using library_slug, because library_slug is just pulling a field from location_hash.

dnoneill:
@jcoyne you are suggesting something like this?

def origin_library_next_business_day
    from = Settings.libraries[from] ? from : Folio::Types.service_points.find_by(code: from)&.library&.code
    LibraryHours.new(from).next_business_day(as_of, business_days_later)
  end

  def destination_library_hours_next_business_day_after_delivery
    to = Settings.libraries[to] ? to : Folio::Types.service_points.find_by(code: to)&.library&.code
    LibraryHours.new(to).next_business_day(origin_library_next_business_day)
  end
  
  Or are you suggesting I change it all the way up in?
  
  @to =  Settings.libraries[to] ? to : Folio::Types.service_points.find_by(code: to)&.library&.code
  @from = Settings.libraries[from] ? from : Folio::Types.service_points.find_by(code: from)&.library&.code
  
  Because I think this way might be probematic.

jcoyne:
Yes, I was thinking the former.  I only think it needs to be in `destination_library_hours_next_business_day_after_delivery` not in `origin_library_next_business_day`, right?

dnoneill:
That makes sense. yeah.

jcoyne (approved):

PR #2631, opened 2025-01-30
Remove unused routes for feedback

jcoyne:
We only need the post route


app/views/shared/_sul_header.html.erb
```diff
@@ -21,19 +21,17 @@
         <li>
           <%= link_to t('.my_account'), "https://mylibrary.stanford.edu/" %>
         </li>
-        <% unless current_page?(feedback_form_path) %>
```
jcoyne, line 24: This path is for a POST route, which never returns HTML. The condition could never be met.

PR #2646, opened 2025-03-31
update service points to account for patron group

### LINKED ISSUE #2641: Restrict groups from selecting GSB or Law as a pickup location
These groups should not see GSB as a pickup destination:
```
lane-guest
law-alumni
law-guest
sul-guest
sul-contractprograms
sul-short-term
visitor
```

These groups should not see Law as a pickup destination:
```
lane-guest
bus-guest
sul-guest
sul-contractprograms
sul-short-term
visitor
```

@saseestone may add more

dnoneill:
closes #2641


cbeer:
I'm curious what happens if an ineligible user tries to page something that can only be delivered to the one service point (e.g. `BUS-SAL3-PAGE-BU`: https://searchworks.stanford.edu/view/10012542). Do we need a specific error message for them?

dnoneill:
@cbeer Right now this only addresses the drop down (i.e. when the user can choose their service point). I haven't been able to test what happens. My assumption is there is specific user rules on things coming from a specific location but I definitely could be wrong. Right now the test users we have aren't working.

dnoneill:
@cbeer I checked with Sarah. The two examples you pointed out need their own ticket because we need design from Sarah and Darcy. Sarah said this one is okay to go forward since the other two are going to be their own ticket.

app/models/patron_request.rb
```diff
@@ -550,7 +550,9 @@ def default_pickup_service_point
   # Returns default service point codes for all requests
   # @return [Array<String>]
   def default_pickup_service_points_codes
-    Folio::Types.service_points.where(is_default_pickup: true).map(&:code)
+    Folio::Types.service_points.where(is_default_pickup: true).reject do |service_point|
+      service_point.cant_pick_up.include?(patron.patron_group_name)
```
cbeer, line 554: I think this filtering belongs in `#pickup_destinations`, where we're already doing filtering for true visitors. The method (and its comment) suggest to me that it is supposed to get the default service points regardless of the request.

cbeer (changes requested):

app/models/patron_request.rb
cbeer: Ok, I can try a PR with my idea. Maybe there's a reason you needed to introduce the logic here that I'm not getting.

dnoneill: Thanks.

app/models/folio/service_point.rb
```diff
@@ -25,6 +25,10 @@ def self.from_dynamic(json)
           is_default_for_campus: json.dig('details', 'isDefaultForCampus'))
     end
 
+    def cant_pick_up
```
taylor-steve, line 28: What do you think about renaming this (and the settings)? I find the name slightly confusing in the context of usage. I wouldn't know what to expect in return from `destination.cant_pick_up` without looking at the implementation details.

Maybe `excluded_pickup_groups` (`excluded_for_pickup_groups`?) or `excluded_pickup_groups_codes`?

Sub your favorite synonym in: `forbidden`, `disallowed`, `unpermitted`, etc...

dnoneill: Done

dnoneill: Something got messed up. Don't merge

dnoneill: Okay fixed. I forgot to pull down changes after I merged Chris' PR. should be fixed.

PR #2653, opened 2025-04-09
Add chrome arguments to JS driver

taylor-steve:
Testing fix for: https://github.com/teamcapybara/capybara/issues/2800


taylor-steve:
Sure. For reference here's the chromedriver issue: https://issues.chromium.org/issues/402796660

PR #2655, opened 2025-04-16
Update alert text when a Library cardholder tries to page an item for which they are ineligible based on folio circrules.

### LINKED ISSUE #2647: Update alert text when a Library cardholder tries to page an item for which they are ineligible based on folio circrules.
Related to https://github.com/sul-dlss/sul-requests/issues/2641

If the FOLIO circrules restrict a usergroup (e.g. lane-guest) from being able to Request a page/hold on a particular material type/collection (e.g. Business SAL3 PAGE-BU), we show this alert: 

Image

The alert text makes it sound like this is a temporary situation. Instead, we should display a message that is closer to the alert we display for Visitors when they are ineligible to request. Here is that alert:

Image

@dbranchini how about text like:
"This item must be used in the [permanent location Library name] and is not available to request for Library cardholders. If you have a SUNet ID, please log in to complete the request. If you have questions about library access, please contact gsb_librarycirc@stanford.edu."

I know the statement lumps all "Library cardholders" together. And some "Library cardholders" might be able to make the request. But they wouldn't see this alert. It would only display after the circulation rules have been checked and it's certain **this** Library cardholder doesn't have access to the collection. Thoughts?

cbeer:
Fixes #2647 

Screenshot 2025-04-16 at 09 30 51

TODO: 
- [ ] figure out what we're doing with the patron group `desc` field (singular/plural? include "cardholder" or not?)

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -9,15 +9,15 @@ class Ability
   include CanCan::Ability
 
   def self.anonymous
-    @anonymous ||= Ability.new(User.new(name: 'generic', email: 'external-user@example.com'))
+    @anonymous ||= Ability.new(NullUser.new(name: 'generic', email: 'external-user@example.com'))
   end
 
   def self.with_a_library_id
-    @with_a_library_id ||= Ability.new(User.new(library_id: '0000000000'))
+    @with_a_library_id ||= Ability.new(NullUser.new(library_id: '0000000000'))
   end
 
   def self.sso
-    @sso ||= Ability.new(User.new(sunetid: 'generic'))
+    @sso ||= Ability.new(NullUser.new(sunetid: 'generic'))
   end
 
   # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/MethodLength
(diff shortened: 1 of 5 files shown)


dnoneill:
I am not sure if this is a separate ticket or part of this ticket but if you log into http://localhost:3000/patron_requests/new?instance_hrid=L217321&origin_location_code=LANE-SAL3X&step=select with a sul-purchased user the page doesn't show this updated view. It acts like the user can request the item

cbeer:
I think the Lane circ rules currently allow `sul-purchased` patrons. I've been using the lane/business example which has the appropriate circ rules :
https://github.com/sul-dlss/sul-requests/issues/2647#issuecomment-2770905662

dnoneill:
Yeah this issue was spun off of https://github.com/sul-dlss/sul-requests/issues/2641 which purposefully didn't address items with one pick up location because we didn't have language. I my understanding is the rules implemented there would apply to a different ticket specific to single pick up points, but I might have misunderstood.

cbeer:
Sarah confirmed that staff will be updating circ rules, we don't need additional authorization logic in requests.

taylor-steve (approved):

PR #2679, opened 2025-05-20
Use the patron's library id instead of barcode (which is now a proxcard id)

cbeer:


app/models/user.rb
```diff
@@ -35,10 +35,10 @@ def sucard_number=(card_number)
     self.library_id = card_number[/\d{5}(\d+)/, 1]
   end
 
-  # Prefer the patron barcode from the ILS, but fall back to the library ID as-provided
+  # Prefer the patron id from the ILS, but fall back to the library ID as-provided
   # so that the system can function when the ILS is offline
   def barcode
-    patron&.barcode || library_id
+    patron&.blibrary_id || library_id
```
jcoyne, line 41: Typo?

jcoyne (approved):

PR #2688, opened 2025-06-12
add report generator ui

### LINKED ISSUE #2685: Add ability for Request admin to query data
I'm often trying to **count** how many requests have been submitted:
- in a set time period (e.g. 9/1/23-8/31/24, 1/1/25-6/1/25, etc.)
- by type of request (e.g. Hold, Page, Scan, _but also_ Hold + Page, or Hold + Recall + Page)
- from a particular origin library/location (e.g. Education Stacks, or SAL3 Stacks. Even better would be to select from a dropdown based on data in this field, but that would be a nice to have)
- going to a particular destination library (e.g. Green Library, _but also_ Green Library + Business Library)
- (or some combination of two or more of these criteria)

I don't need to see the results of a query onscreen.  It's best if I can build the query and have the results exported to a csv file. 

For example, I asked Chris recently if we could find out how many Page requests have been submitted since the beginning of FY24, by origin library/location and in a way that could help us understand the volume of requests over time. He exported the data in a way that created a count by week, placed it in a [Google sheet](https://docs.google.com/spreadsheets/d/1okrhrCuUiMw-o5DFEXY2veatxbwIcFtXpU7AiRKcvpI/edit?gid=0#gid=0), and created a pivot table. It was perfect to help leadership understand the volume of the current paging service, week to week. 

Chris went above and beyond by adding the count by week, but even being able to export the raw data for a period of time, by request type, and origin library would have helped me (and Chris' time would have been spared.)

dnoneill:
closes #2685 
Screenshot 2025-06-12 at 2 05 51 PM
Screenshot 2025-06-12 at 2 07 52 PM

Example csv when Green and SAL3 are selected
Screenshot 2025-06-13 at 11 41 53 AM

Example csv when Green and SAL3 and hold and recall and page are selected
 
Screenshot 2025-06-13 at 11 41 59 AM

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+###
+#  Controller for creating csv reports on requests
+###
+class ReportsController < ApplicationController
+  def index
+    authorize! :read, :admin
+
+    respond_to do |format|
+      format.html {} # for rails
+      format.csv do
+        send_data(generate_csv_report, filename: "requests_report_#{Time.zone.today.strftime('%Y-%m-%d')}.csv")
+      end
+    end
+    origin_libraries
+    destination_libraries
+  end
+
+  private
+
+  def origin_libraries
+    @origin_libraries ||= PatronRequest.all.map { |elem| Folio::Types.libraries.find_by(code: elem.origin_library_code) }.uniq
+  end
+
+  def destination_libraries
+    @destination_libraries ||= PatronRequest.all.map do |elem|
+      Folio::Types.service_points.find_by(code: elem.service_point_code)
+    end.uniq.compact
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def search_data
+    start_date = Date.parse(params['start_date'])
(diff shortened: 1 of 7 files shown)


saseestone:
@dnoneill Thanks for the screenshots! I'm curious about the values that are showing up under "Request type" in your screenshot.  I assumed they would mirror the values in the ["Type" column of the dashboard](https://requests.stanford.edu/admin). 
image

But it appears that these values might actually be "fulfillment type" in the report.  Can you confirm?  If yes, then I don't know that I understand what "Type" is (and I can't quite figure it out based on the screenshots).  It doesn't really matter, but it would be good for me to understand the differences.  

That's my only question for now.  I think to give more feedback, I'll need to be able to run it myself. Since I'm the only one that is likely to see it, let's deploy it to production as is and then I can run a few reports. If I run across anything that needs tweaking, I can open new issues. 

Thanks! I'm excited.

saseestone:
Ah. Ok. Yes, I'll only need "Type" as in Recall, Hold, Page or Scan.  No need to do anything fancy. :)

taylor-steve (approved):

PR #2728, opened 2025-09-10
Remove old requests cleanup cron

cbeer:
Fixes #2727

PR #2733, opened 2025-09-30
Remove obsolete cleanup task

jcoyne:


lib/tasks/data_removal.rake
```diff
@@ -1,13 +0,0 @@
...
-
-namespace :data_removal do
-  desc 'Remove old requests'
-  task remove_old_requests: :environment do
-    t = ActiveSupport::Duration.build(Settings.data_cleanup.age).ago
-    raise 'Date is too recent' if t > 1.month.ago
-
-    Request.obsolete(t).delete_all
-
-    PatronRequest.obsolete(t).delete_all
```
corylown, line 11: `PatronRequest` still exists. Do we not need to clean up that data anymore? If we don't then this setting can also be removed: https://github.com/sul-dlss/sul-requests/blob/4e523a6cbb09f970b854cb2da8fdeb6175cd2716/config/settings.yml#L96

jcoyne: I don't know. I was just following on from what @cbeer had created here: https://github.com/sul-dlss/sul-requests/pull/2728. I don't think we want a rake task that isn't called.

PR #2750, opened 2025-10-31
add filters and download to requests dashboard

### LINKED ISSUE #2687: Allow admins to filter the requests dashboard and export data
Extracted from #2685

From Sarah:

> I'm often using the same criteria but instead of a count, I need a list of individual requests. For these lists, the report should include all the request elements that appear on the Request dashboard (i.e. Type, Origin, Destination, Title, Requester, Requested on, and Status). However, I need to be able to filter by one or more of the criteria above (set time period, type, particular origin library/location, etc.) AND I need to export it to a csv file.

dnoneill:
closes #2687 

Screenshot 2025-10-30 at 6 19 38 PM
Screenshot 2025-10-30 at 6 19 31 PM


app/controllers/application_controller.rb
```diff
@@ -45,4 +45,28 @@ def redirect_after_action
   def post_action_redirect_url
     params[:referrer].presence || root_url
   end
+
+  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def search_patron_request
```
jcoyne: why is this in ApplicationController instead of ReportsControllers?

jcoyne (changes requested):

PR #2778, opened 2025-12-04
Port over folio.config_set from searchworks.

cbeer:


jcoyne:
Would you mind saying why we want this to be ported into sul-requests?

jcoyne:
Should we remove the old (non-config set) configs then?

cbeer:
Didn't I? Or is there some other set of configurations I'm not thinking of?

Screenshot 2025-12-04 at 09 07 20

jcoyne:
Ooops, I missed that. I it showed me a rename, but I was looking for deletes.

jcoyne (approved):

PR #2805, opened 2026-01-13
Update paging scheduling to estimate delivery times based on the actual physical processes

cbeer:
We've historically estimated delivery times based on point-to-point schedules (e.g. GREEN to ART takes 1 day). With expanded paging, some of these estimates aren't quite right (especially to/from MARINE-BIO). More accurate modeling gives us more accurate estimates (🤞 ) and makes it easier to update our estimates as the physical process schedules change.


config/settings.yml
taylor-steve: I think we can get rid of `worst_case_paging_padding` as well.

PR #2863, opened 2026-02-03
Add lookbook.

cbeer:


config/routes.rb
```diff
@@ -66,4 +66,9 @@
   end
   resource :feedback_form, path: 'feedback', only: %I[new create]
   get 'feedback' => 'feedback_forms#new'
+
+
+  Rails.application.routes.draw do
```
dnoneill: Do we want this on prod as well? I think it is fine if we do I want to make sure that is a correct assumption

spec/component_previews/accordion_step_component_preview.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AccordionStepComponentPreview < ViewComponent::Preview
```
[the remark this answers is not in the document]
dnoneill: Nevermind this doesn't work

config/routes.rb
cbeer: I think we do, for the same reasons we did this in searchworks... it'd be nice if Darcy et al could see it in a deployed environment, and it's harmless enough 🤷

PR #2875, opened 2026-02-04
Fix various things with the component library.

cbeer:


app/assets/stylesheets/application.css
```diff
@@ -319,3 +319,8 @@ label.btn-close {
 #pre-footer {
   background-color: var(--stanford-stone-light);
 }
+
+/* Restore the upstream bootstrap behavior that the component library broke. */
+.alert-link {
```
jcoyne: Can you explain what behavior the "Style library" broke?  I'm pretty sure @dbranchini specified the weight for links in alerts to be 400 (as is shown here https://sul-dlss.github.io/component-library/alerts/). If you wanted to change it, could you set `--sul-link-font-weight: 700;`?

Do we want to make this kind of change in the style library or just here?

As an aside, it's not called the "component library" anymore.

cbeer: I guess I don't understand why the style library forces any and all links to be font-weight 400. That seems like a fine default, but Darcy wants these alert links to be bold (and `.alert-link` is the upstream way to accomplish that). I don't get why the style library is so heavy-handed, but I'm not interested in exploring that right now.

app/assets/stylesheets/application.css
```diff
@@ -310,3 +316,11 @@ label.btn-close {
   text-wrap: nowrap;
 }
 
+#pre-footer {
+  background-color: var(--stanford-stone-light);
```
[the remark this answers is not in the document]
cbeer: Darcy explicitly requested a different background.

PR #2887, opened 2026-02-05
Refactor Aeon request model

taylor-steve:
Helps set the pattern for the appointment model

### DIFF
--- a/app/views/requests/index.html.erb
+++ b/app/views/requests/index.html.erb
@@ -3,11 +3,7 @@
 <% if current_user&.aeon&.requests %>
   <% current_user.aeon.requests.each do |request| %>
     <%# This might need to change but we still have to figure out  %>
-    <% if request.data['appointment'] %>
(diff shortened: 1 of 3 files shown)


app/views/requests/index.html.erb
```diff
@@ -3,12 +3,8 @@
 <% if current_user&.aeon&.requests %>
   <% current_user.aeon.requests.each do |request| %>
     <%# This might need to change but we still have to figure out  %>
-    <% if request.data['appointment'] %>
-      <%= render Aeon::AppointmentRequestComponent.new(request:) %>
-    <% else %>
-      <%= render Aeon::DigitizedRequestComponent.new(request:) %>
-    <% end %>
+    <%= render Aeon::DigitizedRequestComponent.new(request:) %>
```
cbeer: Was this change intentional?

cbeer: I'm not sure I understand where the view is going, but I think the model changes are good 🤷‍♂️

Issue #2838, opened 2026-02-02
New request (via Finding Aid)


dbranchini:
### Context

Special collections are frequently coupled with finding aids and users need to request those items via the finding aid. Archival Collections at Stanford and Online Archive of California are both finding aids. These materials can be organized by series, boxes, and/or folders, and the user needs to select contents as part of their request. 

### Design notes

PR #2894, opened 2026-02-06
Handle requests for EAD XML and show basic display of data

marlo-longley:
This is pretty demo-y. 
Visit http://localhost:3000/archives_requests?value=https%3A%2F%2Farchives.stanford.edu%2Fdownload%2Fsc0097.xml to see the show page.

This code also works with an OAC finding aid:
http://localhost:3000/archives_requests?value=https%3A%2F%2Fcinco-prd.s3.amazonaws.com%2Fmedia%2Fead%2Fsc0097_FGvsRNd.xml

I disabled a lot of Rubocops etc just to show proof of concept.

~~`ead_remove_namespace` XSLT is taken wholesale from Stanford Arclight. Once this is in production we could remove that logic and the `without_namespace` param from Stanford Arclight.~~
- Nokogiri's `remove_namespaces` works completely fine for our situation. No XSLT needed at all!
- I believe we could still remove that logic from Stanford Arclight in the future.

A lot of the logic in `ead_client` is adapted from Aeon's XSLT which is linked in the related ticket, https://github.com/sul-dlss/sul-requests/issues/2838

Screenshot 2026-02-06 at 3 45 10 PM

### DIFF
--- a/app/services/ead_client.rb
+++ b/app/services/ead_client.rb
@@ -11,16 +11,12 @@ class EadClient
   def self.fetch(url)
     response = Faraday.get(url)
 
-    raise "Failed to fetch EAD XML: HTTP #{response.code}" unless response.success?
+    raise EadClient::Error, "Failed to fetch EAD XML: HTTP #{response.code}" unless response.success?
 
     doc = Nokogiri::XML(response.body)
     doc.remove_namespaces!
 
     Ead::Document.new(doc, url: url)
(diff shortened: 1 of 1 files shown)


cbeer:
I've made some suggestions over in #2901 (I needed to puzzle out if some things were even a good idea).. I'm happy for you to either merge the PR into this or pick+choose commits that make sense and we can go from there.

Issue #2845, opened 2026-02-03
Create a component for Aeon Digitization


taylor-steve:
`Digitization ready`/`Digitization pending` in the designs. I believe all the needed values are in the user's requests in the Aeon API. Assumed data from the API can be used and whoever picks this up and whoever picks up https://github.com/sul-dlss/sul-requests/issues/2842 can meet coordinate/meet in the middle to decide what shape makes sense for that data.

PR #2896, opened 2026-02-06
update digitized component to include most fields, header still needs…

dnoneill:
… discussion

related to #2845 

image


app/models/aeon/request.rb
```diff
@@ -3,38 +3,40 @@
...
-        title: dyn['itemTitle'],
+        digitized_request: dyn['shippingOption'] == 'Electronic Delivery',
+        location: dyn['xlocation'],
+        item_metadata: {
+          'title' => dyn['itemTitle'], 'pages' => dyn['itemInfo5'],
+          'date' => dyn['itemDate'], 'format' => dyn['format'],
+          'document_type' => dyn['documentType'], 'aeon_link' => dyn['itemInfo1'],
+          'volume' => dyn['itemVolume'], 'call_number' => dyn['callNumber'],
+          'author' => dyn['itemAuthor']
+        },
```
taylor-steve: Could you speak to what you are trying to do here? Is `item_metadata` just a convenience? I'm concerned about deviating to much from a 1:1 with Aeon's schema until we understand updates better.

I think logic like `dyn['shippingOption'] == 'Electronic Delivery'` is more clearly expressed in a method. At least it's helping me keep track of what is Aeon and what is us at this point.

Did you mean to change `location` to `xlocation`?

app/components/aeon/digitized_request_component.html.erb
```diff
@@ -1,17 +1,28 @@
...
   </div>
   <div class="card-body">
-    <h2 class="h3"><%= title %></h2>
+    <h2 class="h3"><%= item_metadata['title'] %></h2>
     <div class="d-flex">
-      <div class="me-4"><replace_with_icon><%= document_type %></div>
-      <div class="me-4">Call number: <%= call_number %></div>
-      <div><%= aeon_link %></div>
+      <div class="me-4">
+        <replace_with_icon><%= item_metadata['document_type'] %><% if item_metadata['date'] %> (<%= item_metadata['date'] %>)<% end %>
```
taylor-steve: Maybe this is my own personal pet peeve but it'd be great if we could avoid digging for data in component templates.

Issue #2846, opened 2026-02-03
Handle requests with appointments


taylor-steve:
This builds on https://github.com/sul-dlss/sul-requests/issues/2845 to implement the designs for 'Reading room appointment'. There are almost no new design elements. This issue is primarily about pulling in user request appointment information from Aeon.

Aeon appears to expand the appointment in the request. Maybe whoever picked up https://github.com/sul-dlss/sul-requests/issues/2842 will have solved this with some OpenAPI/swagger magic.

PR #2897, opened 2026-02-06
Add a method to get a user's Aeon appointments

taylor-steve:
Part of #2846


dnoneill:
What are we going to need this for? All the data is already in requests_for?

taylor-steve:
🤷 It's a thing that can happen that we don't have 100% control over. I'm not in a hurry to get this in so time will tell if we actually need it or not.

PR #2940, opened 2026-02-12
Submit an archives request to Aeon

marlo-longley:
I realize this doesn't totally align with the Figma yet - button text, Accordion flow, I am displaying certain info in the last Accordion panel before submit just for our own debugging purposes for now.

Also note, I combined multiple volume requests into one here for the initial proof of concept. I also added extra data to the volume info -- namely the containing folder. So now instead of `Box 1`, `Box 1`, you would see the disambiguated `Folder 1: Box 1`, `Folder 2, Box 1`. Again this is just for now, to stand this up, and I'm interested in discussing volume data as a group.

Screenshot 2026-02-12 at 3 36 20 PM
Screenshot 2026-02-12 at 3 36 27 PM
Screenshot 2026-02-12 at 3 36 36 PM

### DIFF
--- a/app/controllers/archives_requests_controller.rb
+++ b/app/controllers/archives_requests_controller.rb
@@ -20,8 +20,7 @@ class ArchivesRequestsController < ApplicationController
     # Fetch EAD data to get the actual collection information
     @ead = EadClient.fetch(params[:ead_url])
 
-    username = current_user.sunetid.present? ? "#{current_user.sunetid}@stanford.edu" : current_user.email
-    # volumes = params[:volumes]&.reject(&:blank?)
+    username = current_user.email_address
     volumes = params[:volumes]&.reject(&:blank?)&.map { |json_str| JSON.parse(json_str) }
 
     # Submit each volume request and collect results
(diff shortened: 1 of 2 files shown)


app/controllers/archives_requests_controller.rb
```diff
@@ -7,8 +7,45 @@ class ArchivesRequestsController < ApplicationController
...
     @ead = EadClient.fetch(ead_url_param)
+    @ead_url = ead_url_param
   end
 
+  # This is the action triggered by the form submission to create an Aeon request.
+  def submit_to_aeon
+    # Fetch EAD data to get the actual collection information
+    @ead = EadClient.fetch(params[:ead_url])
+    
+    # Join selected volumes into a comma-separated string
```
cbeer: I thought the current Aeon behavior was to create a new, separate request for each selected volume?

marlo-longley: Yes I combined multiple volume requests into one here for the initial proof of concept for connecting with their API. I didn't know if we needed to discuss this further or if there any changes being made to the way volumes are handled.

PR #2953, opened 2026-02-17
Update Aeon digitization request draft status

taylor-steve:
Digitization requests initially stay in the draft transaction queue. We must check the photoduplication queue status in these cases.

### DIFF
--- a/app/models/aeon/request.rb
+++ b/app/models/aeon/request.rb
@@ -27,6 +27,7 @@ module Aeon
         location: dyn['location'],
         pages: dyn['itemInfo5'],
         photoduplication_status: dyn['photoduplicationStatus'],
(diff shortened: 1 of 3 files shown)


app/models/aeon/request.rb
```diff
@@ -88,6 +90,8 @@ def canceled?
     end
 
     def draft?
+      return false if digital? && !photoduplication_queue&.draft?
+
       transaction_queue&.draft?
```
cbeer: Can we rewrite this condition to be clearer? E.g. (i'm not sure if this is right:)

```suggestion
      if digital?
        photoduplication_queue&.draft?
      else
        transaction_queue&.draft?
      end  
```

PR #2971, opened 2026-02-18
Wire up Aeon's request routes endpoint to see if it's useful for any …

cbeer:
…of our state management.

I'm not convinced 🤷 

Part of https://github.com/sul-dlss/sul-requests/issues/2993

### DIFF
--- a/app/models/aeon/possible_routes.rb
+++ b/app/models/aeon/possible_routes.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Aeon
+  # Wrap the request routes response for a transaction
+  class PossibleRoutes
+    def self.from_dynamic(dyn)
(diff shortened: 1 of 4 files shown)


cbeer:
Per Steve: 
> One mystery I mentioned during stand-up resolved? Given a draft Aeon request that is only missing intended use (itemInfo3), an update to that field via the API does NOT change the request's queue state, it remains in draft ( Awaiting User Review ). The built-in workflow logic for that Aeon "Submit request" button must live in their webapp, not the API.
>
> With the same request scenario, but this time leaving itemInfo3 blank, the request can be moved out of draft into the next queue (Awaiting Future Request Processing) via the API despite being "incomplete". It seems to be a real active request in the Aeon interface.
>
> Basically the rules are fake and we can do what we want...until we discover a situation where that's not true. That means we should sit on https://github.com/sul-dlss/sul-requests/pull/2971 until a situation like that pops up.

PR #2994, opened 2026-02-19
Improve logic and display for archives request

marlo-longley:


app/views/archives_requests/_series_items.html.erb
```diff
@@ -0,0 +1,130 @@
...
+    else
+      # Subseries - use object_id for uniqueness
+      key = [:subseries, item.object_id]
+      next if seen_keys.include?(key)
+      seen_keys.add(key)
+      
+      display_groups << { type: :subseries, data: item }
+    end
+  end
+%>
```
cbeer: Can we pull this out into a helper method or into `Ead::Document`?

app/assets/stylesheets/application.css
```diff
@@ -437,4 +437,12 @@ label.btn-close {
...
     display: none;
   }
+}
+
+/* EAD Item selector styles */
+.series-title {
+  font-size: 1.125rem; 
+}
+.form-check-label {
+  color: var(--bs-link-color);
```
cbeer: This should probably be scoped to the item picker, right? I'm seeing it applied other places, like:

Image

app/models/ead/display_group.rb
```diff
@@ -0,0 +1,66 @@
...
+    # This method can build 3 types of groups:
+    # 1. :item_container: multiple items sharing the same top container (e.g., Box 9).
+    # "container" is identified in Ead::Document by the presence of <container> elements and refers to physical containers.
+    # 2. :individual_item: These are leaf nodes in the EAD hierarchy that don't have a physical container.
+    # 3. :subseries: This is an intellectual grouping in the EAD not tied to physical containers. Each subseries gets its own group.
+    def build
+      return [] if @contents.empty?
+
+      # Separate contents by type
+      contents_by_type = @contents.group_by { |item| item.is_a?(Ead::Document::Item) ? :item : :subseries }
```
cbeer: Do you ever use `contents_by_type[:subseries]`?

app/models/ead/display_group.rb
```diff
@@ -0,0 +1,66 @@
...
+      leaf_items = contents_by_type[:item] || []
```
cbeer: or is this just:

```suggestion
      leaf_items = @contents.select { |item| item.is_a?(Ead::Document::Item) }
```

app/models/ead/display_group.rb
```diff
@@ -0,0 +1,66 @@
...
+
+          seen_keys.add(key)
+
+          display_groups << { type: :subseries, data: c }
+        end
+      end
+      # rubocop:enable Metrics/PerceivedComplexity, Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity
+
+      display_groups
+    end
```
cbeer: Is something like this equivalent (it seemed to be for SC0097):

```suggestion
    def build
      return [] if @contents.empty?

      # Build display groups preserving original order
      @contents.group_by { |c| c.try(:top_container) || c.object_id }.each_value.map do |group|
        c = group.first

        if c.try(:top_container).present?
          { type: :item_container, name: container_name, contents: group }
        elsif c.is_a?(Ead::Document::Item)
          { type: :individual_item, item: c }
        else
          { type: :subseries, data: c }
        end
      end
    end
```

I think `group_by` preserves the existing order, and your clever `object_id` thing should keep those non-boxes ordered too, I think?

PR #3010, opened 2026-02-20
Extract Ead::Document::Node

cbeer:
Is it any clearer? 🤷


marlo-longley (approved):

marlo-longley:
This is definitely cleaner. Thanks @cbeer. Better to work with class Node than have plain hashes mixed in.

PR #3071, opened 2026-02-25
add character counter for additional info

### LINKED ISSUE #3048: Guard against Aeon fields with character limits
Some of the designs show a warning about the character limit:

Image

We should enforce this on the user side and we should consider guarding against this in our Aeon API calls, as they will `400` error if exceeded.

The database tables page Chris found might give some useful hints:

https://support.atlas-sys.com/hc/en-us/articles/360011920013-Aeon-Database-Tables

dnoneill:
closes #3048 
https://github.com/user-attachments/assets/68d07079-97a2-434e-bca6-8d1221af4dc7

### DIFF
--- a/app/components/aeon/digitization_form_item_component.html.erb
+++ b/app/components/aeon/digitization_form_item_component.html.erb
@@ -34,8 +34,11 @@
 
         <div class="mb-3">
           <%= f.label :additional_information, class: 'form-label fw-semibold' %>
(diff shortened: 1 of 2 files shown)


taylor-steve:
I think you may need to rebase this, things might have changed with previous merged PRs. We might need a test for this?

Also there are more places where we would need this:
https://github.com/search?q=repo%3Asul-dlss%2Fsul-requests%20255%20limit&type=code

PR #3098, opened 2026-02-26
Refactor item selection to use the same stimulus controller for all types of item selection

marlo-longley:
... and use the same digitization + reading room components in all Aeon request flows.

TODO:

- [x] replace the current available/unavailable slots with generic code
- [x] replace the current scan slots with generic code
- [x] update archives_requests to use itemselector
- [x] replace the new FOLIO -> Aeon requests with generic code 
- [x] rename itemselector -> item-selector

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -308,7 +308,7 @@ label.btn-close {
   margin-bottom: 0;
 }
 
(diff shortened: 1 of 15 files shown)


marlo-longley:
@cbeer if I am out and you want to make changes please feel free

Issue #3132, opened 2026-03-02
Single selected box with appointment


dnoneill:
<img width="468" height="275" alt="Image" src="https://github.com/user-attachments/assets/88f8ad3f-c05a-411a-a3d8-b376c442dd6f" />

taylor-steve:
Might intersect with https://github.com/sul-dlss/sul-requests/issues/2848

PR #3155, opened 2026-03-03
Use CSS to change the appointment form display if there's only one item selected

cbeer:
Part of #3132

https://github.com/user-attachments/assets/f9f337b3-eea0-4de0-bfcc-7f58ba1d8dab

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -547,6 +547,7 @@ label.btn-close {
   }
 }
 
+#reading:has(.appointments-item:only-child),
 #pickup:has(.appointments-item:only-child) {
   .hide-when-1-item {
     display: none;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/application.css
```diff
@@ -546,3 +546,55 @@ label.btn-close {
...
+#reading:has(.appointments-item:only-child),
+#pickup:has(.appointments-item:only-child) {
+  .hide-when-1-item {
+    display: none;
+  }
+}
+
+.appointments-container .appointments-item:not(:only-child) {
+  /* .border */
+  border: var(--bs-border-width) var(--bs-border-style) var(--bs-border-color) !important;
```
dnoneill: Is there a reason the border and bg-light were moved from the form?

cbeer: I don't think there's any way to remove those styles when we're only displaying a single item, is there?

PR #3158, opened 2026-03-03
Consolidate archives requests into patron requests

### LINKED ISSUE #3040: Change SW Aeon request submit and evaluate feasibility of merging Aeon request approaches
SearchWorks Aeon requests are currently getting submitted here, which was temporary, to get us going:
https://github.com/sul-dlss/sul-requests/blob/ec3421cccf893752c110091b8b7dbb1a26093a0a/app/controllers/patron_requests_controller.rb#L38

That should change.

https://github.com/sul-dlss/sul-requests/pull/2965 added `SubmitAeonPatronRequestJob`, which is a different mechanism than the finding aid/EAD submission process. Is there commonality here or do they need to remain distinct?

cbeer:
Fixes #3040

### DIFF
--- a/app/views/patron_requests/_ead_series_contents.html.erb
+++ b/app/views/patron_requests/_ead_series_contents.html.erb
@@ -23,7 +23,7 @@
       <% end %>
       <%# Checkbox to select entire container %>
       <%- checked = f.object.selected_items.any? { |item| series_title == item[:series] && container_name == item[:subseries] } %>
-      <%= f.check_box :volumes, 
+      <%= f.check_box :barcodes,
             { multiple: true, class: 'form-check-input mt-1', id: checkbox_id, checked: checked,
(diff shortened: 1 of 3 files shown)


app/views/patron_requests/_ead_series_contents.html.erb
```diff
@@ -23,13 +23,13 @@
       <% end %>
       <%# Checkbox to select entire container %>
       <%- checked = f.object.selected_items.any? { |item| series_title == item[:series] && container_name == item[:subseries] } %>
-      <%= f.check_box :volumes, 
+      <%= f.check_box :barcodes,
```
marlo-longley: Do we have barcodes in Finding Aids?

cbeer: I looked at renaming this `item_id` for EAD + FOLIO holdings, but it wasn't going to be pretty (either a data migration, or lots of "is it an item id or a barcode" scattered around)... and maybe some day we will have barcodes in EAD 😁

PR #3184, opened 2026-03-04
Add placeholder text to appointment select

taylor-steve:
Screenshot 2026-03-04 at 2 58 50 PM

Final language to come later from Darcy.


cbeer:
Should we ship this now and refine the language later?

PR #3186, opened 2026-03-04
integrating terms directly into request type

### LINKED ISSUE #3059: Move "Terms" step before "Select item(s)" for Digitization
@dbranchini would like the user to accept the terms prior to selecting items for digitization requests.

Instead of adding Terms as a separate step, we'd like to move it to Step 1 for digitization requests. 

[Browser view - interactive](https://www.figma.com/proto/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?page-id=1822%3A12044&node-id=4089-31730&viewport=3744%2C-18323%2C0.28&t=2VlzUl6J6JgGlbg0-9&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=4089%3A31730&show-proto-sidebar=1)

Design specs

- [State - before user agrees](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=4089-31580&t=D5V4OATKrPryJMdN-4)
- [State - after user agrees](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=4089-31669&t=D5V4OATKrPryJMdN-4)

hudajkhan:
Closes #3059 

Patron Requests (current)
Screenshot 2026-03-04 at 4 51 04 PM
--
Screenshot 2026-03-04 at 4 51 43 PM

Patron Requests (after changes)

Screenshot 2026-03-04 at 4 49 44 PM
--
Screenshot 2026-03-04 at 4 49 54 PM
--
Screenshot 2026-03-04 at 4 50 01 PM

---

Archives Requests page (current)

Screenshot 2026-03-04 at 4 47 22 PM

---

Archives Requests page (after changes)

Screenshot 2026-03-04 at 4 47 49 PM
--
Screenshot 2026-03-04 at 4 48 01 PM
--
Screenshot 2026-03-04 at 4 48 16 PM

### DIFF
--- a/spec/features/ead_patron_request_spec.rb
+++ b/spec/features/ead_patron_request_spec.rb
@@ -105,7 +105,7 @@ RSpec.describe 'Requesting an item from an EAD', :js do
       fill_in 'Requested pages', with: 'Pages 1-10'
       fill_in 'Additional information', with: 'Testing only'
 
-      click_button 'Submit to Aeon'
+      click_button 'Submit request'
 
       expect(page).to have_content('We received your digitization request')
 


app/javascript/controllers/patron_request_controller.js
```diff
@@ -53,6 +53,17 @@ export default class extends Controller {
 
     this.element.dataset.accordionFormTypeValue = requestType;
     this.element.dataset.itemSelectorItemLimitValue = this.typeValue != 'aeon' && requestType == 'scan' ? 1 : -1;
+
+    // if this request type needs to display terms
+    if(this.hasTermsTarget) {
+      if(event.target.dataset.terms && event.target.dataset.terms == 'true') {
```
taylor-steve: This isn't a big deal but would it be worth doing:

```
      if (requestType == 'scan') {
        this.termsTarget.classList.remove('d-none')
        this.enableRequiredInputs(this.termsTarget);
      } else {
        this.termsTarget.classList.add('d-none')
        this.disableRequiredInputs(this.termsTarget)
      }
```

We're already making decisions in this function based on `requestType`. Then you could drop the terms data attributes? Up to you, I don't mind it either way really.

taylor-steve (approved):

app/javascript/controllers/patron_request_controller.js
hudajkhan: I would have to check how this intersects with the regular scan/pickup flow - also those pages don't have the terms target available so it might throw an error in that case

PR #3282, opened 2026-03-13
Redirect FOLIO requests for things with finding aids to the EAD

cbeer:


cbeer:
Ok, let's give `allow_other_host` a try. I was settling for the archives site because getting from the finding aid to the EAD XML wasn't straightforward.

PR #3304, opened 2026-03-18
Wait for all submitted Aeon requests to be available for display

taylor-steve:
Nobody asked for this but it's been bugging me in the demos. Opinions welcome 😄 

This opts for polling until we see the Aeon request count we expect for a given patron request. I found this preferable to updating over time. This is because Aeon requests start as "Awaiting Request Processing" (queue 8), but we send a subsequent API call to move them to "Awaiting User Review" (queue 5) if the request is a draft. When displaying in real time, the user sees each draft request first as a real "submitted" request, then sees it move to draft.

I added a loading icon and placeholder text: 
Screenshot 2026-03-18 at 8 11 01 AM

### DIFF
--- a/app/javascript/controllers/aeon_confirmation_poll_controller.js
+++ b/app/javascript/controllers/aeon_confirmation_poll_controller.js
@@ -1,0 +1,35 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Polls the confirmation page turbo frame until all expected Aeon requests
+// have appeared, or max attempts are reached.
+export default class extends Controller {
+  static values = {
+    attempts: Number,
+    interval: { type: Number, default: 1000 },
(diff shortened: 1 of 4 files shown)


app/javascript/controllers/aeon_confirmation_poll_controller.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  static targets = ["frame"]
+
+  connect() {
+    this.attempts = 0
+
+    if (this.complete()) return
+
+    this.frameTarget.addEventListener("turbo:frame-load", this.onFrameLoad)
```
cbeer: If the target idea below doesn't work, maybe this could be `data-action` values on the frame so you don't need to do as much manual bookkeeping?

Issue #3313, opened 2026-03-19
Disable accordion behavior for single-item digitization display


cbeer:
<img width="718" height="720" alt="Image" src="https://github.com/user-attachments/assets/8900efb5-6536-49e7-b787-61d28a664055" />

If you've only selected one item to digitize, we've hidden the accordion chrome, but the behavior is still active.

taylor-steve:
Update as promised. The only remaining issue is when editing a digitization request. There are no more JS accordion state issues as far as I know.

We've got three approaches so far:
* Passing in a new param to tell the component to change its behavior: https://github.com/sul-dlss/sul-requests/pull/3407
* Inferring when the component should change its behavior based on our pattern of usage: https://github.com/sul-dlss/sul-requests/pull/3437
* They are different things, split them into two components: https://github.com/sul-dlss/sul-requests/pull/3412

This is only used in two places (three calls):
https://github.com/search?q=repo%3Asul-dlss%2Fsul-requests%20%22DigitizationFormItemComponent.new%22&type=code

It'd be nice if this was a Ruby solution so [save for later](https://github.com/sul-dlss/sul-requests/pull/3455) could use the same approach to avoid rendering the actions in this circumstance.

PR #3321, opened 2026-03-19
Disable digitization accordion button if only one item is selected

marlo-longley:
Closes https://github.com/sul-dlss/sul-requests/issues/3313


dnoneill (changes requested):
I think this can be done with css. Add `pointer-events: none;` to https://github.com/sul-dlss/sul-requests/blob/f101e4b93eaf8d3812ffcca2c7b5e554f94ff260/app/assets/stylesheets/application.css#L587

PR #3338, opened 2026-03-20
Hook up a simple version of removing items from an appointment

### LINKED ISSUE #3316: Remove item from appointment
A user should be able to remove items from an appointment (from the [appointments page](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=4967-24167&t=5lVdT2NEKDnPFFkn-4)).

**Design notes:**
- If a user removes an item, it is moved back into a "draft" status and should appear in the list of drafts to the right. 

----
Image

----

**Related tickets:**
- #3314 
- #3315 
- #3317

cbeer:
Fixes #3316

It'd be nice if this PR had better tests, but all the mocking required to handle the state change seems problematic.

### DIFF
--- a/app/components/aeon/request_group_item_component.html.erb
+++ b/app/components/aeon/request_group_item_component.html.erb
@@ -2,7 +2,18 @@
   <div class="d-flex flex-row align-items-start justify-content-between gap-5">
     <div class="d-flex flex-row align-items-start justify-content-between gap-4 flex-grow-1">
       <div class="d-flex flex-column">
-        <%= render Aeon::RequestItemIdentifierComponent.new(request:) %>
+        <div class="d-flex justify-content-start align-items-center">
+          <%= render Aeon::RequestItemIdentifierComponent.new(request:) %>
+
+          <% if remove_from_appointment? %>
+            <%= form_with(url: redraft_aeon_request_path(request)) do |f| %>
+              <%= tag.button type: :submit, class: 'btn btn-link su-underline' do %>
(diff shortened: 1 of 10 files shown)


dnoneill:
Isn't there a way with turbo to do only the card reload? i.e. https://github.com/sul-dlss/sul-requests/blob/c10009b184b4994fd22d5cbde399bb98b629d0ea/app/views/aeon_appointments/create.turbo_stream.erb#L4-L20

taylor-steve:
Works well for multiple. Should it work with the single case at this point?

E.g., for http://localhost:3000/patron_requests/new?instance_hrid=4103002&origin_location_code=SPEC-SAL3-U-ARCHIVES

Before:
Screenshot 2026-04-30 at 9 14 15 AM

After clicking remove:
Screenshot 2026-04-30 at 9 14 29 AM

It does the work, it's only the display.

cbeer:
ok, try now. I had #3562 on my mind when I was rebasing.

taylor-steve:
Screenshot 2026-04-30 at 2 59 14 PM

Kind of a challenge. What do we want here? The "No items have been requested..." alert + the item limit updated? Do we need to replace the whole appointment in these cases?

cbeer:
I hope we don't need to do the whole appointment. #3573 should handle hiding + showing the alert, and maybe with #3574 we can do a clever turbo update for the limit.

taylor-steve (approved):

Issue #3317, opened 2026-03-19
Add items to appointment


dbranchini:
A user should be able to add items to an appointment (from the [appointments page](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=4967-24167&t=5lVdT2NEKDnPFFkn-4)).

**Designs**
- [With current assigned/scheduled items accordion collapsed](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=4967-23491&t=ElNAO6btjoBf1QeN-4)
- [With current assigned/scheduled items accordion expanded](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=4967-23576&t=ElNAO6btjoBf1QeN-4)

**Browser view** 
- [Add items](https://www.figma.com/proto/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?page-id=1822%3A28297&node-id=4967-23491&p=f&viewport=4436%2C-6057%2C0.39&t=DQCH14RWTwJknp1T-9&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=4967%3A23491&show-proto-sidebar=1) (if you click on the screen, interactive elements for prototype will flash in blue)

**Design notes:**
- If a user selects, "Add items," then a modal is displayed with a closed accordion that shows items already assigned to the appointment and a section for items that are eligible to add to the appointment. Items that are eligible are reading room requests without an appointment AND associated with the appointment's reading room.
- If a user adds an item, it is moved out of "draft" status and should appear in the appointment accordion and in the appointment on the appointments page (once the user closes the modal).

----
<img width="400" alt="Image" src="https://github.com/user-attachments/assets/54bee817-a5b0-4762-8e0e-e2191cc880d4" />

----

**Related tickets:**
- #3314 
- #3315 
- #3316

PR #3342, opened 2026-03-20
scaffold out add items modal, no backend hookup

dnoneill:
part of #3317 

Screenshot 2026-04-09 at 4 26 14 PM
Screenshot 2026-04-09 at 4 26 02 PM


taylor-steve (commented):
How do you want to handle the scaffolding? If this is to get the markup in, I wonder if we can get it a bit closer to the design first. If you scaffolded this because of limited time at the end of the work-cycle, I'd be okay waiting to see the whole thing.

taylor-steve (approved):

PR #3344, opened 2026-03-20
"no" turns out to be truthy

cbeer:
(in reality, we're changing for 'yes', but....)


app/components/aeon/digitization_form_item_component.html.erb
```diff
@@ -26,8 +26,8 @@
               <%= f.label :for_publication_yes, 'Yes' %>
             </span>
 
-            <%= f.radio_button :for_publication, 'no' %>
-            <%= f.label :for_publication_no, 'No' %>
+            <%= f.radio_button :for_publication, 'falase' %>
```
cbeer:
```suggestion
            <%= f.radio_button :for_publication, 'false' %>
```

taylor-steve (approved):

PR #3367, opened 2026-03-24
put equal gap between elements bracketing OR

dnoneill:
Before:
Screenshot 2026-03-24 at 2 16 54 PM
After:

Screenshot 2026-04-01 at 11 36 37 AM


cbeer:
It looks like we lost some spacing on the multi-item picker.

Before:
Screenshot 2026-03-24 at 14 17 48

After:
Screenshot 2026-03-24 at 14 17 55

It also seems like maybe we lost some padding between the appointment time and the item count?

Bootstrap seems happy with the top-aligned dropdown arrows. Is this something that's just bugging you, or a chance Darcy or DLUX requested?

taylor-steve (approved):

PR #3378, opened 2026-03-27
Add the reading room directions to the requests confirmation screen

taylor-steve:
Screenshot 2026-03-27 at 10 34 08 AM

### DIFF
--- a/app/models/aeon/request_grouping.rb
+++ b/app/models/aeon/request_grouping.rb
@@ -9,7 +9,7 @@ module Aeon
 
     delegate :each, to: :requests
 
-    delegate :appointment?, :submitted?, :appointment, :base_callnumber, :call_number, :date, :digital?,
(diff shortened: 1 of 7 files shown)


app/models/aeon/request_grouping.rb
```diff
@@ -38,8 +40,10 @@ def status_request
       requests.find { |r| !r.scan_delivered? } || first
     end
 
-    def reading_room_name
-      appointment&.reading_room&.name
+    private
+
+    def request_with_appointment
```
dnoneill: I think this is unnecessary. I think it should just be delegate :reading_room to: :first in line 12. reading_room is defined for drafts as well is my understanding https://github.com/sul-dlss/sul-requests/blob/794978ee1a9b7629507da4155266d9ebc8c2ae5f/app/models/aeon/request.rb#L148-L152

dnoneill (approved):

PR #3407, opened 2026-04-02
Disable accordion button for single pre-rendered digitization items

### LINKED ISSUE #3313: Disable accordion behavior for single-item digitization display

marlo-longley:
(again) closes https://github.com/sul-dlss/sul-requests/issues/3313

I wasn't sure the right place to test this.

Herb required that I set button attributes in the .rb file because of `disabled`


cbeer:
Merged effectively the same approach in #3437. That PR lacked  the javascript change in https://github.com/sul-dlss/sul-requests/pull/3407/changes#diff-00bba4e1ac42a0da1a34e92292781fd4e3fec81b68ba406b1224f8c728a409e2R151, but I'm not sure that's necessary either.

PR #3408, opened 2026-04-02
Serve the JS sourcemap for development debugging

marlo-longley:
We build but don't serve, so it's broken in devtools


taylor-steve:
@marlo-longley could you show me what this is doing differently? On main, I'm seeing the map get generated and placed in `app/assets/builds` and the map load:

Screenshot 2026-04-03 at 8 51 14 AM

PR #3412, opened 2026-04-03
don't create an accordion for single items

dnoneill:


taylor-steve:
I know you put this up as an example, so please take these comments as to the group. No action needed.

I can see the logic in saying these are different things. A few things:

* This doesn't remove the need to deal with the single accordion case in the item selector form, where `selectable_items.one?` isn't true (e.g., an EAD with multiple boxes but only one selected), so in a way this seems like additional complexity, going from "everything is an accordion, sometimes we want to pretend they are not" to "sometimes things are accordions, sometimes we want to pretend they are not, sometimes they aren't accordions at all"
* In the case of the new component path (`selectable_items.one?`) the item step form doesn't work (e.g., I can't change request type), presumably because the new component doesn't have the controller
* It'd be nice to see it solving the issue https://github.com/sul-dlss/sul-requests/issues/3313, this still renders the accordion:

Screenshot 2026-04-10 at 8 28 24 AM

cbeer:
Until we have a stronger motivation to extract separate components, #3437 just adds a flag to choose between an accordion presentation or a disabled-accordion-button presentation.

Issue #3396, opened 2026-04-01
Cancel appointment - Updates


dbranchini:
### Context

Appointments usually have items assigned to them. When a user cancels an appointment, what should happen with the items? Initially, we decided to keep it simple and save the items for later (aka Drafts) automatically. We received feedback on this choice, and so now we will let the user decide. 

### Preview

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/fa4e8894-93df-49cb-9d8a-3ad0f0358df2" />

### Notes

- Change heading from "Delete appointment?" to "Cancel appointment?"
- Change buttons to new copy/text on linked design
- The "safe" option is selected for the user by default, but they have the option to change it

PR #3418, opened 2026-04-06
refactor modal to use form id to submit via footer

dnoneill:
part of #3396

### DIFF
--- a/app/components/aeon/appointment_component.html.erb
+++ b/app/components/aeon/appointment_component.html.erb
@@ -52,22 +52,20 @@
     <%= render Aeon::AppointmentLimitComponent.from_appointment(appointment) %>
   </div>
   <!-- Delete modal -->
-  <%= render ModalComponent.new(id: "delete-appointment-#{appointment.id}") do |component| %>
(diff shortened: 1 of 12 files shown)


cbeer:
I've proposed some PRs to unblock the blocked PRs (#3432, #3427, maybe one more to come). I think having examples of how the proposed refactor gets used in practice will be informative, because I'm having trouble seeing the hoped-for-state.

I think we have some soon-to-come examples of modals without buttons, but it makes me wonder if `cancel_text`/`submit_text` is creating the right seam. Maybe when we start that new work, it'll help clarify the requirements.

I agree with Steve that `params[:modal]` is surprising; I guess it's a consequence of using `ModalComponent` as both the modal-rendered-inline and modal-contents-loaded-by-turbo? I wonder if trying to conflate those two concepts is causing more trouble than benefit (and the benefit seems to be not repeating some div tags?). I wonder if extracting some other components (`ModalLayoutComponent`? `ModalFooterButtons`?) could DRY the code without introducing complexity. I haven't gone far enough down this road to figure out other consequences..

PR #3437, opened 2026-04-08
Disable uses of single accordion buttons for item forms

taylor-steve:
I needed to take a stab at #3313 in order to understand https://github.com/sul-dlss/sul-requests/pull/3407 and https://github.com/sul-dlss/sul-requests/pull/3412

I'm reasonably confident 54d3c25a1b0465f3add831663e7983813d1db399 is on the right track.

731e5169ee9c3a401ceea52e26ca7d75443a27f1 makes an assumption that only holds true if we do not intend to render sets of existing requests to edit. I use this pattern in 'Save for later' too.


dnoneill:
@taylor-steve you also need for single item, see https://github.com/sul-dlss/sul-requests/blob/a4493df65367fbe3c95e2e46891bfd78b7b65d2c/app/views/patron_requests/_digitization_options.html.erb#L5-L7. But maybe that is what you mean with your render sets. I don't quite understand the sentence

PR #3443, opened 2026-04-09
Look up the EAD XML url from the finding aid (at least for archives.s…

cbeer:
…tanford.edu finding aids)

### DIFF
--- a/app/services/finding_aid_lookup_service.rb
+++ b/app/services/finding_aid_lookup_service.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# Service for going from the ARK url in a FOLIO record to the actual EAD XML url.
+class FindingAidLookupService
+  def initialize(ark_url)
+    @ark_url = ark_url
+  end
+
+  def ead_url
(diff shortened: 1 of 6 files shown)


app/services/finding_aid_lookup_service.rb
```diff
@@ -0,0 +1,29 @@
...
+    finding_aid_id = actual_location.split('/').last
+    "https://archives.stanford.edu/download/#{finding_aid_id}.xml"
+  end
+
+  private
+
+  def http_client
+    @http_client ||= Faraday.new do |f|
+      f.use Faraday::FollowRedirects::Middleware, limit: 3
+      f.request :retry, max: 4, interval: 5, backoff_factor: 2
```
marlo-longley, line 25: If the retries fail, does the whole thing error out? Do we need a rescue?

cbeer: Hm. I was thinking there isn't really a reasonable rescue here (besides the normal HTTP 500 error page).

marlo-longley: OK that makes sense. I think I was confused about the fallback flow. If the EAD XML fails that's where the chain ends.

marlo-longley (approved):

PR #3465, opened 2026-04-14
add alert for new appointments, remove warning alert

dnoneill:
closes #3410 

Figma:
Screenshot 2026-04-15 at 1 20 19 PM
PR:

Screenshot 2026-04-15 at 2 27 06 PM
Screenshot 2026-04-15 at 2 25 54 PM

### DIFF
--- a/app/models/aeon/reading_room.rb
+++ b/app/models/aeon/reading_room.rb
@@ -91,6 +91,10 @@ module Aeon
 
     def persisted? = id.present?
 
+    def item_limit
+      Settings.aeon.item_limits[sites.first]
+    end
+
     private
 
     def format_hours(reading_room_open_hours)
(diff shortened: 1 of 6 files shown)


taylor-steve:
Looking good.

* #3410 wants this on the appointments page as well
* It'd be nice if this worked when creating more than one appointment. Right now I create an appointment, see the alert, clear the alert, add a second new appointment and I don't get an alert.

dnoneill:
@taylor-steve Both should be fixed. I would like to confirm my assumption we should only ever have the most recent alert. For example, I create an appointment for April 22 and then another for April 23. Right now I have April 23 replacing April 22. I am assuming we don't want this to get long and we should just have the most recent.
I should have read the ticket better. It does bring up a point for why I didn't think it would be on the create appointments page based on my reading of the alert. The appointment alert doesn't include a reading room and the create appointments allow for reading room to be set by the user (by comparison to create requests which do not). Do we need to be including that info to be clear? Or are we going to pretend we didn't see it.

taylor-steve:
When I have no existing appointments for the given reading room (I was requesting a single box from [here](http://localhost:3000/archives_requests/new?Value=https%3A%2F%2Farchives.stanford.edu%2Fdownload%2Feal0087.xml%3Fwithout_namespace%3Dtrue)) I got this behavior after creating a new appointment:

Screenshot 2026-04-15 at 12 28 04 PM

dnoneill:
@taylor-steve fixed

taylor-steve (approved):

PR #3468, opened 2026-04-15
Just tell the component if we want an accordion or not.

cbeer:
I think this pushes #3437 back in line with some of Marlo's original work in #3407


marlo-longley:
Thanks @cbeer. Yes this seems essentially my same approach but better explicit naming of the parameter as `accordion` vs my `single_item`.

marlo-longley (approved):

taylor-steve (approved):

PR #3472, opened 2026-04-15
Use a special unset token to distinguish between unset values and val…

cbeer:
…ues we actively want to remove.


cbeer:
I thought we merged this already, but  🤷

taylor-steve (approved):

PR #3476, opened 2026-04-15
Add search to the EAD item selector

taylor-steve:
Closes #3425

Darcy approved dropping the bold text and using different highlight colors for this first round.

Screenshot 2026-04-15 at 3 44 37 PM

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -431,11 +431,6 @@ label.btn-close {
   z-index: 6;
 }
 
-.ead-search-pill {
-  font-size: 0.875rem;
-}
-
-
 .series-header[aria-expanded="true"] .disclosure-icon::before {
  transform: rotate(90deg);
 }
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -419,7 +419,8 @@ label.btn-close {
   transition: none !important;
 }
 
-.ead-search-input {
+/* Create space for the search results pill that's absolutely positioned on top of the control */
+.ead-search-input-group:has(.ead-search-overlay:not(.d-none)) .ead-search-input {
   padding-right: 11rem;
 }
 
@@ -427,6 +428,7 @@ label.btn-close {
   display: none;
 }
 
+/* Minimum z-index required for the search results pill buttons to overtake the search input */
 .ead-search-overlay {
   z-index: 6;
 }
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/application.css
```diff
@@ -405,6 +405,37 @@ label.btn-close {
...
+.ead-search-input::-webkit-search-cancel-button {
+  display: none;
+}
+
+.ead-search-overlay {
+  z-index: 6;
+}
+
+.ead-search-pill {
+  font-size: 0.875rem;
```
cbeer: Can we just a `.small` to the DOM element?

app/assets/stylesheets/application.css
```diff
@@ -405,6 +405,37 @@ label.btn-close {
...
+  background-color: var(--stanford-poppy-light);
+}
+
+.ead-search-no-transition .collapsing,
+.ead-search-no-transition .collapse {
+  height: auto !important;
+  transition: none !important;
+}
+
+.ead-search-input {
```
cbeer:
```suggestion
/* Create space for the search widget that's absolutely positioned on top of the control */
.ead-search-input {
```

I feel like there's maybe a future refactor to pull the textbox styling up so we don't need this hack, but in the meantime, I think a comment helps explain what's special about 11rem

app/assets/stylesheets/application.css
```diff
@@ -405,6 +405,37 @@ label.btn-close {
...
+  padding-right: 11rem;
+}
+
+.ead-search-input::-webkit-search-cancel-button {
+  display: none;
+}
+
+.ead-search-overlay {
+  z-index: 6;
```
cbeer, line 433: Is 6 special (to get it over something?), or would one of the e.g. `z-3` classes do just as well?

cbeer:
I tried  this with https://archives.stanford.edu/catalog/m1711 searching for e.g. catgut. I think I was surprised at the matches-the-series-name behavior, especially when it is the first match. It doesn't expand anything and seems to highlight everything in the first box of the series?

taylor-steve:
> I tried this with https://archives.stanford.edu/catalog/m1711 searching for e.g. catgut. I think I was surprised at the matches-the-series-name behavior, especially when it is the first match. It doesn't expand anything and seems to highlight everything in the first box of the series?

Good catch, thanks. I figured out the Safari specific issue (I think 🤞) and changed matching series titles to expand by default.

Issue #3466, opened 2026-04-14
Update design for Aeon request components


taylor-steve:
From the parent issue:

- Designs with specs
  - [Lists](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6417-97417&t=Wxx1BBfF3Hhp0AH3-4)

Let's get the request/request group and associated components looking like the new designs and the specs updated. This issue should look critically at the components and see if we still need the same components, new components, etc.

I think ideally we keep this work small and handle responsiveness in a subsequent issue to make the work easier to keep track of and review. If you feel ambitious and want to look ahead to responsiveness:

  - [Responsive - submitted](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6452-102001&t=Wxx1BBfF3Hhp0AH3-4) (no sidebar example)
  - [Responsive - saved for later](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6452-110690&t=Wxx1BBfF3Hhp0AH3-4) (sidebar example)

PR #3496, opened 2026-04-17
Add the filled background to the appointments card

cbeer:
Part of #3466

Screenshot 2026-04-17 at 11 20 54

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -24,6 +24,10 @@ layer(components);
   --bs-danger-text-emphasis: #58151c;
   --bs-danger-bg-subtle: #f8d7da;
   --bs-danger-border-subtle: #f1aeb5;
+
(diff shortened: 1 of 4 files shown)


cbeer:
The padding on the searchworks single-item request looks maybe a little funky in the lookbook, but I'm pretty sure actual requests have more data that'd show up there (format, pages, etc).

app/assets/stylesheets/application.css
```diff
@@ -491,13 +491,16 @@ label.btn-close {
...
 }
+.request-group-per-appointment.request-group.card:first-child {
+  border-top: 0;
+}
+
 
 .request-group-per-appointment.request-group.card:last-child {
-  border-bottom: none;
   margin-bottom: 0;
   padding-bottom: 0;
```
dnoneill: This should probably get removed as well. There is uneven padding if there is the last item. See 

Image
Image

dnoneill (approved):

PR #3501, opened 2026-04-17
wrap earliest appointment inline with label

### LINKED ISSUE #3485: Fix earliest appointment wrapping
Image

The "2026" should wrap under "Reading room" instead of the radio button.

Darcy would also be happy if the radio button was vertically centered for the whole group.

dnoneill:
closes #3485 
This removes the wrapping of the text to a new line. In my opinion this looks better but I can update this to do what we were previously doing, it is going to be kind of a pain and I think look worse but I can do it.
Screenshot 2026-04-17 at 3 31 29 PM
Screenshot 2026-04-17 at 3 31 24 PM

### DIFF
--- a/app/views/patron_requests/_reading_or_digitization.html.erb
+++ b/app/views/patron_requests/_reading_or_digitization.html.erb
@@ -1,15 +1,16 @@
 
 <fieldset class="ps-2">
-  <div>
+  <div class="d-flex">
     <%= f.radio_button :request_type, 'pickup', class: 'form-check-input', required: true, data: { action: 'patron-request#updateType', terms: false } %>
-    <%= f.label :request_type_pickup, class: 'form-check-label ms-1' do %>
-      Reading room appointment
-    <% end %>
-    <span class="text-cardinal ms-2 mt-1">
-      Earliest appointment available:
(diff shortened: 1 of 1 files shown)


app/views/patron_requests/_reading_or_digitization.html.erb
```diff
@@ -1,15 +1,16 @@
...
     <%= f.radio_button :request_type, 'pickup', class: 'form-check-input', required: true, data: { action: 'patron-request#updateType', terms: false } %>
-    <%= f.label :request_type_pickup, class: 'form-check-label ms-1' do %>
-      Reading room appointment
-    <% end %>
-    <span class="text-cardinal ms-2 mt-1">
-      Earliest appointment available:
-      <%= turbo_frame_tag 'earliest_appointment', src: available_aeon_appointments_path(f.object.aeon_reading_room&.id, Date.today) %>
-    </span>
-
+    <div class="ms-1 d-flex flex-wrap ms-1 d-flex flex-wrap column-gap-1">
```
taylor-steve: Some duplicate classes here. Also any chance we can retain the margin we had before, we lost it here.

Image

I do think this wrapping approach looks good, I'd be down to try it.

taylor-steve (changes requested):

taylor-steve (approved):

PR #3502, opened 2026-04-17
Update the layout + top-level design for the RequestGroup

cbeer:
Part of #3466, just getting the top-level card layout closer. Dealing with the individual boxes + items will happen in a future PR.
Screenshot 2026-04-17 at 13 26 29

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -18,6 +18,7 @@ layer(components);
 :root {
   --stanford-jungle-green-rgb: 39, 153, 137;
   --stanford-process-black-50-rgb: 151, 150, 148;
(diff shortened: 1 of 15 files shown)


taylor-steve (approved):
We might want to update https://github.com/sul-dlss/sul-requests/blob/73629c498b9d2c38dcd690f1e598d435568890f8/app/helpers/aeon_requests_helper.rb#L9 to use the group component here, in https://github.com/sul-dlss/sul-requests/pull/3505, or later... I'm approving everything so you can merge how you want 😄

cbeer:
I think you're probably right, but there's (already) some funky stuff going on. I'd like to do a little more refactoring and then figure out the update action?

PR #3505, opened 2026-04-20
Always render grouped requests.

cbeer:
My impression from #3466 is we're always rendering grouped requests. If that's true, there's a next step for cleaning + consolidating.


dnoneill:
I mean the text doesn't look quite right but I don't know if that is outside the scope of this PR, but yes I think your understanding is correct.
Screenshot 2026-04-20 at 1 51 20 PM
Screenshot 2026-04-20 at 1 51 11 PM

taylor-steve (approved):
I also think that's right. I'm happy we can consolidate to a single approach.

PR #3510, opened 2026-04-21
Extract aeon_appontments#items action.

cbeer:


app/controllers/aeon_appointments_controller.rb
```diff
@@ -69,6 +70,16 @@ def destroy
     redirect_to aeon_appointments_path, notice: 'Appointment cancelled successfully'
   end
 
+  def add_items
```
dnoneill: Can there be an indicator that this is the modal view or something. I use this path as a post for actually adding the items

cbeer: Just so I understand, you mean, in a different PR you plan to use `add_items` as your action name too and you'd like me to pick a different name here  so it doesn't conflict with your work?

PR #3511, opened 2026-04-21
Extract aeon_appointments#items action.

cbeer:


app/components/aeon/appointment_component.rb
```diff
@@ -26,8 +26,8 @@ def add_item_disabled?
       appointment.requests.count >= appointment.reading_room.appointment_item_limit
     end
 
-    def add_items_path
-      draft_aeon_requests_path(filter: 'reading_room', sort: 'title', appointment_id: appointment.id)
+    def items_path
+      aeon_appointment_items_path(sort: 'title', aeon_appointment_id: appointment.id)
```
dnoneill, line 31: Is there a reason we can't just change this to id? And not have the or in the controller?

cbeer: 🤷 Rails ends up wanting to call the route `items_aeon_appointment` instead of `aeon_appointment_items`. I'm not sure it matters either way.

dnoneill: I mean right now we are passing appointment_id as the param. Is there a reason we can't make it id so we don't have to have the or in this line 

https://github.com/sul-dlss/sul-requests/pull/3511/changes#diff-15944d0dd4827beac57feba1bb8d81b28da7cfe90260d268882c6dc800d74b83R109

cbeer: I understand. If we pass `id`, the Rails generated route name is weird. I guess I'd prefer a sensible route name and a pretty innocuous `||` over a confusing route name 🤷‍♂️

dnoneill: Thanks for explaining. I wasn't sure if it was a holdover from the original route.

taylor-steve (approved):

PR #3516, opened 2026-04-22
update item limit display

### LINKED ISSUE #3514: Item limit display - Update
### Context

Since reading rooms have different item limits, and some that go up to 9 or 10 items, we decided against 5 dots. Instead we will update the "progress" indicator to a short line or bar. [See Bootstrap progress bars.](https://getbootstrap.com/docs/5.3/components/progress/)

This affects two screens - Appointments + Saved for later.

### Design

- [Appointments](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6417-98597&t=m2NfkvphKJl3mrls-4)
- [Saved for later](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6417-98040&t=m2NfkvphKJl3mrls-4)

dnoneill:
closes #3514 

Screenshot 2026-04-23 at 5 26 37 PM
Screenshot 2026-04-23 at 5 26 31 PM

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -91,6 +91,10 @@ layer(components);
   color: var(--stanford-30-black);
 }
 
+.text-80-black {
+  color: var(--stanford-80-black);
+}
+
 @font-face {
   font-family: "Stanford";
   src: url("https://www-media.stanford.edu/assets/fonts/stanford.woff") format("woff"), url("https://www-media.stanford.edu/assets/fonts/stanford.ttf") format("truetype");
@@ -857,4 +861,18 @@ label.btn-close {
       transform: none;
     }
   }
+}
+
+.progress {
+  border-radius: 99px;
+  min-width: 50px;
+  --bs-progress-bg: var(--stanford-20-black);
+}
+
+.progress-bar.text-bg-success {
+  --bs-success-rgb: var(--stanford-digital-green-rgb);
+}
+
+.progress-bar.text-bg-danger {
+  --bs-danger-rgb: var(--stanford-digital-red-light-rgb);
 }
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/application.css
```diff
@@ -858,3 +862,17 @@ label.btn-close {
...
   }
 }
+
+.progress {
+  border-radius: 99px;
+  min-width: 50px;
+  --bs-progress-bg: var(--stanford-20-black);
+}
+
+.progress-bar .text-bg-success {
```
taylor-steve: This isn't applying for me, I assume because you've got a descendant combinator here in the CSS but not the HTML.

Image

app/components/aeon/appointment_limit_component.rb
```diff
@@ -9,35 +9,26 @@ def self.from_appointment(appointment, **)
...
-      @filled_dots ||= (total_dots * (count / limit.to_f)).to_i.clamp(1, total_dots)
-    end
-
-    def dot_used_class
-      if filled_dots == total_dots
-        'text-danger'
-      elsif filled_dots == total_dots - 1
-        'text-warning'
+    def bar_color
+      if percentage == 100
```
taylor-steve:
```suggestion
      if percentage >= 100
```

We probably shouldn't trust ourselves or Aeon 😄

app/models/aeon/reading_room.rb
```diff
@@ -40,7 +40,7 @@ def self.from_dynamic(dyn) # rubocop:disable Metrics/AbcSize, Metrics/MethodLeng
     end
 
     def appointment_item_limit
-      Settings.aeon.default_appointment_item_limit
+      Settings.aeon.item_limits[sites.first]
```
taylor-steve, line 43: This can be `nil` if the site isn't in the settings, right? I think this might raise an ArgumentError in the comparison in `AppointmentComponent#add_item_disabled?`. Maybe we should maintain a default.

dnoneill: If we maintain a default then displaying the no item limit wouldn't make sense. I am going to change add_item_disabled? instead

taylor-steve (approved):

PR #3524, opened 2026-04-27
hook up activities backend

dnoneill:
part of #3423

### DIFF
--- a/app/models/aeon/activity.rb
+++ b/app/models/aeon/activity.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Aeon
+  # Wraps an Aeon appointment record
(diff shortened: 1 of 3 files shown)


taylor-steve (approved):
This is fine for now, but just noting for future us that in prod this is an expensive API call. A ~2.3 MB response that takes almost a full second under ideal circumstances (8s when Aeon was cold).

dnoneill:
@taylor-steve Is there a way around this?

taylor-steve:
You're correct to do it this way now. I wouldn't be surprised if we eventually reach for caching and/or fetching it via a job. Even better, lobby Atlas to add a date filter or something to their endpoint 😄

PR #3532, opened 2026-04-28
Add a method to get active requests only from the Aeon API

taylor-steve:
I'm a special case because I have so many cancelled requests, but this reduces the Aeon response time from 8s -> 0.75s for me.


taylor-steve:
Unfortunately, it seems we can't get away with this.

1. Our notion of persisting digitization requests as submitted after the scan has been delivered is incompatible
2. We currently consider "Awaiting Future Request Processing" as a "submitted" state, while Aeon considers it "inactive".

Screenshot 2026-05-01 at 9 05 15 AM

PR #3536, opened 2026-04-28
add activity to request form

dnoneill:
part of #3423. Confirmation screen on separate PR.

Screenshot 2026-04-30 at 11 15 35 AM

Screenshot 2026-04-30 at 11 15 43 AM

### DIFF
--- a/app/models/aeon/request.rb
+++ b/app/models/aeon/request.rb
@@ -8,6 +8,9 @@ module Aeon
     # appointment attributes
     attr_accessor :appointment, :appointment_id
 
+    # activity attributes
+    attr_accessor :activity_type, :activity_id
+
     # identifiers
     attr_accessor :call_number, :ead_number, :reference_number, :site
 
@@ -27,6 +30,8 @@ module Aeon
     def self.from_dynamic(dyn) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       photoduplication_date = dyn['photoduplicationDate'].presence
       new(
+        activity_type: dyn.dig('requestFor', 'type'),
+        activity_id: dyn.dig('requestFor', 'reference'),
         appointment: dyn['appointment'] ? Appointment.from_dynamic(dyn['appointment']) : nil,
         appointment_id: dyn['appointmentID'],
         call_number: dyn['callNumber'],
@@ -111,6 +116,8 @@ module Aeon
     def valid?
       if digital?
         requested_pages.present?
+      elsif activity?
+        true
       else
         appointment_id.present?
       end
@@ -122,6 +129,10 @@ module Aeon
 
     def digital?
       shipping_option == 'Electronic Delivery'
+    end
+
+    def activity?
+      activity_id.present?
     end
 
     def ead?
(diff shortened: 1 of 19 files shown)


app/models/aeon/request.rb
```diff
@@ -111,6 +116,8 @@ def draft?
     def valid?
       if digital?
         requested_pages.present?
+      elsif activity?
+        activity?
```
taylor-steve:
```suggestion
        true
```
Is this what you're saying?

app/views/patron_requests/_activity.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <div class="activity-item" data-content-id="__ID__">
+      <span class="text-black fw-semibold selected-item-title">__TITLE__</span>
+    </div>
+  </template>
+  <div class="selected-items-container px-0 gap-3 d-flex flex-row" data-item-selector-target="selectedItems" data-template="#activityTemplate">
+  </div>
+  <div class="dropdown custom-select px-0 pt-3">
+    <%= button_tag 'Select activity', class: 'py-1 btn border dropdown-toggle', type: 'button', data: { bs_toggle: 'dropdown' }, aria: { expanded: false } %>
+    <menu class="dropdown-menu appt-dropdown">
+      <% current_user.aeon.active_reading_room_activities(site: @patron_request.aeon_site)&.each do |activity| %>
```
taylor-steve: Given this and the similar pattern in `app/views/patron_requests/_reading_or_digitization.html.erb`, it seems like adding something to the controller or adding a helper might be in order?

taylor-steve (commented):
@dnoneill I'm literally not trying to be obtuse here, it'd help me review this quicker if you could include more detail as to what part of the issue this covers and what's still to come.

app/models/aeon/request.rb
dnoneill: yes. I kept waffling on which one to use. activity? checks if activity_id.present? so I thought maybe that was better but yeah that is what I am checking.

app/views/patron_requests/_activity.html.erb
dnoneill: I don't understand. What am I adding. And what pattern are you seeing?

app/models/aeon/request.rb
taylor-steve: All I'm saying is if the condition is true, we should return true, to eliminate the tautology and avoid re-computing the predicate method.

app/views/patron_requests/_activity.html.erb
taylor-steve: I'm saying to me, this sort of work doesn't seem like a view's responsibility: `current_user.aeon.active_reading_room_activities(site: @patron_request.aeon_site)&.each do |activity| `

Same with: `current_user.aeon.active_reading_room_activities(site: @patron_request.aeon_site).present?`

Typically I'd expect to do that sort of model drilling in a helper or the controller 🤷‍♂️

dnoneill: Gotcha. Done.

taylor-steve (approved):

PR #3545, opened 2026-04-29
Only pass Aeon item data appropriate for the request type

taylor-steve:
I'm still noodling on this one, but you can make digitization requests with appointments currently 😈 

Screenshot 2026-04-28 at 5 47 11 PM

### DIFF
--- a/app/models/patron_request.rb
+++ b/app/models/patron_request.rb
@@ -18,6 +18,14 @@ class PatronRequest < ApplicationRecord
 
   delegate :instance_id, :finding_aid, :finding_aid?, to: :folio_instance
 
+  # Per-item fields that are meaningful for each Aeon request_type. Hidden accordion sections in the
+  # form still submit their inputs, so we drop anything that doesn't belong to the selected type.
+  AEON_ITEM_FIELDS_BY_REQUEST_TYPE = {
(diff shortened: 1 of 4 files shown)


dnoneill:
I think you are going to have to account for activity as well

cbeer:
Do we need to remove them at the controller level, or just when we send over the Aeon request? It might be easier to do later 🤷‍♂️

taylor-steve:
I was debating that, but it seems odd and confusing to me to store garbage data in the request.

cbeer:
Related to #3588?

taylor-steve:
How angry does this make folks? It only makes me a little angry. I could push this same logic to the job, but I do like storing only the data we need.

If we keep it this way, there's an argument for dropping the two feature specs I was using for confirming the bug.

dnoneill:
Have you looked at keeping the same pattern we have for the folio-items and getting rid of data-required-for-submit?

taylor-steve:
> Have you looked at keeping the same pattern we have for the folio-items and getting rid of data-required-for-submit?

Are you talking about the client side code? I wouldn't be opposed to improving the client side, but I wouldn't want to rely on it.

dnoneill:
yes I am talking about the client side code

PR #3553, opened 2026-04-29
add confirmation screen for activities

dnoneill:
Screenshot 2026-04-29 at 3 50 34 PM

part of #3423 

Last part is to have the dropdown show which activities are selected.

### DIFF
--- a/app/components/aeon/confirmation_request_list_component.html.erb
+++ b/app/components/aeon/confirmation_request_list_component.html.erb
@@ -1,4 +1,3 @@
-
 <div class="border rounded p-3 mb-2">
   <h2 class="h3"><%= title %></h2>
   <ul class="list-unstyled mb-0">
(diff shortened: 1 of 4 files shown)


taylor-steve:
The designs show the date of the activity (`beginDate` presumably). Did we decide against that?

Screenshot 2026-04-30 at 2 50 08 PM

dnoneill:
@taylor-steve My understanding is that is supposed to be the name. I could have misunderstood @dbranchini incorrectly but my understanding is that right now a lot of the activities have the date included in the name.

taylor-steve:
Ah okay thanks. I don't think that's a true assumption. Out of the 1414 activities in prod I'm seeing maybe ~40 with dates in their name. We can iterate on it 🤷‍♂️

taylor-steve (approved):

PR #3555, opened 2026-04-29
Remove flag for draft requests with an assigned appointment.

cbeer:
I think we got rid of this case, right?


app/components/aeon/request_group_per_appointment_component.html.erb
```diff
@@ -2,9 +2,6 @@
   <div class="metadata">
     <h3 class="h4 d-inline">
       <%= request_group.title %>
-      <% if request_group.draft_requests.any? %>
```
dnoneill: I think this can be historically true (i.e. old Aeon) but not with the new forms?

PR #3562, opened 2026-04-29
Have the update action remove the whole request group if needed, instead of relying on a stimulus side-effect.

cbeer:
Is it better?? 

- [x] replace implicit "remove when empty" logic with an explicit removal
- [x] update the appointment sidebar


taylor-steve:
We'd need to update `destroy_multiple`, but this seems like a more maintainable approach. Needing to think about rows vs groups got confusing fast for me.

app/models/aeon/request_grouping.rb
```diff
@@ -14,7 +14,7 @@ class RequestGrouping
 
     def self.from_requests(requests)
       multi, single = requests.partition(&:multi_item_selector?)
-      groups = multi.group_by { |r| [r.title, r.digital?] }.values.map { |group| new(group) }
+      groups = multi.group_by { |r| [r.status, r.title, r.digital?] }.values.map { |group| new(group) }
```
taylor-steve, line 17: Do we need to update `dom_id`, now that we're grouping by status as well?

taylor-steve (approved):

app/models/aeon/request_grouping.rb
cbeer: I don't think so. In practice, our pages are already partitioned by status. This just preserves that on the update side.

PR #3566, opened 2026-04-30
Add custom datepicker component and Lookbook views

marlo-longley:
I'd like to walk folks through this and then deploy it to stage so that the experts in this flow can test live.
Hard to test the JS code in our testing environment, but no existing tests break. If this approach is generally approved I can add deeper testing.

Full disclosure the JS was largely vibecoded.

I tried to review for accessibility -- among other things, tabbing through the calendar grid works correctly.

I included multiple Lookbook pages to demonstrate different states of the calendar
Screenshot 2026-04-30 at 11 18 41 AM

----

Screenshot 2026-04-30 at 11 15 38 AM
Screenshot 2026-04-30 at 11 15 44 AM
Screenshot 2026-04-30 at 11 15 59 AM

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -907,4 +907,39 @@ label.btn-close {
 
 .zero-items-alert:only-child {
   display: flex;
+/* ---- Custom date picker ---- */
+.date-picker-popup {
+  position: absolute;
+  z-index: 1000;
+  background: white;
+  border: 1px solid var(--bs-border-color);
+  border-radius: var(--bs-border-radius);
+  box-shadow: var(--bs-box-shadow);
+  padding: 0.75rem;
+  min-width: 280px;
+}
+
+.date-picker-grid {
+  border-collapse: separate;
+  border-spacing: 2px;
+}
+
+.date-picker-day {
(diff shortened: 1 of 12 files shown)


dnoneill:
The date should probably be black 
Screenshot 2026-04-30 at 4 26 52 PM

dnoneill:
Since we are never going to let anyone set a past date should be really be showing the left arrow and allow users to check old dates?

dnoneill:
I have concerns that the disabled would be the rails backend side/gets loaded via the component. I think that is going to make the page load slow and we should probably be doing it on the front-end side. I am not quite sure what this looks like but I think this is currently structured from backend. But maybe that isn't an issue if we download the closures every night but I was under the impression we can't do that.

taylor-steve:
Hey @marlo-longley, this looks good to me, we should get it in front of Darcy and get her feedback.

A couple things we might consider:
* A way to identify and/or jump to "today"
* Adding similar behavior of the native date picker for page up/page down & arrow keys (Safari is weird, but I think Firefox and Chrome have consistent behavior here). Darcy might know more than me how important this is.

taylor-steve (approved):

PR #3574, opened 2026-05-01
Calculate the progress bar display client-side

cbeer:
This should allow add/remove items to update the progress bar atomically (and with only having to know the item count, not the appointment limit 🤞 )


dnoneill:
This is going to cause huge problems for add items. The progress bar is supposed to update when you add items but only on the modal. This will create a huge mess for me. Is there any way we can hold off on this.

cbeer:
I'm not sure I understand what problem you're flagging.

taylor-steve:
It seems like you could manipulate count dynamically?

taylor-steve (approved):

dnoneill:
Can we not merge this and I can show you what I mean

PR #3576, opened 2026-05-01
move activities dropdown to list

dnoneill:
closes #3423 

Figma:
Screenshot 2026-05-01 at 6 00 04 PM

Branch:
Screenshot 2026-05-01 at 6 29 01 PM

Before:
Screenshot 2026-05-01 at 6 16 26 PM

### DIFF
--- a/app/views/patron_requests/_activity.html.erb
+++ b/app/views/patron_requests/_activity.html.erb
@@ -6,17 +6,19 @@
   </template>
   <div class="selected-items-container px-0 gap-3 d-flex flex-row" data-item-selector-target="selectedItems" data-template="#activityTemplate">
   </div>
-  <div class="dropdown custom-select px-0 pt-3">
-    <%= button_tag 'Select activity', class: 'py-1 btn border dropdown-toggle', type: 'button', data: { bs_toggle: 'dropdown' }, aria: { expanded: false } %>
-    <menu class="dropdown-menu appt-dropdown">
+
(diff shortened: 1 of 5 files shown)


app/views/patron_requests/_activity.html.erb
```diff
@@ -6,17 +6,19 @@
...
       <% @activities&.each do |activity| %>
-        <%= tag.li class: 'd-flex align-items-end' do %>
-          <label class="mx-4">
-            <%= check_box_tag 'patron_request[activity_ids][]', class: 'visually-hidden', value: activity.id %>
-            <%= activity.name %>
+        <%= tag.li class: 'd-flex align-items-center p-2' do %>
+          <%= check_box_tag 'patron_request[activity_ids][]', id: "activity-" + activity.id.to_s,  value: activity.id %>
+          <label class="d-flex flex-column ms-3" for="activity-<%= activity.id %>">
+            <div class="fw-semibold"><%= activity.activity_type %><i class="bi bi-dot"></i><%= activity.name %></div>
+            <div><%= render AppointmentTimeRangeComponent.new(appointment: activity, show_hours: true, location: activity.location) if activity.start_time && activity.location %></div>
```
taylor-steve, line 18: With the render guards here and the added conditionals in `AppointmentTimeRangeComponent` I wonder if these are two different components.

dnoneill: I don't think so and considering how many time components we have I would prefer not to but I can if it is a blocker.

taylor-steve (approved):

Issue #3525, opened 2026-04-27
New appointment - messaging updates


dbranchini:
### Context

This ticket outlines further messaging refinements for the Create a new appointment modal. Keep in mind that appointments can be made from the request form (and then the reading room is known) or from the appointments page where the user must select the appointment first. Messaging should address the following:

- Reading room hours (general)
- [x] business days in advance
- Earliest available (it's on Step 1 of the new request form, and now we'll repeat it on the new appointment modal)
- Save (not a message, but currently it jumps after you select the date... can we do something to keep it in the same place? maybe a minimum height on the body of the modal?)

### Designs

- [RR known - date and time](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5335-22828&t=WbMwQcSMyq0GA16O-4) (East Asia, David Rumsey, ARS)
- [RR known - date only](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5335-22829&t=WbMwQcSMyq0GA16O-4) (Field Reading Room)
- [RR unknown/user selects - date and time](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5335-22830&t=WbMwQcSMyq0GA16O-4)
- [RR unknown/user selects - date only](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5335-22831&t=WbMwQcSMyq0GA16O-4)

PR #3578, opened 2026-05-02
Add create new appointment info alert

taylor-steve:
Part of #3525 

Screenshot 2026-05-01 at 5 00 01 PM

If we're cool with the Alert component, I'll use that going forward and start replacing existing alerts.

### DIFF
--- a/app/components/alert_component.rb
+++ b/app/components/alert_component.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# Alert with a leading icon. Pass a type (:info, :success, :warning).
+class AlertComponent < ViewComponent::Base
+  ICONS = {
+    info: 'info-circle-fill',
+    success: 'check-circle-fill',
+    warning: 'exclamation-triangle-fill'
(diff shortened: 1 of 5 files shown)


app/components/alert_component.rb
```diff
@@ -0,0 +1,28 @@
...
+  }.freeze
+
+  def initialize(type:)
+    @type = type
+  end
+
+  attr_reader :type
+
+  def icon_class
+    "bi bi-#{ICONS.fetch(type)}"
```
cbeer:
```suggestion
    info: 'bi-info-circle-fill',
    success: 'bi-check-circle-fill',
    warning: 'bi-exclamation-triangle-fill'
  }.freeze

  def initialize(type:)
    @type = type
  end

  attr_reader :type

  def icon_class
    "bi #{ICONS.fetch(type)}"
```

I think something like this was the actual intent behind the herb warning (so you could grep for all the `bi-info-circle-fill` and turn them into something else)?

PR #3582, opened 2026-05-04
add activity page

dnoneill:
part of #3580 

Filtering needs to be added

image


app/models/aeon/user.rb
```diff
@@ -27,15 +27,25 @@ def sso_auth?
...
-      @requests ||= self.class.aeon_client.requests_for(username:)
+      @requests ||= self.class.aeon_client.requests_for(username:).reject(&:activity?)
     end
 
     def activities
-      @activities ||= self.class.aeon_client.activities_for(username:)
+      @activities ||= self.class.aeon_client.activities_for(username:).sort_by(&:sort_key)
+    end
+
+    def activites_with_requests # rubocop:disable Metrics/AbcSize
```
taylor-steve:
```suggestion
    def activities_with_requests # rubocop:disable Metrics/AbcSize
```

app/models/aeon/user.rb
```diff
@@ -27,15 +27,25 @@ def sso_auth?
...
+      @cached_requests = {}
+      activities.each do |activity|
+        activity.users.each do |user|
+          @cached_requests[user.username] ||= self.class.aeon_client.requests_for(username: user.username)
+          activity.requests.concat(@cached_requests[user.username].select { |request| request.activity_id == activity.id })
```
taylor-steve: `activity.requests` is memoized in  `Aeon::Activity` and concat is mutating it in place. The user's `Aeon::User` is also memoized. The effect here is this `activities_with_requests` isn't idempotent, `activity.requests` size will grow with every subsequent call to `activities_with_requests`. This isn't a problem now, because we only have the one call, but I think it's worth changing now.

app/controllers/aeon_activities_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+###
+#  Controller for displaying Aeon activites for a user
+###
+class AeonActivitiesController < ApplicationController
```
taylor-steve, line 6: Should we be including the `AeonController` concern?

taylor-steve (commented):
Do we know how remove is supposed to work? Right now it errors out. I'm asking in a "I don't know know, we haven't talked about it" sort of way. Can anybody attached to the activity delete any request associated with the activity?

dnoneill:
@taylor-steve Yeah I don't know how delete is supposed to work. Are people allowed to delete requests other uses have made? I assume so? But yeah I missed that functionality in the first pass.

app/models/aeon/user.rb
dnoneill: Are you thinking we would return @activities if @activities.requests.present?

taylor-steve: I haven't thought about it that much; I can if you're stuck. I'd expect to be able to run `activites_with_requests` twice in the same request and get the same result from each call.

dnoneill: I added that guard. We can revisit but I think that works.

app/models/aeon/user.rb
```diff
@@ -27,15 +27,27 @@ def sso_auth?
...
+    end
+
+    def activities_with_requests # rubocop:disable Metrics/AbcSize
+      return @activities if @activites && @activities.first.requests.present?
+
+      @cached_requests = {}
+      activities.each do |activity|
+        activity.users.each do |user|
+          @cached_requests[user.username] ||= self.class.aeon_client.requests_for(username: user.username)
+          activity.requests.concat(@cached_requests[user.username].select { |request| request.activity_id == activity.id })
```
taylor-steve: I guess I was hoping to avoid the mutation completely (also you have a spelling error on the memoize check).

Something like https://github.com/sul-dlss/sul-requests/compare/3580-activity-page...activity-no-mutate?expand=1

dnoneill: Makes sense. Done.

taylor-steve (approved):

PR #3593, opened 2026-05-05
Add the earliest appointment available messaging

taylor-steve:
Part of #3525 

Screenshot 2026-05-05 at 3 55 39 PM
Screenshot 2026-05-05 at 3 57 48 PM

### DIFF
--- a/app/views/aeon_appointments/_form.html.erb
+++ b/app/views/aeon_appointments/_form.html.erb
@@ -15,7 +15,10 @@
 
     <div class="mb-3">
       <%= f.label :date, class: 'd-block fw-semibold redesign-required-label' %>
-      <%= f.date_field :date, required: true, data: { action: 'appointment#refreshAvailability' }, min: (@appointment.reading_room.next_appointment.start_time.to_date.iso8601 if @appointment.reading_room&.next_appointment), class: 'border rounded px-2 py-1' %>
+      <div class="d-flex align-items-center gap-2">
+        <%= f.date_field :date, required: true, data: { action: 'appointment#refreshAvailability' }, min: (@appointment.reading_room.next_appointment.start_time.to_date.iso8601 if @appointment.reading_room&.next_appointment), class: 'border rounded px-2 py-1' %>
+        <%= turbo_frame_tag 'earliest_appointment', src: available_aeon_reading_room_url(@appointment.reading_room_id, date: Date.today) %>
+      </div>
     </div>
 
     <% if @appointment&.reading_room&.day_only_appointments? %>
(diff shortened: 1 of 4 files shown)


app/views/aeon_appointments/_form.html.erb
```diff
@@ -15,7 +15,14 @@
 
     <div class="mb-3">
       <%= f.label :date, class: 'd-block fw-semibold redesign-required-label' %>
-      <%= f.date_field :date, required: true, data: { action: 'appointment#refreshAvailability' }, min: (@appointment.reading_room.next_appointment.start_time.to_date.iso8601 if @appointment.reading_room&.next_appointment), class: 'border rounded px-2 py-1' %>
+      <div class="d-flex align-items-center gap-2">
+        <%= f.date_field :date, required: true, data: { action: 'appointment#refreshAvailability' }, min: (@next_available_appointment.start_time.to_date.iso8601 if @next_available_appointment), class: 'border rounded px-2 py-1' %>
+        <% if @next_available_appointment %>
+        <span class="text-cardinal">
```
dnoneill: We should probably do this on the front end like we are doing with the form. https://github.com/sul-dlss/sul-requests/blob/da6487b2a260c368f53f3c5bcec49c9aed988feb/app/views/patron_requests/_reading_or_digitization.html.erb#L11

dnoneill: Or since there is already a turbo load it should be added to that turbo template

app/views/aeon_appointments/_form.html.erb
```diff
@@ -15,7 +15,10 @@
 
     <div class="mb-3">
       <%= f.label :date, class: 'd-block fw-semibold redesign-required-label' %>
-      <%= f.date_field :date, required: true, data: { action: 'appointment#refreshAvailability' }, min: (@appointment.reading_room.next_appointment.start_time.to_date.iso8601 if @appointment.reading_room&.next_appointment), class: 'border rounded px-2 py-1' %>
+      <div class="d-flex align-items-center gap-2">
+        <%= f.date_field :date, required: true, data: { action: 'appointment#refreshAvailability' }, min: (@appointment.reading_room.next_appointment.start_time.to_date.iso8601 if @appointment.reading_room&.next_appointment), class: 'border rounded px-2 py-1' %>
+        <%= turbo_frame_tag 'earliest_appointment', src: available_aeon_reading_room_url(@appointment.reading_room_id, date: Date.today) %>
```
dnoneill, line 20: I don't think you need to add src to this. I could be wrong but the same url is on the page and it loads both ids

dnoneill (approved):

PR #3598, opened 2026-05-06
add in filtering, sort to activities, add past activities

dnoneill:
closes #3580 

https://github.com/user-attachments/assets/b72900c3-ad4f-45d5-b273-f39d04100357

### DIFF
--- a/app/controllers/aeon_activities_controller.rb
+++ b/app/controllers/aeon_activities_controller.rb
@@ -4,9 +4,40 @@
 #  Controller for displaying Aeon activites for a user
 ###
 class AeonActivitiesController < ApplicationController
+  ALLOWED_SORTS = %w[sort_key name activity_type].freeze
   include AeonController
 
-  def index
-    @activities = current_user.aeon.activities_with_requests.select(&:active?)
+  def index; end
+
+  def active
+    @activities = all_activities.select(&:active?)
+    @activity_types = @activities.map(&:activity_type)
+    filter
+    sort_results
+  end
+
+  def past
+    @activities = all_activities.reject(&:active?)
+    @activity_types = @activities.map(&:activity_type)
+    filter
+    sort_results
+  end
+
+  private
+
+  def filter
+    return unless params[:filter]
+
+    @activities = @activities.filter { |activity| activity.activity_type == params[:filter] }
+  end
+
+  def sort_results
+    sort = params[:sort].presence_in(ALLOWED_SORTS) || 'sort_key'
+    sort_keys = ([sort] + ALLOWED_SORTS).compact.uniq
+    @activities = @activities.sort_by { |obj| sort_keys.map { |k| obj.public_send(k) } }
+  end
+
+  def all_activities
+    @all_activities ||= current_user.aeon.activities_with_requests
   end
 end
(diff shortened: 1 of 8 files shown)


app/controllers/aeon_activities_controller.rb
```diff
@@ -4,9 +4,34 @@
...
+
+  def filter
+    return unless params[:filter]
+
+    @activities = @activities.filter { |activity| activity.activity_type == params[:filter] }
+  end
+
+  def sort_results
+    sort_keys = [params[:sort], 'sort_key', 'name', 'activity_type'].compact.uniq
+    @activities = @activities.sort_by { |obj| sort_keys.map { |k| obj.public_send(k) } }
```
taylor-steve: I'd feel better if we were verifying params[:sort] was an expected value before sending unfiltered user input to a `public_send`

app/controllers/aeon_activities_controller.rb
```diff
@@ -4,9 +4,34 @@
...
+  end
+
+  def past
+    @activities = all_activities.reject(&:active?)
+    @activity_types = @activities.map(&:activity_type)
+    filter
+    sort_results
+  end
+
+  def filter
```
taylor-steve, line 28: All of these public methods are routable actions, I think some of them should probably be private.

config/routes.rb
```diff
@@ -72,6 +72,8 @@
   end
 
   resources :aeon_activities, only: [:index]
+  get "/aeon_activities/active", to: 'aeon_activities#active', as: :aeon_activities_active
+  get "/aeon_activities/past", to: 'aeon_activities#past', as: :aeon_activities_past
```
taylor-steve: I think we could do this here if we wanted to use the idiomatic path helpers.
```suggestion
resources :aeon_activities, only: [:index] do
  collection { get :active; get :past }
end
```

app/controllers/aeon_activities_controller.rb
```diff
@@ -4,9 +4,34 @@
 #  Controller for displaying Aeon activites for a user
 ###
 class AeonActivitiesController < ApplicationController
-  include AeonController
+  def index; end
```
taylor-steve, line 10: Was dropping `include AeonController` intentional?

app/controllers/aeon_activities_controller.rb
dnoneill: So make sure it is one of the 3 values we expect?

app/controllers/aeon_activities_controller.rb
dnoneill: no. probably a rebase issue.

app/controllers/aeon_activities_controller.rb
taylor-steve: yeah that'd work

dnoneill:
@taylor-steve comments addressed

taylor-steve:
Looks good. Only thing now is we might want labels:

Screenshot 2026-05-11 at 2 50 11 PM

dnoneill:
@taylor-steve oops. I had them. They weren't implemented correctly.

taylor-steve (approved):

PR #3599, opened 2026-05-06
Address accessibility concerns and add tests for custom datepicker

marlo-longley:


spec/features/aeon_date_picker_spec.rb
```diff
@@ -0,0 +1,186 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# Tests keyboard/focus behaviour of the custom Stimulus date picker.
+# Uses Lookbook previews as test pages so these run independently of
+# the appointment form (which uses the native date field today).
+RSpec.describe 'Date picker keyboard navigation', :js do
+  # Today is the initial focusedDate (min = today, no reading room in the preview).
+  let(:today) { Time.zone.today }
```
taylor-steve, line 9: If you're going to use fixed day offsets, like `today + 3`, I think you'll need to set today via something like `travel_to` in order to get a deterministic test. Otherwise it depends on what day of the week the test is run on (I re-ran them and this file failed).

marlo-longley: Thanks @taylor-steve -- addressed.

spec/features/aeon_date_picker_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+
+    context 'with marked days' do
+      before { visit '/lookbook/preview/aeon/appointment_date_picker/with_marked_days' }
+
+      it 'is visible when the current month has marked days' do
+        open_picker
+        legend = find('[data-date-picker-target="legend"]', visible: :all)
+        expect(legend[:style]).to include('visibility: visible')
+        expect(legend).to have_text('Existing appointment')
+      end
```
taylor-steve, line 141: This is still going to fail at the end of every month, when the `+2` overflows into the next month. E.g., try setting `today = Date.new(2026, 5, 30)` in `with_marked_days.html.erb` and running the test.

You might want to set that preview to be more predictable? Something like:
```
today = Time.zone.today.beginning_of_month
marked = [today + 2, today + 5, today + 9, today + 14].map(&:iso8601)
```

marlo-longley: @taylor-steve you are right. Thank you.

taylor-steve (approved):

PR #3604, opened 2026-05-07
add turbo frame spinner when loading

dnoneill:
https://github.com/user-attachments/assets/343b8940-d610-4e7e-9b29-54f99b1bfbc5

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1121,3 +1121,48 @@ label.btn-close {
     display: inline;
   }
 }
+
+turbo-frame[busy]:not(#available_appointments){
+  &:not(.modal-content):not(#earliest_appointment) {
+    height: 500px;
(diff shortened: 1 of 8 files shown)


taylor-steve:
I think standardizing our loading spinner is a good idea, but I wonder if we also need to standardize how we communicate that to screen readers, given how slow Aeon requests can be. I just checked with Voice Over and main tells me when the appointment sidebar is loading, while this does not, so I guess turbo busy doesn't give us that for free. We should ask Darcy, I'm not sure what's right/best here.

dbranchini:
I think using [placeholders on the sidebar](https://getbootstrap.com/docs/5.3/components/placeholders/) makes more sense than a spinning circle. Is there anyway to show the modal header while you're waiting for the rest of it to load?

dnoneill:
@dbranchini Pretty sure we can't show the modal header until the actual page we are waiting for loads.

PR #3609, opened 2026-05-07
Instantiate the model from the model's class, not the client.

cbeer:


app/models/folio/patron.rb
```diff
@@ -3,11 +3,15 @@
...
-    def self.find_by(sunetid: nil, library_id: nil, patron_key: nil, **_kwargs)
-      return folio_client.find_patron_by_id(patron_key) if patron_key.present?
-      return folio_client.find_patron_by_barcode_or_university_id(library_id) if library_id.present?
-      return folio_client.find_patron_by_sunetid(sunetid) if sunetid.present?
-
+    def self.find_by(sunetid: nil, library_id: nil, patron_key: nil, **_kwargs) # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+      user_info = folio_client.find_user_by_id(patron_key) if patron_key.present?
+      user_info ||= folio_client.find_user_by_barcode(library_id) if library_id.present?
+      user_info ||= folio_client.find_user_by_legacy_barcode(library_id) if library_id.present?
+      user_info ||= folio_client.find_user_by_sunetid(sunetid) if sunetid.present?
```
taylor-steve, line 11: Did `find_user_by_university_id` get lost, or is it unneeded?

cbeer: Yeah, my mistake. The legacy barcode finder threw off my counting.

That said, I'm really not sure we actually need all these cases. `patron_key` is used for sure, `library_id` is used for some pseudopatron stuff (but presumably we control whether that data is in the university id, barcode, or legacy barcode?), and I'm pretty sure sunetid is unused (but useful for development?).

taylor-steve (approved):

PR #3618, opened 2026-05-08
add appointment dropdown into drafts sidebar

### LINKED ISSUE #3517: Saved for later sidebar list on Appointments screen - Updates
### Context

Sidebar that contains "Saved for later" on the Appointments screen needs to be cleaned up

### Requested changes

- [x] Update Heading and subheading
- [ ] Add padding
- [ ] Update "cards" for each title
- [ ] Add Appointment dropdown to each item

### Preview

Image

[Browser view](https://www.figma.com/proto/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6417-98597&viewport=81%2C25%2C0.43&t=ACGOgLB0JslOIb0w-9&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=6417%3A98597&show-proto-sidebar=1&page-id=) (note Appointments dropdown active for the first two items)

### Notes

- Saved for later should be a list of requested items for reading room appointments (and does not include digitization items)
- If a user clicks on the Appointments link from the list of Saved for later items, a dropdown is displayed with available appointments associated with that item - meaning the reading room associated with that item.

dnoneill:
closes #3517 

https://github.com/user-attachments/assets/32ecf410-da2c-4a6d-aaa5-7504fa81edca

### DIFF
--- a/app/components/aeon/appointment_option_component.rb
+++ b/app/components/aeon/appointment_option_component.rb
@@ -4,11 +4,16 @@ module Aeon
   # Render an option for an appointment in the custom appointment select-ish dropdown.
   class AppointmentOptionComponent < ViewComponent::Base
     with_collection_parameter :appointment
-    attr_reader :appointment
+    attr_reader :appointment, :data_action
 
-    def initialize(appointment:, data: {})
+    def initialize(appointment:, data: {}, data_action: 'appointment-select#select')
       @appointment = appointment
       @data = data
+      @data_action = data_action
+    end
+
+    def at_limit?
+      appointment.requests.count >= (appointment.reading_room.appointment_item_limit || 100)
     end
   end
 end
--- a/app/components/sidebar_appointment_dropdown_component.html.erb
+++ b/app/components/sidebar_appointment_dropdown_component.html.erb
@@ -1,0 +1,11 @@
+<%= form_with(model: request, data: { turbo: true}) do |f| %>
+  <div class="dropdown custom-select" data-controller="appointment-select-submit">
+    <%= hidden_field_tag 'sidebar', true %>
+    <%= f.hidden_field :appointment_id, data: { appointment_select_submit_target: "input" } %>
+    <%= button_tag 'Appointment', class: 'dropdown-toggle btn btn-link su-underline p-0', type: 'button', data: { bs_toggle: 'dropdown', appointment_select_target: 'button' }, aria: { expanded: false }, disabled: selectable_appointments.blank? %>
+    <menu class="dropdown-menu sidebar-appt-dropdown">
+      <li></li>
+      <%= render Aeon::AppointmentOptionComponent.with_collection(selectable_appointments, data_action: 'appointment-select-submit#updateInput') %>
+    </menu>
+  </div>
+<% end %>
--- a/app/controllers/aeon_requests_controller.rb
+++ b/app/controllers/aeon_requests_controller.rb
@@ -48,7 +48,7 @@ class AeonRequestsController < ApplicationController
     @updated_request = Aeon::UpdateRequestService.new(@aeon_request, aeon_request_params).call
 
     respond_to do |format|
-      format.turbo_stream
+      format.turbo_stream { update_turbo_stream }
       format.html do
         aeon_requests_path = updated_request.draft? ? aeon_requests_path(kind: 'drafts') : aeon_requests_path(kind: 'submitted')
         redirect_to aeon_requests_path, notice: 'Request was successfully updated.'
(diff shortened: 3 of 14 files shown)


app/components/aeon/appointment_option_component.rb
```diff
@@ -4,11 +4,16 @@ module Aeon
...
 
-    def initialize(appointment:, data: {})
+    def initialize(appointment:, data: {}, data_action: 'appointment-select#select')
       @appointment = appointment
       @data = data
+      @data_action = data_action
+    end
+
+    def at_limit?
+      appointment.requests.count >= appointment.reading_room.appointment_item_limit
```
taylor-steve: Image

app/controllers/aeon_requests_controller.rb
```diff
@@ -48,14 +48,26 @@ def update
...
-      format.turbo_stream
+      format.turbo_stream { update_turbo_stream }
       format.html do
         aeon_requests_path = updated_request.draft? ? aeon_requests_path(kind: 'drafts') : aeon_requests_path(kind: 'submitted')
         redirect_to aeon_requests_path, notice: 'Request was successfully updated.'
       end
     end
   end
 
+  def update_turbo_stream
```
taylor-steve: Can we move this to private?

app/javascript/controllers/appointment_select_submit_controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "@hotwired/stimulus"
```
taylor-steve, line 1: Can we drop the extra 'p' in the filename?

spec/features/aeon_appointment_sidebar_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+require 'rails_helper'
+
+RSpec.describe 'Appointments', :js do
+  let(:user) { create(:sso_user) }
+  let(:current_user) { CurrentUser.new(username: user.sunetid, shibboleth: true) }
+  let(:aeon_user) { Aeon::User.new(username: user.email_address, auth_type: 'Default') }
+  let(:reading_rooms) { JSON.load_file('spec/fixtures/reading_rooms.json').map { |rr| Aeon::ReadingRoom.from_dynamic(rr) } }
+
+  let(:field_reading_room) { reading_rooms[-1] }
+  let(:appointment) { build(:aeon_appointment, reading_room: field_reading_room, start_time: 1.week.from_now) }
```
taylor-steve, line 12: I think you'll need to `travel_to` or set a fixed start time, rather than base the test on the real date it's run. I could be wrong, but it's failing for me today.

taylor-steve (commented):
I couldn't test the functionality yet, because of the error.

spec/features/aeon_appointment_sidebar_spec.rb
dnoneill: Setting to a fixed start time will make the test fail in the future. The appointment needs to be editable (3 days before whatever today is). So if I set it for 2027 in 2027 we need to update the date. I could make it like 2050 but I figured it would be better to click_on "#{I18n.l(1.week.from_now, format: :date_only)}"

dnoneill:
@taylor-steve functionality should be working now

app/components/sidebar_appointment_dropdown_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= form_with(model: request, data: { turbo: true}) do |f| %>
+  <div class="dropdown custom-select" data-controller="appointment-select-submit">
+    <%= hidden_field_tag 'sidebar', true %>
+    <%= f.hidden_field :appointment_id, data: { appointment_select_submit_target: "input" } %>
+    <%= button_tag 'Appointment', class: 'dropdown-toggle btn btn-link su-underline p-0', type: 'button', data: { bs_toggle: 'dropdown', appointment_select_target: 'button' }, aria: { expanded: false }, disabled: selectable_appointments.blank? %>
+    <menu class="dropdown-menu appt-dropdown">
+      <li></li>
```
taylor-steve, line 7: Do we need the empty `li`?

taylor-steve (commented):
Anything we can do about the appointment deletion case?
For example if I have two Field appointments:

Image

Delete one:
Image

The appointment in the main content goes away, but I can still select it in the sidebar, to failure:

Image

taylor-steve:
@dbranchini here are the three states you wanted recorded:

Request with an active option:
Screenshot 2026-05-12 at 3 12 36 PM

Request that only has appointments that exceed their limit:
Screenshot 2026-05-12 at 3 12 19 PM

Request with no available appointments for the given reading room:
Screenshot 2026-05-12 at 3 12 10 PM

app/components/sidebar_appointment_dropdown_component.html.erb
dnoneill: I am matching the pattern in the form https://github.com/sul-dlss/sul-requests/blob/327d40b57a29a2e90284b8d51655bcb7ed32a246/app/components/aeon/appointment_form_item_component.html.erb#L29

dnoneill:
@taylor-steve good catch on the deletion. Also found it on the create appt side. Both fixed now.

taylor-steve (approved):

PR #3628, opened 2026-05-12
fix save for later delete

dnoneill:
closes #3624


taylor-steve:
This is what I'm seeing for a reading room request for [Knuth](http://localhost:3000/archives_requests/new?value=https%3A%2F%2Farchives.stanford.edu%2Fdownload%2Fsc0097.xml). I think I'm clobbered and up to date.

Screenshot 2026-05-12 at 11 28 26 AM

dnoneill:
Don't love the solution I came up with that problem but I am having trouble thinking of something better. Choosing to do something like Box 1 will be problematic for accessibility

taylor-steve (approved):

PR #3629, opened 2026-05-12
Add a stimulus sortable controller for re-sorting DOM elements by som…

cbeer:
…e data attribute.

This is hopefully a sufficient replacement for https://github.com/sul-dlss/mylibrary/blob/main/app/javascript/listSort.js 🤞

### DIFF
--- a/spec/component_previews/sortable_component_preview/default.html.erb
+++ b/spec/component_previews/sortable_component_preview/default.html.erb
@@ -1,0 +1,17 @@
+<div data-controller="sortable">
+  <div class="btn-group" data-sortable-target="menu">
+    <button type="button" class="btn btn-primary dropdown-toggle" data-bs-toggle="dropdown" aria-expanded="false">
+      Sort by ...
+    </button>
+    <ul class="dropdown-menu">
+      <li><button class="dropdown-item" data-action="click->sortable#sort" data-sortable-label-param="Sort by title" data-sortable-sort-param="title">Title</button></li>
+      <li><button class="dropdown-item" data-action="click->sortable#sort" data-sortable-label-param="Sort by price" data-sortable-sort-param="price">Price</button></li>
+    </ul>
+  </div>
+
+  <ul data-sortable-target="list" aria-live="polite">
+    <li data-title-sort-value="Banana" data-price-sort-value="50">Banana $.50</li>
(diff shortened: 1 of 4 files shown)


spec/component_previews/sortable_component_preview/default.html.erb
```diff
@@ -0,0 +1,17 @@
...
+      Sort by ...
+    </button>
+    <ul class="dropdown-menu">
+      <li><button class="dropdown-item" data-action="click->sortable#sort" data-sortable-target="option" data-sortable-label-param="Sort by title" data-sortable-sort-param="title">Title</button></li>
+      <li><button class="dropdown-item" data-action="click->sortable#sort" data-sortable-target="option" data-sortable-label-param="Sort by price" data-sortable-sort-param="price">Price</button></li>
+    </ul>
+  </div>
+
+  <ul data-sortable-target="list" aria-live="polite">
+    <li data-sortable-target="item" data-title-sort-value="Banana" data-price-sort-value="50">Banana $.50</li>
```
taylor-steve: `data-sortable-target="item"` and `data-sortable-target="option"` seem to be undeclared/unused in the controller. Not sure if they're cruft or work to come.

spec/component_previews/sortable_component_preview/default.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <li data-sortable-target="item" data-title-sort-value="Apple" data-price-sort-value="80">Apple $.80</li>
```
taylor-steve: Noting that currently it's a string comparison only right now, right? I can change 80 -> 800 and Apple still sorts to the middle for price.

taylor-steve (approved):

spec/component_previews/sortable_component_preview/default.html.erb
cbeer: 👍 .. i guess i gave up on keeping the lookbook in-sync.

spec/component_previews/sortable_component_preview/default.html.erb
cbeer: Yes, and I think that's what we're actually dealing with for all our values (rightly or wrongly... there's a fee amount sort that i wonder if we've previously done wrong 🤷 )

PR #3640, opened 2026-05-13
Update jest etc

cbeer:
I don't see any indication we're actually using these tools/libraries (besides esbuild, obviously), so updating them should be fine, right?


marlo-longley (commented):
Yes. I think we are just keeping them around in the hopes of adding JS tests?

Needs rebase though.

cbeer:
Yeah, I assume (although if we haven't done anything in 2-3 years... I guess we'll see. Maybe the date picker?)

marlo-longley (approved):

PR #3675, opened 2026-05-15
Fix confirmation page when making Aeon activity requests

taylor-steve:
Maybe `expected_aeon_item_count` is too brittle and we should just write the time all the requests hit Aeon in the DB.

closes #3676


dnoneill:
@taylor-steve Looks good. Can you add a test that makes sure the confirmation screen populates for activities?

dnoneill (approved):

PR #3677, opened 2026-05-15
Update Aeon::RequestGrouping to use delegated methods.

cbeer:


app/models/aeon/request_grouping.rb
```diff
@@ -23,17 +23,17 @@ def initialize(requests)
     end
 
     def dom_id
-      return "group_#{requests.first.id}" unless requests.first.multi_item_selector?
+      return "group_#{first.id}" unless multi_item_selector?
```
marlo-longley, line 26: would `.id` not also be able to be delegated to first?

cbeer: I guess in my mind `RequestGrouping#id` suggests that it's an id for the group, and that's not what we really mean.

marlo-longley (approved):

PR #3681, opened 2026-05-15
Do more granular turbo updates to avoid whole-frame reloads

cbeer:


app/controllers/aeon_requests_controller.rb
```diff
@@ -121,7 +140,7 @@ def load_aeon_requests # rubocop:disable Metrics/AbcSize,Metrics/MethodLength,Me
                      when 'activity'
                        current_user.aeon.activities_with_requests.map(&:requests).flatten
                      else
-                       current_user.aeon.requests
```
dnoneill, line 124: I don't think you need this considering line 129. Also is there a reason we don't want the requests?

cbeer: I'm not sure what you're asking. If we hit this block, we were asked to filter on some unknown "kind", so I thought an empty array made sense.

dnoneill: I could see that. My question is there is the line ```@aeon_requests = [] and return unless current_user&.aeon```

Which sets @aeon_requests to an empty [] which would mean I don't think you need the else because that block is @aeon_requests =

dnoneill: maybe you do. I'm just confusing myself now

cbeer: I'm still confused. That line is covering a different case (no aeon user at all), unless you're suggesting I handle both "error" cases there?

dnoneill: I say ignore me. My brain never works Friday afternoon

taylor-steve:
This seems like a good improvement. Two things I'm still digging into but aren't necessarily blockers (sorry I'm about to get on a plane...):
- [x] On the appointments page, I have two EAL saved for laters and one EAL request with an appointment. When I "save for later" the EAL request with an appointment, the entire EAL saved for later sidebar element goes away.
- [x] We'll need to handle the item limit in the appointment dropdowns in the saved for later sidebar, I think it was depending on the reload. I don't think we need to fix that here, just making note.

taylor-steve (approved):

PR #3711, opened 2026-05-23
add remove button to activities, keep save for later

dnoneill:
closes #3686

### DIFF
--- a/app/components/aeon/cancel_or_save_for_later_button_component.rb
+++ b/app/components/aeon/cancel_or_save_for_later_button_component.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Aeon
+  # Render aeon remove/save for later button
+  class CancelOrSaveForLaterButtonComponent < ViewComponent::Base
+    attr_reader :request
+
+    def initialize(request:)
+      @request = request
+    end
+
+    def call
+      return remove_button if request.activity?
+
+      save_for_later_button
+    end
+
+    private
+
+    def remove_button
+      form_with(url: aeon_request_path(request, kind: request.request_type), method: :delete) do
+        tag.button(type: :submit, class: 'btn btn-link su-underline') do
+          tag.i(class: 'bi bi-trash align-middle me-1')
+        end
+      end
+    end
+
+    def save_for_later_button
+      form_with(url: redraft_aeon_request_path(request, kind: request.request_type)) do
+        tag.button(type: :submit, class: 'btn btn-link su-underline') do
+          tag.i(class: 'bi bi-pin-angle-fill align-middle me-1') + tag.span('Save for later')
+        end
+      end
+    end
+  end
+end
--- a/app/components/aeon/activity_component.html.erb
+++ b/app/components/aeon/activity_component.html.erb
@@ -1,4 +1,4 @@
-<%= render CardComponent.new do |c| %>
+<%= render CardComponent.new(id: dom_id(activity)) do |c| %>
   <% c.with_title do %>
     <div class="d-flex flex-wrap gap-3 align-items-center">
       <%= render PillComponent.new(classes: %w[bg-bay-dark text-white]).with_content(activity.activity_type) %>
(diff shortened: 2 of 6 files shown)


app/components/aeon/cancel_or_save_for_later_button_component.rb
```diff
@@ -0,0 +1,48 @@
...
+    def initialize(request:)
+      @request = request
+    end
+
+    def delete?
+      request.activity?
+    end
+
+    def render?
+      helpers.current_user.email_address == request.username
```
taylor-steve: This seems a little specific for a component, any chance we could use CanCan like in other areas? E.g., `helpers.can? :destroy...`

This also changes current behavior, showing the button only for owned requests on Activities rather than all. I'm not sure if that's the expected behavior or not, I'll have to ask Sarah/Darcy.

app/components/aeon/cancel_draft_button_component.rb
```diff
@@ -0,0 +1,48 @@
...
+      return aeon_request_path(request, kind: request.request_type) if delete?
+
+      redraft_aeon_request_path(request, kind: request.request_type)
+    end
+
+    def text
+      'Save for later' unless delete?
+    end
+
+    def method
```
taylor-steve: It'd be good if we could make this name more specific, `method` is an existing core method on all Ruby objects.

app/components/aeon/cancel_draft_button_component.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Aeon
+  # Render aeon remove/save for later button
+  class CancelDraftButtonComponent < ViewComponent::Base
```
taylor-steve: I'm not sure I understand this name. In the case of the appointments page I'm moving a submitted request to saved for later. Given the name, I would assume the role of the component would be to provide a button to cancel a saved for later request.

app/components/aeon/cancel_or_save_for_later_button_component.rb
dnoneill: I put it in because can can is currently blocking the delete on requests that aren't the user's. I could remove that instead but I figured this might be a better first step?

app/components/aeon/cancel_draft_button_component.rb
dnoneill: It meant cancel or draft which I can see how it is confusing. Maybe CancelOrSaveForLatterButton?

app/components/aeon/cancel_draft_button_component.rb
dnoneill: went with form_method

app/components/aeon/cancel_or_save_for_later_button_component.rb
dnoneill: Okay I see you did that on a different PR. I have removed this.

app/components/aeon/cancel_or_save_for_later_button_component.rb
```diff
@@ -0,0 +1,44 @@
...
+      'bi bi-pin-angle-fill align-middle me-1'
+    end
+
+    def call
+      form_with(url:, method: form_method) do
+        tag.button type: :submit, class: 'btn btn-link su-underline' do
+          tag.i(class: icon_classes) + tag.span(text)
+        end
+      end
+    end
```
taylor-steve: I don't really mind either way, but when I see the multiple calls to `delete?` I wonder if we're really just saying this:

```suggestion
    def call
      return remove_button if request.activity?

      save_for_later_button
    end

    private

    def remove_button
      form_with(url: aeon_request_path(request, kind: request.request_type), method: :delete) do
        tag.button(type: :submit, class: 'btn btn-link su-underline') do
          tag.i(class: 'bi bi-trash align-middle me-1')
        end
      end
    end

    def save_for_later_button
      form_with(url: redraft_aeon_request_path(request, kind: request.request_type)) do
        tag.button(type: :submit, class: 'btn btn-link su-underline') do
          tag.i(class: 'bi bi-pin-angle-fill align-middle me-1') + tag.span('Save for later')
        end
      end
    end
```

taylor-steve (approved):

app/components/aeon/cancel_or_save_for_later_button_component.rb
dnoneill: Yeah I like that better.

Issue #3712, opened 2026-05-23
New Date Picker Time Issues


taylor-steve:
These specs fail every day when I run them past 5 PM PT, presumably because there's a mismatch between the timezone in Ruby and the timezone in the browser/JS. We should figure out if it's a spec problem or a real problem before we ship the new picker. I can't do that on a Friday 😄 

```
Failures:

  1) Date picker keyboard navigation on open focuses today in the grid
     Failure/Error: expect(focused_date).to eq(today.iso8601)

       expected: #<Encoding:US-ASCII> "2026-05-22"
            got: #<Encoding:UTF-8> "2026-05-23"

       (compared using ==)
     # ./spec/features/aeon_date_picker_spec.rb:30:in 'block (3 levels) in <top (required)>'

  2) Date picker keyboard navigation arrow key navigation ArrowRight moves forward one day
     Failure/Error: expect(focused_date).to eq((today + 1).iso8601)

       expected: #<Encoding:US-ASCII> "2026-05-23"
            got: #<Encoding:UTF-8> "2026-05-24"

       (compared using ==)
     # ./spec/features/aeon_date_picker_spec.rb:81:in 'block (3 levels) in <top (required)>'

  3) Date picker keyboard navigation arrow key navigation ArrowLeft moves back one day
     Failure/Error: expect(focused_date).to eq(today.iso8601)

       expected: #<Encoding:US-ASCII> "2026-05-22"
            got: #<Encoding:UTF-8> "2026-05-23"

       (compared using ==)
     # ./spec/features/aeon_date_picker_spec.rb:88:in 'block (3 levels) in <top (required)>'

  4) Date picker keyboard navigation arrow key navigation ArrowDown moves forward one week
     Failure/Error: expect(focused_date).to eq((today + 7).iso8601)

       expected: #<Encoding:US-ASCII> "2026-05-29"
            got: #<Encoding:UTF-8> "2026-05-30"

       (compared using ==)
     # ./spec/features/aeon_date_picker_spec.rb:94:in 'block (3 levels) in <top (required)>'

  5) Date picker keyboard navigation arrow key navigation ArrowUp moves back one week
     Failure/Error: expect(focused_date).to eq(today.iso8601)

       expected: #<Encoding:US-ASCII> "2026-05-22"
            got: #<Encoding:UTF-8> "2026-05-23"

       (compared using ==)
[...]

PR #3716, opened 2026-05-26
fix min date to take reading room policy into account for fallback, update date picker component to take appointment as variable instead of form

dnoneill:
I found this when looking at #3712. I think? it might fix it but won't know until we test after 5pm pacific. At the very least it should go in because our min should not be today if there is no next appointment. Use hide white space to review.


cbeer:
Are there reading rooms that aren't PT? We're [setting the application timezone](https://github.com/sul-dlss/sul-requests/blob/main/config/application.rb#L50), so `Time.zone.today` should always be PT.

dnoneill:
Yeah I see that now.

dnoneill:
done with https://github.com/sul-dlss/sul-requests/pull/3738

PR #3725, opened 2026-05-29
add authorization on activities

dnoneill:
Before:
Screenshot 2026-05-29 at 11 16 40 AM

after:
Screenshot 2026-05-29 at 11 16 19 AM

### DIFF
--- a/app/controllers/aeon_activities_controller.rb
+++ b/app/controllers/aeon_activities_controller.rb
@@ -7,6 +7,8 @@ class AeonActivitiesController < ApplicationController
   ALLOWED_SORTS = %w[sort_key name activity_type].freeze
   include AeonController
 
+  before_action :authorize_activity
+
   def index; end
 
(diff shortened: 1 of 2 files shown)


app/controllers/aeon_activities_controller.rb
```diff
@@ -25,6 +27,10 @@ def past
 
   private
 
+  def auth
+    authorize! :read, Aeon::Activity
+  end
+
```
cbeer: `auth` can mean so many things. Can we be more explicit, e.g.?

```suggestion
  before_action :authorize_resources

def index; end

def active
@@ -25,6 +27,10 @@ def past

private

  def authorize_resources
    authorize! :read, Aeon::Activity
  end

```

dnoneill: Went with authorize_activity to be more explicit

PR #3737, opened 2026-06-04
Extend Aeon ability for requests associated with an activity

taylor-steve:


app/models/abilities/aeon_ability.rb
```diff
@@ -18,14 +18,27 @@ def initialize(aeon_user) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
...
     can :read, Aeon::Request
     can :manage, Aeon::Appointment
   end
+
+  private
+
+  def member_of_request_activity?(aeon_user:, request:)
+    request.activity_id.present? &&
+      aeon_user.activities.any? { |activity| activity.id == request.activity_id }
+  end
```
cbeer: I wonder if it'd be clearer if we did some refactoring to write this as:

```
request.activity.users.include?(aeon_user)
```

app/models/abilities/aeon_ability.rb
```diff
@@ -18,14 +18,26 @@ def initialize(aeon_user) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
...
 
     can :read, Aeon::Activity
     can :read, Aeon::Request
     can :manage, Aeon::Appointment
   end
+
+  private
+
+  def member_of_request_activity?(aeon_user:, request:)
+    request.activity&.users&.include?(aeon_user)
```
cbeer, line 37: I suspect we'll have to define `Aeon::User#==` to make this work too?

taylor-steve: Yes, I had to define it

PR #3738, opened 2026-06-04
Update AppointmentDatePickerComponent to be more open to extension

cbeer:
... and use it for selecting appointments

### DIFF
--- a/app/components/aeon/appointment_date_picker_component.rb
+++ b/app/components/aeon/appointment_date_picker_component.rb
@@ -44,7 +44,7 @@ module Aeon
       # make a list of closed dates (where a closure covers the entire day a reading room is actually open)
       closures.flat_map do |closure|
         closure.start_date.to_date.upto(closure.end_date.to_date).to_a.select do |date|
-          hours_on_day = open_hours.find { |oh| oh.day_name == date.strftime('%A') }
+          hours_on_day = open_hours.find { |oh| oh.day_of_week == date.wday }
 
           next if hours_on_day.nil?
 
--- a/app/components/aeon/appointment_date_picker_component.rb
+++ b/app/components/aeon/appointment_date_picker_component.rb
@@ -7,19 +7,19 @@ module Aeon
   #   <%= render Aeon::AppointmentDatePickerComponent.new(:date, form: f) %>
   #   <%= render Aeon::AppointmentDatePickerComponent.new(:date, form: f, disabled: ['2026-05-01'], marked: ['2026-05-10']) %>
   class AppointmentDatePickerComponent < ViewComponent::Base
-    attr_reader :key, :form, :disabled, :marked
+    attr_reader :key, :form, :reading_room, :disabled, :marked
 
-    def initialize(key, form: nil, disabled: [], disabled_daynames: [], marked: [])
+    def initialize(key, form: nil, reading_room: nil, disabled: [], disabled_daynames: [], marked: [])
       @key = key
       @form = form
+      @reading_room = reading_room || form.object.reading_room
       @disabled = disabled
       @disabled_daynames = disabled_daynames
       @marked = marked
     end
 
     def min
-      appointment = form.object
-      next_start = appointment.reading_room&.next_appointment&.start_time
+      next_start = reading_room&.next_appointment&.start_time
       next_start&.to_date&.iso8601 || Time.zone.today.iso8601
     end
 
--- a/app/components/aeon/appointment_date_picker_component.rb
+++ b/app/components/aeon/appointment_date_picker_component.rb
@@ -4,12 +4,13 @@ module Aeon
   # Renders the custom Stimulus date-picker for the Aeon appointment date field.
   #
   # Usage:
-  #   <%= render Aeon::AppointmentDatePickerComponent.new(form: f) %>
-  #   <%= render Aeon::AppointmentDatePickerComponent.new(form: f, disabled: ['2026-05-01'], marked: ['2026-05-10']) %>
+  #   <%= render Aeon::AppointmentDatePickerComponent.new(:date, form: f) %>
+  #   <%= render Aeon::AppointmentDatePickerComponent.new(:date, form: f, disabled: ['2026-05-01'], marked: ['2026-05-10']) %>
   class AppointmentDatePickerComponent < ViewComponent::Base
-    attr_reader :form, :disabled, :marked, :disabled_daynames
+    attr_reader :key, :form, :disabled, :marked
 
-    def initialize(form:, disabled: [], marked: [], disabled_daynames: [])
+    def initialize(key, form: nil, disabled: [], disabled_daynames: [], marked: [])
+      @key = key
       @form = form
       @disabled = disabled
       @disabled_daynames = disabled_daynames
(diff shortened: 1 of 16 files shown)


app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -7,27 +7,22 @@ module Aeon
   #   <%= render Aeon::AppointmentDatePickerComponent.new(form: f) %>
   #   <%= render Aeon::AppointmentDatePickerComponent.new(form: f, disabled: ['2026-05-01'], marked: ['2026-05-10']) %>
```
dnoneill: This comment needs to be changed. I think it would also be good to include a list of possible data values that can be passed and will work. I would hate to have to track down what to pass in order for the JS to work. I think we should have these values documented somehwere

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -7,27 +7,22 @@ module Aeon
...
+    attr_reader :key, :form, :reading_room, :data
 
-    def initialize(form:, disabled: [], marked: [], disabled_daynames: [])
+    def initialize(key, form: nil, reading_room: nil, data: {})
+      @key = key
       @form = form
-      @disabled = disabled
-      @disabled_daynames = disabled_daynames
-      @marked = marked
+      @reading_room = reading_room || form.object.reading_room
```
dnoneill, line 16: Is this in here for a specific reason? I am not seeing where it is used and I am not sure what it is used for.

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -7,27 +7,22 @@ module Aeon
   #   <%= render Aeon::AppointmentDatePickerComponent.new(form: f) %>
   #   <%= render Aeon::AppointmentDatePickerComponent.new(form: f, disabled: ['2026-05-01'], marked: ['2026-05-10']) %>
   class AppointmentDatePickerComponent < ViewComponent::Base
-    attr_reader :form, :disabled, :marked, :disabled_daynames
+    attr_reader :key, :form, :reading_room, :data
 
-    def initialize(form:, disabled: [], marked: [], disabled_daynames: [])
+    def initialize(key, form: nil, reading_room: nil, data: {})
+      @key = key
       @form = form
```
dnoneill, line 15: I am wondering if instead of passing form we should be passing a model object and doing a fields_for(model) instead. Might not be relevant / too much for this PR.

app/components/aeon/appointment_date_picker_component.rb
cbeer: I didn't want to keep writing `form.object.appointment.reading_room` 🤷‍♂️

app/components/aeon/appointment_date_picker_component.rb
cbeer: I guess I'm unclear which parameters we'd actually pass into this component. Maybe it's just the marked dates, because everything else is determined by the reading room?

app/components/aeon/appointment_date_picker_component.rb
dnoneill: Ah. I missed below we are using reading room for min/max.

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -4,30 +4,66 @@ module Aeon
...
-      @marked = marked
+      @reading_room = reading_room || form.object.reading_room
+      @data = data
     end
 
     def min
-      appointment = form.object
-      next_start = appointment.reading_room&.next_appointment&.start_time
-      next_start&.to_date&.iso8601 || Time.zone.today.iso8601
+      next_start = reading_room&.next_appointment&.start_time
```
dnoneill, line 21: This probably needs to be updated to 

```suggestion
      next_start = reading_room&.next_appointment&.start_time || reading_room.appointment_min_lead_days&.days&.from_now
```

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -4,30 +4,66 @@ module Aeon
...
+      (next_start&.to_date || Time.zone.today).iso8601
+    end
+
+    def max
+      (reading_room&.policies || []).filter_map do |policy|
+        policy.request_max_lead_days.days.from_now.to_date.iso8601 if policy.request_max_lead_days
```
dnoneill: Maybe this shoud be it's own method in reading room? Also should this be request_max_lead_days or appointmentMaxLeadDays? I would assume appointment but maybe they are the same and it doesn't matter.

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -4,30 +4,66 @@ module Aeon
...
+      closures = reading_room.closures
+
+      # get the open hours for a reading room
+      open_hours = reading_room.open_hours
+
+      # select the closures that cover an entire day the reading room is actually open
+      closures.flat_map do |closure|
+        # iterate the days for which there are open hours
+        closure.start_date.to_date.upto(closure.end_date.to_date).to_a.select do |date|
+          hours_on_day = open_hours.find { |oh| oh.day_name == date.strftime('%A') }
```
dnoneill:
```suggestion
          hours_on_day = open_hours.find { |oh| oh.day_of_week == date.wday }
```

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -4,30 +4,66 @@ module Aeon
...
+
+          next if hours_on_day.nil?
+
+          open_hours_range_on_date = date.to_time.change(hour: hours_on_day.open_time.hour,
```
dnoneill, line 53: I have read this a couple times and I am in pretty heavy burn out so that might be why I just don't understand what this means, but open_hours_range_on_date doesn't mean anything to me

app/components/aeon/appointment_date_picker_component.rb
cbeer: I'll use `appointment_max_lead_days`, but it's 365 for both values for all reading rooms so 🤷‍♂️

app/components/aeon/appointment_date_picker_component.rb
cbeer: I'm open to suggestions. It's the ruby `Range` representation of the open hours for the reading room on the given date 🤷‍♂️

app/components/aeon/appointment_date_picker_component.rb
```diff
@@ -4,30 +4,66 @@ module Aeon
...
+      # get the closures for a reading room:
+      closures = reading_room.closures
+
+      # get the open hours for a reading room
+      open_hours = reading_room.open_hours
+
+      # select the closures that cover an entire day the reading room is actually open
+      closures.flat_map do |closure|
+        # iterate the days for which there are open hours
+        closure.start_date.to_date.upto(closure.end_date.to_date).to_a.select do |date|
```
dnoneill, line 48: I had to look this up. Can this comment be updated to indicated closures can span multiple days. Maybe that is obvious but the comment `iterate the days for which there are open hours` wasn't super clear to me. Especially since that isn't technically correct. We are iterating over everything we are just skipping closed days in line 50. Something like: iterate over all closure days in closure range

taylor-steve (approved):

PR #3750, opened 2026-06-05
make duration options accessible

dnoneill:
Before tab would skip duration options. Now it works
Screenshot 2026-06-05 at 5 31 45 PM

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -572,7 +572,7 @@ label.btn-close {
 }
 
 /* This can't go inline or display: none for when the turbo-frame is loading won't work */
-.appointment-slots, .duration-selectors {
+.appointment-slots {
   display: flex;
 }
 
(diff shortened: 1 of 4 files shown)


cbeer:
Is this the right fix, or are we just missing role="radiogroup" on the parent element (and maybe something to set an initial focus state)? Radio buttons controls are usually navigated using arrow keys (and that works today).

dnoneill:
I don't see focus state when I do that but I might be doing it wrong.

dnoneill:
@cbeer I tried pasting the html on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role to confirm that it should work. It did not. I wasn't able to tab/arrow to the next item. Also I think if we do this we will have an issue in terms of the focus ring.
Screenshot 2026-06-08 at 3 49 48 PM

cbeer:
I'm still not sure I understand. Over in https://github.com/sul-dlss/sul-requests/pull/3758, I've added an explicit focus-visible state and included a screen recording of keyboard interactions, and other than the semi-oddity of "no default selection for a required value", it behaves just as I'd expect?

PR #3758, opened 2026-06-08
Add a focus-visible style for the radio button fieldsets.

cbeer:
https://github.com/user-attachments/assets/9a2da3ce-a290-4774-9cfd-b05123a344ae

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -627,9 +627,7 @@ label.btn-close {
   }
 }
 
-.appointment-form {
-  --bs-border-color: var(--stanford-50-black);
-
+.radio-button-group {
   .btn-check:checked + .btn {
     --bs-border-color: var(--stanford-digital-blue);
     background-color: var(--stanford-digital-blue);
@@ -644,6 +642,11 @@ label.btn-close {
     margin-right: 0.25rem;
     text-decoration: none;
     vertical-align: middle;
+  }
+
+  .btn-check:focus-visible + .btn {
+    outline: 2px solid white;
+    box-shadow: 0 0 0 4px var(--stanford-black);
   }
 }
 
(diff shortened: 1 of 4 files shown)


dnoneill:
There is a issue where the radio will loose focus after the duration are reloaded. Also do we want the duration to be selected when a user tabs over?

https://github.com/user-attachments/assets/b5a83c14-3316-4304-b2ca-fb92816575e4

cbeer:
> There is a issue where the radio will loose focus after the duration are reloaded.

I don't think that behavior is newly introduced by the explicit focus styling, is it?  The only thing I can think to do is pull the duration out of the turbo frame, but unless it was caused by something here, it seems like it'd be better in a different PR. 

> Also do we want the duration to be selected when a user tabs over?

I'm not sure what you're asking.

dnoneill:
@cbeer You can see when I arrow over it actually selects the duration. I am not using the space bar to select the duration, it is doing this automatically when I arrow over.

cbeer:
I think that's just how radio buttons are. It's a little weird in our case because we don't have a value selected by default, but I think that issue is independent of adding a focus style?

Per the relevant ARIA docs:

> [Right Arrow and Down Arrow](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role#right_arrow)
> Moves focus to the next radio button in the group, unchecking the previously focused button, and checking the newly focused button. If focus is on the last button, focus moves to the first button.

> [Left Arrow and Up Arrow](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role#left_arrow)
> Moves focus to the previous radio button in the group, unchecking the previously focused button, and checking the newly focused button. If focus is on the first button, focus moves to the last button.

taylor-steve (approved):

PR #3765, opened 2026-06-09
Add a find method to reading room

taylor-steve:


app/models/aeon/reading_room.rb
```diff
@@ -13,6 +13,10 @@ def self.all
       aeon_client.reading_rooms
     end
 
+    def self.find(id)
+      all.find { |rr| rr.id == id.to_i }
```
cbeer, line 17: Do we need to e.g. memoize `all` to avoid multiple requests to Aeon?

taylor-steve: I don't think so, `reading_rooms` is already memoized in the aeon client

PR #3768, opened 2026-06-09
Add validations to Aeon appointments

taylor-steve:
Closes https://github.com/sul-dlss/sul-requests/issues/3753

Debating if it's worth displaying any of these failures to the user (if our front-end is correct, they should never see them...) or if we are only protecting against bad actors.

### DIFF
--- a/app/models/aeon/appointment.rb
+++ b/app/models/aeon/appointment.rb
@@ -77,13 +77,16 @@ module Aeon
 
     def save # rubocop:disable Metrics/AbcSize, Naming/PredicateMethod
       return false unless valid?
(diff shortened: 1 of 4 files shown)


app/models/aeon/appointment.rb
```diff
@@ -69,5 +74,36 @@ def editable?
     end
 
     def persisted? = id.present?
+
+    def save # rubocop:disable Naming/PredicateMethod
+      return false unless valid?
```
cbeer, line 79:
```suggestion
      return false unless valid?
      raise ArgumentError, 'Unable to re-create an existing appointment' if persisted?
```

Maybe? Just so I'm not confused when we add this for updates too.

PR #3785, opened 2026-06-09
Trim existing appointments from available Aeon appointment times

### LINKED ISSUE #3166: Prevent appointment overlap/conflicts
E.g., if I have an appointment at Rumsey on 03/04/2026 from 9AM - 11AM I shouldn't be able to schedule another appointment at 03/04/2026 from 10 AM to 11 AM.

The Aeon API doesn't handle this. Aeon handles this on the webapp side.

Darcy doesn't have designs for this yet. It seems sensible to start by either disabling or removing the time conflicts from the appointment picker, keeping in mind Darcy might eventually want to communicate the existing appointment to the user in some way.

taylor-steve:
Closes #3166

* No server side validation. Easy enough to add, but I think that'd take another Aeon API call and I'm not sure if that's worth it.
* ~~Currently this is a "user can't be in two places at once" approach, not scoped to a particular reading room~~

### DIFF
--- a/app/controllers/aeon_reading_rooms_controller.rb
+++ b/app/controllers/aeon_reading_rooms_controller.rb
@@ -45,9 +45,12 @@ class AeonReadingRoomsController < ApplicationController
   def own_appointment_conflicts
     return [] unless current_user&.aeon
 
-    excluded_id = params[:appointment_id]&.to_i
     current_user.aeon.appointments
-                .reject { |a| a.id == excluded_id }
+                .select { |a| a.reading_room_id == @reading_room.id && a.id != excluded_appointment_id }
                 .map { |a| a.start_time...a.stop_time }
+  end
+
+  def excluded_appointment_id
+    params[:appointment_id]&.to_i
   end
 end
--- a/app/controllers/aeon_reading_rooms_controller.rb
+++ b/app/controllers/aeon_reading_rooms_controller.rb
@@ -31,8 +31,23 @@ class AeonReadingRoomsController < ApplicationController
 
   def load_appointments
     @date = (Date.parse(params.expect(:date)) if params[:date]) || Time.zone.now.to_date
-    @available_appointments = @reading_room.available_appointments(@date, include_next_available: true)
+    @available_appointments = available_appointments_without_conflicts
     @available_appointments_on_selected_date = @available_appointments.select { |x| x.start_time.to_date == @date }
     @appointment_lengths = @available_appointments.map(&:maximum_appointment_length)
+  end
+
+  def available_appointments_without_conflicts
+    @reading_room
+      .available_appointments(@date, include_next_available: true)
+      .filter_map { |slot| slot.trimmed_for(own_appointment_conflicts) }
+  end
+
+  def own_appointment_conflicts
+    return [] unless current_user&.aeon
+
+    excluded_id = params[:appointment_id]&.to_i
+    current_user.aeon.appointments
+                .reject { |a| a.id == excluded_id }
+                .map { |a| a.start_time...a.stop_time }
   end
 end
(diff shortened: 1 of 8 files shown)


cbeer:
> user can't be in two places at once

Maybe an all-day appointment as SPEC, briefly interrupted by a 30 minute visit to RUMSEY?

taylor-steve:
Good point, very true. Maybe I shouldn't make it harder for the user than it needs to be anyway? I can dream up other cases (intend to cancel existing EAL, want to book ARS, but I made them do it in order?). It's scoped to the reading room now, which matches our date picker indicator behavior.

app/models/aeon/available_appointment.rb
```diff
@@ -9,5 +9,25 @@ def self.from_dynamic(dyn)
         maximum_appointment_length: dyn['maximumAppointmentLengthMinutes'].minutes
       )
     end
+
+    def trimmed_for(conflicts)
```
cbeer, line 13:
```suggestion
    # given a set of conflicts (e.g. other, existing appointment times), adjust this appointment slot by:
    # a) returning the original unmodified (because it doesn't conflict)
    # b) return nil, because the start of the slot wholly conflicts
    # c) modify the maximum time to the time before the first conflict
    def trimmed_for(conflicts)
```

Wrapping my head around the cases took a little doing; maybe a code comment like the above would help?

app/controllers/aeon_reading_rooms_controller.rb
```diff
@@ -31,8 +31,26 @@ def load_reading_room
...
+    @available_appointments = available_appointments_without_conflicts
     @available_appointments_on_selected_date = @available_appointments.select { |x| x.start_time.to_date == @date }
     @appointment_lengths = @available_appointments.map(&:maximum_appointment_length)
   end
+
+  def available_appointments_without_conflicts
+    @reading_room
+      .available_appointments(@date, include_next_available: true)
+      .filter_map { |slot| slot.trimmed_for(own_appointment_conflicts) }
+  end
```
cbeer, line 43:
```suggestion
  def available_appointments_without_conflicts
  appointment_slots = @reading_room
      .available_appointments(@date, include_next_available: true)
      existing_appointments = current_user.aeon.appointments
      AppointmentDeconflictionService.new(appointment_slots, existing_appointments.reject { |a| a.id == params[:appointment_id] }).call
  end
```

I wonder if extracting a service with the logic currently in the controller + model would help make what we're doing easier to test + reason about?

app/controllers/aeon_reading_rooms_controller.rb
```diff
@@ -31,8 +31,26 @@ def load_reading_room
...
+
+  def own_appointment_conflicts
+    return [] unless current_user&.aeon
```
cbeer, line 46: I /think/ we'll always have current_user.aeon if we got here (either the actual aeon user or the Aeon::NullUser)?

```suggestion
```

app/views/aeon_appointments/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with(model: @appointment, class: "appointment-form", data: { controller: 'appointment', 'appointment-availability-route-value': available_aeon_reading_room_url(id: @appointment.reading_room_id) }) do |f| %>
+<%= form_with(model: @appointment, class: "appointment-form", data: { controller: 'appointment', 'appointment-availability-route-value': available_aeon_reading_room_url(id: @appointment.reading_room_id, appointment_id: (@appointment.id if @appointment.persisted?)) }) do |f| %>
```
cbeer, line 1:
```suggestion
<%= form_with(model: @appointment, class: "appointment-form", data: { controller: 'appointment', 'appointment-availability-route-value': available_aeon_reading_room_url(id: @appointment.reading_room_id, appointment_id: @appointment.id) }) do |f| %>
```

I think `@appointment.id` is nil for a new appointment anyway?

PR #3788, opened 2026-06-10
Stub out the Aeon API so we can write reasonable data round-trip integration tests. 

cbeer:
Writing reasonable tests for the result of updating data is annoying when we're stubbing out all the things. Instead, if we create a stub version of the Aeon API, we can round-trip data and make sure things stay consistent 🤷‍♂️ 

This first pass tries to leave the existing tests alone (as much as possible), but some future refactoring might clean up some of the duplication, wrangle factories to be more consistent (and maybe attempt to unify the local application models with the stub models?), etc.


config/routes.rb
```diff
@@ -97,6 +97,29 @@
   resources :messages
   match 'reports', to: 'reports#index', via: [:get], as: :reports
 
+  namespace :stub_aeon_client do
```
dnoneill: probably want to make this env specific

db/migrate/20260610141652_create_stub_aeon_client_tables.rb
```diff
@@ -0,0 +1,41 @@
+class CreateStubAeonClientTables < ActiveRecord::Migration[8.1]
+  def change
+    create_table :stub_aeon_client_users do |t|
+      t.string :username, index: { unique: true }
```
taylor-steve, line 4: I think this `username` gets shadowed with the data accessor in the stub client model. I think this wins, so we can drop the `username` store in the model.

spec/features/aeon_appointment_spec.rb
```diff
@@ -91,8 +94,8 @@
         expect(page).to have_text 'Change appointment'
         expect(page).to have_text 'Field Reading Room'
         expect(page).to have_text 'Current'
-        expect(page).to have_text appointment.date.strftime('%b %-d, %Y')
-        expect(page).to have_no_text appointment.start_time.strftime('%l:%M %p')
+        # expect(page).to have_text appointment.date.strftime('%b %-d, %Y')
+        # expect(page).to have_no_text appointment.start_time.strftime('%l:%M %p')
```
taylor-steve: Work to be done or no longer needed?

db/migrate/20260610141652_create_stub_aeon_client_tables.rb
taylor-steve, line 1: Did you intend to have this migration here and have both DBs be identical, or does this need to live in `aeon_migrate`? I guess it may not matter, mostly curious.

config/puma.rb
```diff
@@ -28,6 +28,9 @@
 threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
 threads threads_count, threads_count
 
+
+workers ENV.fetch("WEB_CONCURRENCY") { 2 }
```
taylor-steve, line 32: It doesn't look like we don't set `WEB_CONCURRENCY` on prod. Do we need a guard or is it fine?

db/migrate/20260610141652_create_stub_aeon_client_tables.rb
cbeer: I don't know how rails multi-db schemas are supposed to look, but this is what i ended up with 🤷‍♂️

config/puma.rb
cbeer: We're using passenger in production, so I don't think we care 🤷‍♂️

spec/features/aeon_appointment_spec.rb
cbeer: Probably to be done, although I'm also not convinced we need to test things like this in an expensive feature test instead of the component spec?

taylor-steve (approved):

Issue #2853, opened 2026-02-03
Fees and fines


dbranchini:
### Context

[Fees and fines page](https://mylibrary.stanford.edu/fines) exists now in MyLibrary. This is a redesign of that page. Fees/fines are defined in FOLIO. 

### Designs

- [Browser view](https://www.figma.com/proto/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6961-11967&p=f&viewport=534%2C500%2C0.04&t=bdWb2Iw3cy2nYPSu-9&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=6961%3A11967&page-id=3230%3A56325&show-proto-sidebar=1)
- [Design with specifications](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6961-11967&t=3xGSIs66dheXVgMP-4)

### Notes

- Accruing fines are integrated into the list with additional information
- Separate tabs for Current fees (owed) and Past fees (resolved by payment and/or staff waiving the fee) 
- Related Current/Past tab implementation for Activities - #3580 
- Payment is processed by a third-party vendor application

dbranchini:
Sarah informed me that for accruing fines (overdue, but not returned yet), we don't have a current balance. We only know the amount that's accruing, so I modified the design to reflect that.

saseestone:
Moving some details into the ticket - taken from a slack conversation between me & Marlo. The Fines/Fees page is a mix of Folio overdue checkouts and Folio feesAndFines. For overdue checkouts, accruing fines don't show up in Folio feesAndFines. Instead, we need to look at the loan so that we can create the "Accruing..." message on this row from Figma: <img width="1127" height="323" alt="Image" src="https://github.com/user-attachments/assets/00e5127f-f52b-4f60-874f-dd3aa0752f72" />

If the loan is overdue (or overdue because of recalls), then we need to check for an overdue fine policy. There are five overdue fine policies, but two are different versions of "no fine". The two 'no fines" overdue fine policies are "No fines" and "Overdue fine policy". 

Here are the three overdue fine policies we'll need to account for:
* 1.00/30.00 hourly fine: translate to "Accruing $1.00/hour until returned"
* 3.00/21.00 recall overdue fine: translate to "Accruing $3.00/day until returned"
* 1.00/30.00 daily fine: translate to "Accruing $1.00/day until returned" 
[...]

PR #3791, opened 2026-06-10
style fines

dnoneill:
part of #2853 

Before:

before

After:

Screenshot 2026-06-12 at 12 43 09 PM


app/components/fine_component.html.erb
```diff
@@ -1,30 +1,42 @@
...
   <% end %>
   <% c.with_actions do %>
-    <h4><%= number_to_currency(fine.owed) %></h4>
+    <% if accruing_rate_label %>
+      <span class="small fw-medium rounded-pill text-white bg-danger">
+        <%= accruing_rate_label %>
+      </span>
+    <% else %>
+      <div class="h4 mb-0"><%= number_to_currency(fine.owed) %></div>
+    <% end %>
```
taylor-steve: Open question, can a `Folio::Checkout` have a nil `overdue_fines_rate`? It appears `Folio::Checkout` doesn't implement `owed`, so the 'else' would "no method" in that case?

Marlo thought so, or was worried enough he was defensive: https://github.com/sul-dlss/sul-requests/blob/bcd8a836a88d2244b124bfab3ecec4032e675e01/app/models/folio/checkout.rb#L222

app/components/fine_component.html.erb
```diff
@@ -1,30 +1,42 @@
...
-          <% end %>
+    <div class="d-flex flex-row gap-3">
+      <div>Call number: <%= fine.call_number %></div>
+      <%= detail_link_to_searchworks(fine.catkey) unless fine.from_ill? %>
+    </div>
+    <% if fine.library_name %>
+      <div class="d-flex gap-3">
+        <span class="visually-hidden">Library:</span><%= fine.library_name %>
+        <%= link_to "mailto:#{fine.contact_info[:email]}", class: 'su-underline' do %>
+          <i class="align-middle bi bi-envelope me-1"></i>Contact
```
taylor-steve: Similar open question, can `contact_info` be nil? In which case we'd "no method" here. Looks like we're using safe navigation for `contact_info` so possibly?

app/components/fine_component.rb
```diff
@@ -12,29 +12,25 @@ def initialize(fine:, patron:)
     super()
   end
 
-  def render_fine_status
-    fine_status_html(css_class: 'status small fw-medium rounded-pill text-white bg-plum-light ready',
-                     text: fine.status_label)
+  def checked_out?
```
taylor-steve, line 15: I could be reading this wrong, but it seems like `Folio::Checkout` isn't a type of fine. It will result in a fine in the future, but isn't one yet.

Given that, the type check, the two conditionals on `checked_out?`, the conditionals on `accruing_rate_label` in the template all signal to me that these are likely two different components, even at the cost of duplicated markup. We could probably get by with one, although the name seems somewhat misleading to me in that case.

app/components/fine_component.html.erb
```diff
@@ -1,30 +1,42 @@
...
     <h3 class="mb-1 clamp clamp-3"><%= body_title %></h3>
-    <% if fine.bib? %>
-      <div class="d-flex flex-row gap-3">
-        <div>Call number: <%= fine.call_number %></div>
-        <% unless fine.from_ill? %>
-          <%= link_to "#{Settings.searchworks_link}/#{fine.catkey}", rel: 'noopener', target: '_blank' do %>
-            <%= sul_icon :'sharp-open_in_new-24px' %> View in SearchWorks
-          <% end %>
+    <div class="d-flex flex-row gap-3">
+      <div>Call number: <%= fine.call_number %></div>
```
taylor-steve: This was guarded before. Do all things have call numbers?

app/components/fine_component.html.erb
dnoneill: Ah. You are thinking the if should be checkout?. That makes sense to me

app/components/fine_component.rb
dnoneill: They used to be but it was A LOT of repeated markup I was going to have to restyle. I think maybe we could do a separate component for actions only. That would make sense to me.

app/components/fine_component.html.erb
dnoneill: It was guarded with bib? which didn't make sense. bib? was just checking for ['item']. I could add a render? check but this didn't make a ton of sense to me

app/components/fine_component.rb
taylor-steve: To me, repeated markup is almost zero cost. It's grep-able and search and replace-able. State and control frequently remove those traits, because they require me to think in order to understand. I don't like thinking 😄

dnoneill: To me that require me to think more because I have to know that the thing is occurring twice. I am going to apply what I am suggesting and you can say if you don't want that.

app/components/checked_out_actions_component.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def initialize(fine:)
+    @fine = fine
+  end
+
+  def accruing_rate_label
+    rate = fine.overdue_fines_rate
+    return unless rate
+
+    "#{sul_icon('sharp-warning-24px')}Accruing #{number_to_currency(rate['quantity'])}/#{rate['intervalId']} until returned"
```
taylor-steve: This isn't `html_safe` so it gets escaped

Image

 in the `call`. Maybe `safe_join` the icon and the label rather than using string interpolation, or push the icon into the call and let the label just be the label?

app/components/fine_component.html.erb
```diff
@@ -1,30 +1,39 @@
...
-    <%= link_to 'Contact library', contact_path %>
   <% end %>
-  <% if fine.barcode %>
+  <% if fine.barcode || fine.sort_date %>
     <% c.with_footer do %>
-      <span>Barcode: <%= fine.barcode %></span>
+      <div class="rgi-footer fs-14 text-body-secondary d-flex gap-sm-3 flex-sm-row flex-column">
+        <span>Barcode: <%= fine.barcode %></span>
+        <span>Date incurred: <%= l(fine.sort_date.to_date, format: :short) %></span>
+      </div>
```
taylor-steve: Do we need to conditionally render these? I see at least one case in the code where `sort_date` could be empty (not nil) and I see `barcode` could definitely be `nil` (but maybe not in practice?)

app/components/fine_component.html.erb
```diff
@@ -1,30 +1,39 @@
...
+      <%= render PillComponent.new(classes: %w[status bg-plum text-white]).with_content(fine.status_label) %>
+      <% if checked_out? %>
+        <span class="small ms-2">Payment cannot be processed until item is returned</span>
+      <% end %>
     </div>
   <% end %>
   <% c.with_actions do %>
-    <h4><%= number_to_currency(fine.owed) %></h4>
+    <%= render CheckedOutActionsComponent.new(fine:) if checked_out? %>
+    <%= render FineActionsComponent.new(fine:) unless checked_out? %>
```
taylor-steve: if/else saves a conditional here 🤷‍♂️

Since you are asserting in `FineComponent` that both `Folio::Account` and `Folio::Checkout` are types of `fine` and both `CheckedOutActionsComponent` and `FineActionsComponent` take a `fine`, I'd argue:
* These are different things 😄 
* If you still assert they aren't different things, you should push this logic to those component's `render?` method. Both of those smaller components would crash out, given the wrong type with specific data.

app/components/fine_component.html.erb
dnoneill: I am going based on what was in the code to begin with. The original code checked for a barcode. Maybe because there is both we don't have to worry anymore.

app/components/fine_component.html.erb
dnoneill: I mean are you saying https://github.com/sul-dlss/sul-requests/blob/c9c538748ddf3e7b921f03ebd217c7d546e3c895/app/views/fines/index.html.erb#L13-L14 is better than this? If so I can re-implement that and use a slot for the actions

taylor-steve: I'm saying it'd be nice if anything that claims to take a `fine` as an argument could take any object being used elsewhere as a `fine`-able without crashing or the developer having to think about implementation details. Maybe that's as simple as renaming the keyword arg of `CheckedOutActionsComponent` from `fine` to `checkout`.

app/components/checked_out_actions_component.rb
dnoneill: How did you get this view? I can't get it in the tests. I just want to make sure there isn't a problem with the fix

taylor-steve (approved):

PR #3812, opened 2026-06-12
disable save button when all form fields aren't filled out

dnoneill:
part of #3778 

This isn't perfect. There is an issue where the form is filled out but that feels like a separate issue see https://stanfordlib.slack.com/archives/G018TJ20XGE/p1781301841755299. 

https://github.com/user-attachments/assets/e5ab89e8-e231-4ad5-af7e-4a929b0a26ec

### DIFF
--- a/app/controllers/aeon_reading_rooms_controller.rb
+++ b/app/controllers/aeon_reading_rooms_controller.rb
@@ -11,8 +11,6 @@ class AeonReadingRoomsController < ApplicationController
   before_action :load_appointments
 
   def available
-    @selected_time = params[:selected]
-
     respond_to do |format|
       format.html
       format.json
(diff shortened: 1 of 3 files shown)


cbeer:
I wonder if we could accomplish something similar with just a required field in the placeholder turbo-frame?

taylor-steve:
I think there's no rush, take the time to make it perfect 😄 

At first glance I think there's an inconsistency between changing a day only and a non-day appointment.
* Day only enables the save button with no date selected
* Non-day only requires the user to select a date

When I create a new day only appointment and use the keyboard to select the date, the save button does not enable.

dnoneill:
@cbeer Not sure how. I could use more information about that implementation.
@taylor-steve I can't make it perfect with information on how to fix https://stanfordlib.slack.com/archives/G018TJ20XGE/p1781301841755299. Also can you show me `Day only enables the save button with no date selected`. This shouldn't be happening.

PR #3820, opened 2026-06-15
restyle checkouts

### LINKED ISSUE #2852: Borrowed (checked out) items

dnoneill:
closes #2852 
Only thing I haven't done is fix fine may apply to accruing message. Right now it is kind of impossible without the folio data

Before:
Screenshot 2026-06-16 at 3 38 39 PM

After:
Screenshot 2026-06-16 at 3 16 01 PM

Before:
Screenshot 2026-06-16 at 3 38 47 PM

After:
Screenshot 2026-06-17 at 2 58 55 PM

Before:
Screenshot 2026-06-16 at 3 38 42 PM
After:

Screenshot 2026-06-16 at 3 16 15 PM

Before:
Screenshot 2026-06-16 at 3 41 26 PM
Screenshot 2026-06-16 at 3 41 19 PM

After:

Screenshot 2026-06-16 at 3 21 46 PM


app/components/checkout_component.html.erb
```diff
@@ -1,9 +1,10 @@
 <%= render CardComponent.new(element: 'li', data: { 'status-sort-value': checkout.sort_key(:status), 'due-date-sort-value': checkout.sort_key(:due_date), 'title-sort-value': checkout.sort_key(:title), 'author-sort-value': checkout.sort_key(:author), 'call-number-sort-value': checkout.sort_key(:call_number)  }) do |c| %>
   <% c.with_title do %>
     <div class="d-flex flex-row gap-4 align-items-center">
-      <div>Due: <span class="fw-semibold"><%= today_with_time_or_date(checkout.due_date, short_term: checkout.short_term_loan?) %></span></div>
+      <div class="text-nowrap">Due: <span class="fw-semibold"><%= today_with_time_or_date(checkout.due_date, short_term: checkout.short_term_loan?) %></span></div>
 
       <%= checkout_status_pill %>
+      <span class="text-digital-red"><%= header_message %></span>
```
taylor-steve:
```suggestion
      <span class="text-digital-red-dark"><%= header_message %></span>
```

app/components/checkout_component.rb
```diff
@@ -17,35 +17,45 @@ def initialize(checkout:, patron:)
 
   def non_renewable_reason
     return 'Assumed lost' if lost?
-    return 'Another user is waiting' if recalled? || renewal_blocked_by_hold?
+    return 'Renew' if recalled? || renewal_blocked_by_hold?
```
taylor-steve, line 20: Can we remove this line and let the `Renew` in the fallback handle it, or do we need to reorder the conditionals?

app/components/checkout_component.html.erb
```diff
@@ -39,7 +48,6 @@
           <div>Call number: <%= checkout.call_number %></div>
           <%= detail_link_to_searchworks(checkout.catkey) unless checkout.from_ill? %>
         </div>
-        <div class="mb-1 clamp clamp-1"><span class="visually-hidden">Author:</span><%= checkout.author %></div>
```
taylor-steve, line 42: If we're dropping author from the display, I wonder if we should drop author from the sort (e.g., `author-sort-value`)

app/components/checkout_component.html.erb
```diff
@@ -30,6 +31,14 @@
         <%= today_with_time_or_date(checkout.checkout_date, short_term: checkout.short_term_loan?) %>
       </span>
     </div>
+    <% if checkout.proxy_checkout? %>
+      <div>
+        <%= render PillComponent.new(classes: %w[text-white bg-poppy-dark me-1]) do %>
+          <i class="bi bi-key-fill me-1"></i>Proxy
+        <% end %>
+        Borrowed by: <%= proxy_borrower&.display_name %>
```
taylor-steve: It'd be nice to avoid displaying "Borrowed by:" with no name. If a `proxy_borrower` isn't guaranteed to have a display name we should guard the whole thing or maybe there's a fallback we could provide.

app/components/checkout_component.rb
dnoneill: We would have to re-order the conditionals which I am not inclined to do. Without this line it says `Too soon to renew`

dnoneill: I guess I could do too_soon_to_renew? && (!recalled? || !renewal_blocked_by_hold?) but I am not inclined to do this since we just had a discussion that all of this logic will probably change and I would like to keep it pretty close to what it was.

app/components/checkout_component.html.erb
dnoneill: That would make sense @dbranchini do you have an opinion?

app/components/checkout_component.html.erb
dnoneill: I wasn't under the impression a user could not have at least a first name but yeah we could fallback to the library_id or email?

dnoneill: I went with library_id considering the folio data seems to include a barcode in the spec tests

app/components/checkout_component.html.erb
cbeer: This seems like it's still unresolved. I'm surprised we're dropping author at all, so I think it's important to check.

dbranchini:
Let's keep the author. Sorry about the confusion on this.

dnoneill:
@dbranchini Where do you want it to go? Before call number? A new line?

taylor-steve:
> @dbranchini Where do you want it to go? Before call number? A new line?

Darcy might be in Dataworks land for a while, so I'm not sure how fast she'll respond. I'd be happy putting it where it was for now, the line after call number, and we can pick it up in a click-through/design review if she doesn't like that.

app/models/folio/patron.rb
```diff
@@ -153,7 +153,7 @@ def proxies
       @proxies ||= proxies_of_response.filter_map do |info|
         next nil unless valid_proxy_relation?(info)
 
-        self.class.new(info['proxyUser']) if info['proxyUser'].present?
+        self.class.new(info['proxyUser'].merge({ 'id' => info['proxyUserId'] })) if info['proxyUser'].present?
```
taylor-steve: Chris is working on the proxy issue and thinks the fixture data may be incomplete. Can we roll this back and if we find out we really do need it, do a one line PR after?

dnoneill: Yeah that works. I can save the code for when we need it.

dnoneill: Actually I left the front end code and a method that needs to be filled out when that data is avaliable

PR #3825, opened 2026-06-16
Disable existing day only appointments in date picker

taylor-steve:
@dnoneill what do you think about pushing https://github.com/sul-dlss/sul-requests/pull/3775 a bit further like this?

I was trying to convert https://github.com/sul-dlss/sul-requests/pull/3790 after we went a different direction with `AppointmentDeconflictionService`, but I came to the conclusion that it's not worth the effort to pre-compute "fully booked" non-day-only reading room appointments (a rare case), so I think you're correct we should just handle the day only disables for now.

We've already got domain specifics in the date picker, so I feel like we should go one way or another. If we're keeping it Aeon specific, I like bringing `marked_dates` inside.

The non-day-only case would be better handled by live availability, which I'll poke at next.


dnoneill:
@taylor-steve I am not sure why this is better than #3775. What are we getting? It seems like we are passing a variable that isn't really getting used all the time to do that same thing as the other PR

PR #3828, opened 2026-06-16
Access appointment attributes through the appointment.

cbeer:


taylor-steve:
I'm seeing "Create appointment" in the context of a new request no longer pre-select the reading room.

cbeer:
Hm. I'm surprised the test isn't catching that.

taylor-steve (approved):

PR #3832, opened 2026-06-16
Fetch reading room availability in date picker

taylor-steve:
If you're quick you'll see a placeholder loading indicator:
Screenshot 2026-06-16 at 3 13 51 PM

Here's one where the 25th gets disabled as a result:
Screenshot 2026-06-16 at 3 34 01 PM

* Cache Aeon `ReadingRooms` response for 5 minutes (open hours, policies)
* ~~Provide a way to get `User#appointments` without augmenting them with requests. Nice when we need it, but it adds overhead to the availability calls (10+ seconds for me :cry:)~~ Fixed by https://github.com/sul-dlss/sul-requests/pull/3834
* Pull the closed days logic out of the date picker component and into the reading room controller
* Let the reading room controller tell us unavailable dates for a given month for the user
* Eagerly fetch the current month's and next month's unavailable dates. Fetch 1 subsequent month as the user arrives on the month prior

It's not _fast_ (2-3 seconds) but it starts the initial fetch before the user has opened the picker, so in practice it feels fine to me.

To give you an idea:
```
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-17 -> 200 (85ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-18 -> 200 (164ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-21 -> 200 (85ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-22 -> 200 (100ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-23 -> 200 (87ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-24 -> 200 (116ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-25 -> 200 (105ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-28 -> 200 (86ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-29 -> 200 (91ms)
[AeonClient] GET ReadingRooms/1/AvailableAppointments/2026-09-30 -> 200 (100ms) 
```

### DIFF
--- a/app/controllers/aeon_reading_rooms_controller.rb
+++ b/app/controllers/aeon_reading_rooms_controller.rb
@@ -7,7 +7,7 @@ class AeonReadingRoomsController < ApplicationController
   include AeonController
 
   before_action :load_reading_room
-  before_action :load_appointments
+  before_action :load_appointments, only: :available
 
   def available
     @selected_time = params[:selected]
@@ -18,7 +18,33 @@ class AeonReadingRoomsController < ApplicationController
     end
   end
 
+  def unavailable_dates
+    month = Date.strptime(params.expect(:month), '%Y-%m')
+    dates = bookable_dates_in(month).reject do |date|
+      deconflict(@reading_room.available_appointments(date)).any?
+    end
+    render json: { month: params[:month], unavailable_dates: dates.map(&:iso8601) }
+  end
+
   private
+
+  def bookable_dates_in(month)
+    closed = @reading_room.fully_closed_dates.to_set
+    month.all_month.select { |date| @reading_room.open_hours_on(date) && closed.exclude?(date) }
+  end
+
+  def deconflict(available_appointments)
+    Aeon::AppointmentDeconflictionService.new(
+      available_appointments:,
+      existing_appointments: existing_appointments_for_deconfliction
+    ).call
+  end
+
+  def existing_appointments_for_deconfliction
+    current_user.aeon.appointments
+                .for_reading_room(@reading_room)
+                .reject { |a| a.id == params[:appointment_id].to_i }
+  end
 
   def load_reading_room
     @reading_room = Aeon::ReadingRoom.find(params.expect(:id))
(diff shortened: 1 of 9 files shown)


dnoneill:
@taylor-steve so you are thinking everything is enabled not disabled? I am asking because I think when I have seen this all the days on the calendar is disabled and then the one's you can choose are enabled. I don't love that but I am not sure I love users being able to click things that will get disabled. I guess I would like more information on how fast users are clicking on dates.

taylor-steve:
Yes that's correct. Closures are disabled at render time, it's only self-conflicts or fully booked reading rooms that are disabled dynamically.

These cases are the exception as I understand it, reading rooms generally have spots available.

Given that, I prefer the optimistic approach, rather than forcing the user to wait. In the event the user is quicker than the availability check, they would see what they do currently, that there are no available appointments for the day they selected.

It's definitely not great, but it's the best idea I have at the moment, given the tools available to us. Darcy may have a different opinion and/or better ideas.

dnoneill:
Yeah I think the only point I was making is if users are quick we are doing work on the front end they will never see because I would assume (might be wrong) that most users know which date they want when they are scheduling an appointment. In which case it seems like a waste of processing power for something they are unlikely to see. I think at this point we have a lot of ideas around date picker/appointment. I think maybe we should have a quick talk with Darcy about the expected/wanted behavior and figure out where to go from there.

taylor-steve:
Agreed. This is why we discussed it in stand-up with Darcy (who felt okay with this plan but hasn't seen in in action) and I asked to hold this PR until she could see it for herself.

dnoneill:
needs rebase

app/controllers/aeon_reading_rooms_controller.rb
```diff
@@ -18,23 +18,42 @@ def available
...
+      deconflict(@reading_room.available_appointments(date)).any?
+    end
+    render json: { month: params[:month], unavailable_dates: dates.map(&:iso8601) }
+  end
+
   private
 
+  def bookable_dates_in(month)
+    closed = @reading_room.fully_closed_dates.to_set
+    month.all_month.select { |date| @reading_room.open_hours_on(date) && closed.exclude?(date) }
```
dnoneill, line 33: Do we want to create a method for full day appointments to add to this? I know it won't save that much time but it might be nice. Something like 
```suggestion
month.all_month.select { |date| @reading_room.open_hours_on(date) && non_bookable_days.exclude?(date) }

def non_bookable_days
  return @reading_room.fully_closed_dates.to_set unless @reading_room.day_only_appointments?
  
  @reading_room.fully_closed_dates.to_set + existing_appointments_for_deconfliction
end
```


I really don't love this. Maybe it should stay as is.

dnoneill (approved):

PR #3839, opened 2026-06-17
Pass users into the activity so we can lazy-load requests

cbeer:


dnoneill (dismissed):
This ends up slowing down all other page loads. Should probably remove this conditional https://github.com/sul-dlss/sul-requests/blob/0f530bcc420940df9cb3f90aba16b646873afadd/app/views/layouts/application_redesign.html.erb#L64

app/models/aeon/activity.rb
```diff
@@ -47,7 +47,7 @@ def requests=(requests)
     end
 
     def requests
-      @requests ||= []
+      @requests ||= Aeon::RequestFinders.new(users.flat_map { |u| u.all_requests.for_activity(self).submitted })
```
taylor-steve: I don't think it's your job to fix this here, but noting that this bypasses our sorting hack in the setter method above

cbeer:
Can you say more? I don't think this introduces any new API calls, so I'm not sure I understand what slow-down you're suggesting this creates.

taylor-steve (approved):
One other thing I'll note, and I'm just complaining about Aeon here and/or misreading things, is I think we've got a tangled web:

`current_user.activities.first.requests` will use the `@all_requests` memoization
`request.activity.requests` won't have the shared user instances, so we'd we'd be doing multiple `requests_for` calls for the same user

I think you've actually made this more correct here by removing the `[]` default, and nothing makes a call like this now, so it's fine. Just leaving breadcrumbs for my future self when I'm very confused.

dnoneill:
It doesn't introduce new api calls but it does introduce new api calls on activities. When you are loading requests in activities it does create new api calls for the other users. I specifically separated out these methods into separate methods because the API calls for users https://github.com/sul-dlss/sul-requests/blob/ad6768d98128b7435e50c7b9625e2d5293eafda2/app/models/aeon/user.rb#L37 wasn't on activities before. And when I had it there it slowed down the loading on all the pages.

cbeer:
@dnoneill : Can you show me where User#activities calls User#all_requests? The whole goal of this refactor was to delay that requests except when Activities#requests is called.

dnoneill:
@cbeer Wouldn't it be called by https://github.com/sul-dlss/sul-requests/blob/0f530bcc420940df9cb3f90aba16b646873afadd/app/views/layouts/application_redesign.html.erb#L64? Am I misunderstanding the code? Oh. I think I understand now. Requests not load until the user calls aeon activities.request

PR #3841, opened 2026-06-17
Update Folio::LoanPolicy to just be the policy and not include any checkout information

cbeer:
Intermingling the checkout data and the policy data makes updating the checkout after renewals a little annoying.

### DIFF
--- a/app/models/folio/checkout.rb
+++ b/app/models/folio/checkout.rb
@@ -6,11 +6,7 @@ module Folio
     include Folio::FolioRecord
 
     attr_reader :record, :patron_type_id
-
-    delegate :too_soon_to_renew?,
-             :renewal_blocked_by_hold?,
-             :unseen_renewals_remaining,
-             to: :loan_policy
+    attr_writer :loan_policy
 
     delegate :loan_policy_interval,
              to: :loan_policy,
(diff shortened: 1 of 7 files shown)


dnoneill (changes requested):
https://github.com/sul-dlss/sul-requests/blob/3210e2985cbc68d6a2e91f92eb358fe02b81d853/app/components/checkout_component.rb#L23 needs to be updated with a due date

dnoneill (changes requested):
This has also broken all the lookbooks for checkout  http://localhost:3000/lookbook/inspect/checkout/default.

1. I think what needs to happen is https://github.com/sul-dlss/sul-requests/blob/3210e2985cbc68d6a2e91f92eb358fe02b81d853/spec/factories/folio.rb#L39 needs to be updated to ```initialize_with { new(record, **attributes.except(:record)) }```

2. Also you need to add mocks to the componentview. 
```
checkout = FactoryBot.build(:checkout, loan_policy: FactoryBot.build(:grad_mono_loans))
checkout.define_singleton_method(:effective_loan_policy_id) do
  ''
end
```

3. Additionally, the `renewal_blocked_by_hold?` needs to be moved out of private or fixed in the CheckoutComponent.

dnoneill (approved):

PR #3844, opened 2026-06-17
add admin menu

dnoneill:
closes #3842 
| Figma | PR |
|--------|-------|
| | Screenshot 2026-06-18 at 1 51 06 PM|

### DIFF
--- a/app/components/admin_nav_dropdown_component.html.erb
+++ b/app/components/admin_nav_dropdown_component.html.erb
@@ -1,0 +1,24 @@
+<% masthead_component.with_nav_item(classes: ['dropdown']) do %>
+  <button class="nav-link dropdown-toggle fw-semibold bg-muted-cardinal px-2 rounded-1" data-bs-toggle="dropdown" aria-expanded="false">
+    Administration
+  </button>
+  <ul class="dropdown-menu">
+    <% if site_admin? %>
+      <li><span class="h6 dropdown-header text-cardinal">APPLICATION</span></li>
+      <li><%= link_to 'Global messages', messages_path, class: 'dropdown-item' %></li>
+      <li><span class="h6 dropdown-header text-cardinal">REQUESTS</span></li>
+      <li><%= link_to 'Dashboard', admin_index_path, class: 'dropdown-item' %></li>
+      <li><%= link_to 'Job queue', main_app.sidekiq_web_path, class: 'dropdown-item' %></li>
+      <li><%= link_to 'Paging schedule', paging_schedule_index_path, class: 'dropdown-item' %></li>
+      <li><%= link_to 'Reports', reports_path, class: 'dropdown-item' %></li>
+    <% end %>
+    <% if mediated_locations? %>
+      <% mediated_locations.each do |code, config| %>
+        <li><span class="h6 dropdown-header text-cardinal">REQUESTS MEDIATION</span></li>
+        <li>
+          <%= link_to(LibraryLocation.library_name_by_code(config&.library_override || code), admin_path(code), class: 'dropdown-item') %>
+        </li>
+      <% end %>
+    <% end %>
+  </ul>
+<% end %>
--- a/app/components/admin_nav_dropdown_component.rb
+++ b/app/components/admin_nav_dropdown_component.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# Admin menu in masthead
+class AdminNavDropdownComponent < ViewComponent::Base
+  attr_reader :masthead_component
+
+  delegate :mediated_locations_for, :main_app, :can?, to: :helpers
+  def initialize(masthead_component:)
+    @masthead_component = masthead_component
+  end
+
+  def render?
+    mediated_locations? || site_admin?
+  end
+
+  def site_admin?
+    can?(:manage, :site)
+  end
+
+  def mediated_locations?
+    mediated_locations.present?
+  end
+
+  def mediated_locations
+    mediated_locations_for(PatronRequest.mediateable_origins)
+  end
+end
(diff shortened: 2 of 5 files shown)


taylor-steve:
Ref https://github.com/sul-dlss/sul-requests/issues/3842

Screenshot 2026-06-17 at 4 00 07 PM

app/views/layouts/application_redesign.html.erb
```diff
@@ -72,6 +72,25 @@
...
+          <ul class="dropdown-menu">
+            <li><span class="h6 dropdown-header text-digital-red">APPLICATION</span></li>
+            <li><%= link_to 'Global messages', messages_path, class: 'dropdown-item' %></li>
+            <li><span class="h6 dropdown-header text-digital-red">REQUESTS</span></li>
+            <li><%= link_to 'Dashboard', admin_index_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Job queue', main_app.sidekiq_web_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Paging schedule', paging_schedule_index_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Reports', reports_path, class: 'dropdown-item' %></li>
+            <li><span class="h6 dropdown-header text-digital-red">REQUESTS MEDIATION</span></li>
+            <!-- TODO: Confirm with Sarah these are the correct codes -->
```
taylor-steve: Is this still a TODO or did you get what you needed from Darsi?

cbeer: wait -- I don't think we should hard-code  these at all; is there a reason we can't use the existing approach to display them dynamically?

dnoneill: I just found the code after looking for a very long time for it, it showed up for me while I was doing something else. Pushed up.

app/views/layouts/application_redesign.html.erb
```diff
@@ -72,6 +72,26 @@
           <%= link_to 'Fees/fines', fines_path, class: 'nav-link fw-semibold' %>
         <% end if current_user.patron.present? %>
 
+        <% component.with_nav_item(classes: ['dropdown']) do %>
+          <button class="nav-link dropdown-toggle fw-semibold bg-muted-cardinal px-2 rounded-1" data-bs-toggle="dropdown" aria-expanded="false">
+            Administration
+          </button>
+          <ul class="dropdown-menu">
+            <li><span class="h6 dropdown-header text-digital-red">APPLICATION</span></li>
```
cbeer: I think these are digital red after Darcy's comment about the searchworks menu, but searchworks uses cardinal red?

app/views/layouts/application_redesign.html.erb
```diff
@@ -72,6 +72,26 @@
...
+          </button>
+          <ul class="dropdown-menu">
+            <li><span class="h6 dropdown-header text-cardinal">APPLICATION</span></li>
+            <li><%= link_to 'Global messages', messages_path, class: 'dropdown-item' %></li>
+            <li><span class="h6 dropdown-header text-cardinal">REQUESTS</span></li>
+            <li><%= link_to 'Dashboard', admin_index_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Job queue', main_app.sidekiq_web_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Paging schedule', paging_schedule_index_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Reports', reports_path, class: 'dropdown-item' %></li>
+            <li><span class="h6 dropdown-header text-cardinal">REQUESTS MEDIATION</span></li>
```
taylor-steve: Do we want to have this sort of section header included in the 'can manage' guard?

Image

dnoneill: I considered it but I figured there isn't any reason to cause I don't think there is anywhere on stage/prod that doesn't have at least one location but I can add

taylor-steve: Ah thanks, I wasn't familiar enough with the particulars

app/views/layouts/application_redesign.html.erb
```diff
@@ -72,6 +72,26 @@
...
+            <li><%= link_to 'Paging schedule', paging_schedule_index_path, class: 'dropdown-item' %></li>
+            <li><%= link_to 'Reports', reports_path, class: 'dropdown-item' %></li>
+            <% mediated_locations_for(PatronRequest.mediateable_origins).each do |code, config| %>
+              <li><span class="h6 dropdown-header text-cardinal">REQUESTS MEDIATION</span></li>
+              <li>
+                <%= link_to(LibraryLocation.library_name_by_code(config&.library_override || code), admin_path(code), class: 'dropdown-item') %>
+              </li>
+            <% end %>
+          </ul>
+        <% end if can? :manage, :site %>
```
cbeer: It looks like this whole menu is now behind `can? :manage, :site`, but there are location-specific mediators who don't have that power and need access to their mediated locations still.

dnoneill: Okay should be 🤞 totally fixed. I moved it into a component because the ifs and origin seems like it would be better in a component.

app/components/admin_nav_dropdown_component.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# Admin menu in masthead
+class AdminNavDropdownComponent < ViewComponent::Base
+  attr_reader :masthead_componet
+
+  delegate :mediated_locations_for, :main_app, :can?, to: :helpers
+  def initialize(masthead_componet:)
+    @masthead_componet = masthead_componet
```
cbeer:
```suggestion
  attr_reader :masthead_component

  delegate :mediated_locations_for, :main_app, :can?, to: :helpers
  def initialize(masthead_component:)
    @masthead_componet = masthead_component
```

app/components/admin_nav_dropdown_component.html.erb
```diff
@@ -0,0 +1,24 @@
+<% masthead_componet.with_nav_item(classes: ['dropdown']) do %>
```
cbeer:
```suggestion
<% masthead_component.with_nav_item(classes: ['dropdown']) do %>
```

PR #3845, opened 2026-06-17
Add proxy group support

cbeer:
#3831

Slightly out of date screenshots:
https://github.com/user-attachments/assets/1fadf78e-f7c6-44dc-b0f4-7fbd07c16276

Screenshot 2026-06-18 at 09 05 35

TODO:
- [ ] figure out what to do on non-FOLIO pages (preferably, don't show the bar at all so we're not making some FOLIO requests to determine if we need to show a bar that doesn't do anything anyway)
- [ ] figure out what to do for patron_requests#new (which already has form elements for requesting for a sponsor)
- [ ] decide if we're ok making somewhat useless FOLIO requests (e.g. that retrieve the proxy's checkouts even if we're looking at the sponsor view)
- [x] move some of the `Folio::Patron#proxy_` methods over to `Folio::ProxyGroup`
- [x] Proxy group member list modal
- [x] get the colors right


app/components/modal_component.rb
```diff
@@ -5,14 +5,16 @@ class ModalComponent < ViewComponent::Base
   attr_reader :id, :data
 
   renders_one :title
+  renders_one :subtitle
```
dnoneill: Cruft?

app/components/proxy_masthead_component.html.erb
```diff
@@ -0,0 +1,79 @@
+<div class="proxy-masthead <%= personal? ? 'bg-light' : 'bg-warning-subtle bg-opacity-10' %> border-bottom border-3 border-cardinal py-2">
+  <div class="container d-flex align-items-center gap-2">
+    <% if personal? %>
+      <%= render PillComponent.new(classes: ['bg-cardinal text-light me-1 text-nowrap']) do %>
+        <i class="bi bi-person-vcard"></i> Personal
+      <% end %>
+
+    <span class="fw-semibold"><%= selected.display_name %></span>
```
dnoneill: Should be tabbed to be inline with pill component I believe?

dnoneill:
All small things but here a couple discrepancies in terms of view:

| Problem | Figma | PR |
|---------|--------|----|
| [x] Proxy group modal underline is wrong color (fine on hover) |  Screenshot 2026-06-19 at 12 50 36 PM | Screenshot 2026-06-19 at 12 50 28 PM |
| [x] Switch dropdown seems to be wrong color and size |  Screenshot 2026-06-19 at 12 50 47 PM | Screenshot 2026-06-19 at 12 50 44 PM |
| [x] Line color on proxy is wrong | Screenshot 2026-06-19 at 12 55 37 PM | Screenshot 2026-06-19 at 12 50 56 PM|
| [x] Spacing on proxy members is off | Screenshot 2026-06-19 at 12 51 50 PM | Screenshot 2026-06-19 at 12 51 46 PM|

taylor-steve (approved):

PR #3849, opened 2026-06-18
add modal feedback form and update feedback form

dnoneill:
closes #3830 
| Figma | PR |
|--------|-------|
| Screenshot 2026-06-18 at 4 36 03 PM | Screenshot 2026-06-19 at 11 28 46 AM |
| Screenshot 2026-06-18 at 4 35 44 PM |  Screenshot 2026-06-19 at 11 28 38 AM |
|Screenshot 2026-06-18 at 4 36 25 PM| Screenshot 2026-06-19 at 11 24 30 AM |


app/views/feedback_forms/_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <h2 class="modal-title order-1 me-auto" id="<%= component.title_id %>">Send Feedback</h2>
+      <%= link_to feedback_form_path, target: '_blank', class: "order-3 order-sm-2 new-tab-link mt-2 mt-sm-0 me-sm-3 su-underline" do %>
+        <i class="bi bi-box-arrow-up-right me-1"></i>Open form in new tab
+      <% end %>
+      <button type="button" class="btn-close order-2 order-sm-3 ms-auto ms-sm-0" data-bs-dismiss="modal" aria-label="Close"></button>
+    </div>
+  <% end %>
+  <% component.with_banner do %>
+    <div class="bg-stanford-10-black py-1 px-3">
+      Reporting from: <span class="reporting-from-field"><%= root_url %></span>
```
taylor-steve: Niqui, I don't know if this is a you question or a @dbranchini question. I'm curious why we're showing the `root_path` to the user and the `fullpath` to the form. The form makes sense, I expected the UI side to match, but maybe that's wrong.

app/views/feedback_forms/_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <%# TODO_SW_2024: update to url that is not request.fullpath and is root_url %>
+    </div>
+  <% end %>
+  <% component.with_body do %>
+    <%= render 'feedback_forms/form' %>
+  <% end %>
+  <% component.with_footer do %>
+    <%= link_to "Close", "#feedback-form", class:"btn btn-outline-primary", data: { bs_dismiss: "modal", bs_target: "#feedback-form"} %>
+    <button type="submit" class="btn btn-primary">Submit</button>
```
taylor-steve: Could you check the submit button? I'm seeing it disconnected from the form currently and cannot submit.

app/views/feedback_forms/_modal.html.erb
dnoneill: I have no idea. This is how it was being done previously. I kept as much HTML as I could from the old code https://github.com/sul-dlss/sul-requests/blob/main/app/views/feedback_forms/new.html.erb#L10-L12. I am not quite sure why it was done this way and I definitely don't understand the comment.

dnoneill: It looks like I put it in there. Blows my mind. No idea.

dnoneill: Looks like we are using request.referer everywhere else. I can put that in.

dnoneill: Actually won't use that. At least locally it gives the previous page which is not good.

app/views/feedback_forms/_modal.html.erb
dnoneill: Oh It would be. Let me fix that.

app/views/feedback_forms/_modal.html.erb
dnoneill: Oh I think I understand the comment. I think we need to double check with Darcy. What url we want to send to staff. Does it matter what page they were on when they sent feedback.

taylor-steve: I think it's a good thing we're sending the full URL to staff. I'm curious if it's intentional we're showing "Reporting from: https://requests.stanford.edu/` on all pages

dnoneill: I think it was. And I agree which is why I think there is a note and not an actual implementation. I think that was a get back to me question. So I think this is Darcy/UX team question

dnoneill: It looks like referrer is used in searchworks and it looks like locally it is the only time when the url is wrong.

dnoneill: I am going to go with what SearchWorks has because that is a new implementation but that can be a double check/clean up question

app/views/feedback_forms/_modal.html.erb
dnoneill: K. This is fixed. I forgot about the weirdness we have to do with modal forms.

taylor-steve:
On form submit, I'm seeing it redirect me to the home page regardless of where I initiated the feedback. Could we stay on the same page?

dnoneill:
@taylor-steve I think that has to do with how request.referer is weird locally. The code indicates it shouldn't do that: 
https://github.com/sul-dlss/sul-requests/pull/3849/changes#diff-acf80438160591bb62755720ccd195013ac123df0547b89be487d088a44c022dR9

https://github.com/sul-dlss/sul-requests/blob/fc35e4bb99ac080f1b0fb9b638b298d994242f63/app/controllers/feedback_forms_controller.rb#L14

taylor-steve:
Hrm, it happens on dev too. If you're out of ideas I can look, but it might not be for a while.

Screenshot 2026-06-19 at 9 33 32 AM
Screenshot 2026-06-19 at 9 33 44 AM

dnoneill:
Hmm. No I wasn't aware of that issue. I will dig in

dnoneill:
@taylor-steve What I am seeing on dev is the same as locally. This is the request.referer issues I was seeing locally but doesn't seem to be an issue in Searchworks prod. request.referer will show the the previous page url and redirect you there. My understanding of request.referer has been that it is the previous page but this is what we are using literally everywhere. I am happy to go back to fullpath but we will need to change all of our forms because that is what they are using. Also it doesn't seem to be an issue in SearchWorks prod which is super weird to me. Actually they might be hitting fetching the url for the modal so it would make sense then, except we are using it in PURL and that works just fine too and it's embedded into the webpage.

https://github.com/user-attachments/assets/59b87ba9-a398-4421-8e4d-95ed7ab2a385

taylor-steve:
Ah okay if it works in prod, that's fine. Thanks for looking into it.

app/views/feedback_forms/_form.html.erb
```diff
@@ -0,0 +1,37 @@
+<% close_button = local_assigns.fetch(:close_button, true) %>
+<%= form_tag feedback_form_path, method: :post, class:"feedback-form" do %>
+  <div class="modal-body">
+    <span class="d-none">
+      <%= label_tag(:email_address, 'Ignore this text box. It is used to detect spammers. If you enter anything into this text box, your message will not be sent.') %><br>
+      <%= email_field_tag :email_address %><br>
+      <%= hidden_field_tag :user_agent %>
+      <%= hidden_field_tag :viewport %>
+      <%= hidden_field_tag :url, request.referer, class:"reporting-from-field" %>
```
taylor-steve, line 9: Remind me again, are we using `request.referer` for a specific reason or is it simply copying convention from other repos? This doesn't make any sense to me over something like `request.fullpath`, but maybe I'm missing something. If I type the url into my browser myself, `request.referer` will be nil which doesn't seem like what we want.

cbeer: I think with the modal, the current path is the feedback modal?

taylor-steve: That'd make sense for the hidden field, the display IDK.

taylor-steve (approved):

app/views/feedback_forms/_form.html.erb
dnoneill: It was copying convention from everywhere else. I am still not quite sure why it works on prod.

PR #3853, opened 2026-06-18
Guard against a missing sortable list target (e.g. in the zero items …

cbeer:
…case)


dnoneill:
When does this case occur?

cbeer:
I saw it on the fees + payments pages.

taylor-steve (approved):

PR #3856, opened 2026-06-18
Don't shadow Folio::FolioRecord#location

cbeer:


app/models/folio/checkout.rb
```diff
@@ -190,10 +190,6 @@ def contact_info
       location&.contact_info
```
dnoneill, line 190: This needs to get removed too if location isn't in checkout anymore. I am a little confused why tests aren't breaking. Maybe we don't have a test for this or there is some magic I am unfamiliar with going on?

cbeer: location is in checkout via the Folio::FolioRecord mixin. We don't need it twice.

dnoneill: Ah so sort of magic. Thank you.

dnoneill (approved):

dnoneill (approved):

PR #3857, opened 2026-06-18
fix proxy and unskip tests

dnoneill:


app/models/folio/patron.rb
```diff
@@ -153,7 +153,7 @@ def proxies
       @proxies ||= proxies_of_response.filter_map do |info|
         next nil unless valid_proxy_relation?(info)
 
-        self.class.new(info['proxyUser']) if info['proxyUser'].present?
+        self.class.new(info['proxyUser'].merge({'id' => info['proxyUserId']})) if info['proxyUser'].present?
```
cbeer, line 156: Is this an artifact of bad test data? I'm not seeing any need for this with real-world data.

dnoneill: It must be bad test data then. My understanding of real world data is based on test data. The real world data doesn't include the id. https://github.com/sul-dlss/sul-requests/blob/d2ad4088cbc3b66629ae88b8c6f4948d2a99f8d6/spec/factories/mylibrary_patrons.rb#L22-L31. Feel free to adjust to what it should look like but I am flying blind. If you have some real data I would appreciated it.

dnoneill: Ah I see https://github.com/sul-dlss/sul-requests/pull/3845, which looks like it includes updated data and un-skips the tests. So I am going to close this.

Issue #3847, opened 2026-06-18
Sponsor view on list pages (borrowed items, fees/fines, requests)


dbranchini:
### Context
Faculty and researchers (sponsors) can assign proxies to take actions on their behalf. Sponsors should be able to identify when an action associated with their account was performed by a proxy.

Rather than creating a separate proxy activity view, we will display an indicator on relevant records showing when the action was performed by a proxy.

### Designs
- [Design with specifications](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=6961-12498&t=w07BMeGKkSNcdlcs-4)

### Preview
<img width="1988" height="530" alt="Image" src="https://github.com/user-attachments/assets/be57e18d-f6eb-42c3-baf9-6f8d3267f322" />

### Borrowed items
  **Display:** Borrowed by: [Proxy Name]
  
  This should appear when an item was checked out by a proxy on behalf of the sponsor.
  
  **Open question:** My understanding is that FOLIO only records who borrowed an item and does not retain proxy information for subsequent actions such as renewals or returns. Therefore, even if an item was subsequently renewed by a proxy (same or someone else), we can only show the original borrower information.

### Requests (FOLIO only)

  **Display:** Requested by: [Proxy Name]
  
  This should appear when a request was submitted by a proxy on behalf of the sponsor.
  
  **Open question:** Please confirm that FOLIO (or our own systems) capture proxy attribution for requests.

### Fees and fines

  *Display:** Borrowed by: [Proxy Name]
  
  This might help sponsors understand which proxy activity resulted in the charge. However, this is an example where who returned it might be more helpful (especially if it was returned late or damaged).
  
  **Open question:** Let's discuss whether this information is available on the fee/fine and whether it provides sufficient value to warrant displaying it in the fees and fines experience.

dnoneill:
Most of this is done in https://github.com/sul-dlss/sul-requests/pull/3820. It just needs to be added to the correct records and we need to figure out how proxies are built.

dnoneill:
@dbranchini There is no footer on the folio-requests. Where is the proxy information supposed to go?

dbranchini:
I wrote this as "Sponsor view," but after discussing this with Chris it makes sense that this information is available to users in a proxy group too.

dbranchini:
Let's try this for requests, @dnoneill:

[...]

PR #3860, opened 2026-06-19
add proxy messaging to fines

dnoneill:
part of #3847 

I am not sure about requests right now since it doesn't have a footer.

### DIFF
--- a/app/components/checkout_component.rb
+++ b/app/components/checkout_component.rb
@@ -50,7 +50,7 @@ class CheckoutComponent < ViewComponent::Base
   def proxy_borrower
     return nil unless checkout.proxy_checkout?
 
-    patron.proxies.find { |proxy| proxy.id == checkout.patron_key }
+    patron.proxy_user(checkout.patron_key)
   end
 
   def checkout_status_pill
--- a/app/components/fine_component.html.erb
+++ b/app/components/fine_component.html.erb
@@ -37,5 +37,6 @@
       <% if fine.barcode %><span>Barcode: <%= fine.barcode %></span><% end %>
       <% if fine.sort_date %><span>Date incurred: <%= l(fine.sort_date.to_date, format: :short) %></span><% end %>
     </div>
+    <%= render ProxyIndicatorComponent.new(proxy: proxy_borrower) %>
   <% end %>
 <% end %>
--- a/app/components/fine_component.rb
+++ b/app/components/fine_component.rb
@@ -2,14 +2,21 @@
 
 # Render a single fine or payment for a patron
 class FineComponent < ViewComponent::Base
-  attr_reader :fine, :sortable
+  attr_reader :fine, :sortable, :patron
 
   delegate :detail_link_to_searchworks, to: :helpers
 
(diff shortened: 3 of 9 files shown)


app/components/checkout_component.html.erb
```diff
@@ -31,7 +31,7 @@
         <%= today_with_time_or_date(checkout.checkout_date, short_term: checkout.short_term_loan?) %>
       </span>
     </div>
-    <%= render ProxyIndicatorComponent.new(proxy: proxy_borrower) %>
+    <%= render ProxyIndicatorComponent.new(proxy_id: checkout.patron_key) if checkout.proxy_checkout? %>
```
cbeer: Hm, I really liked your initial idea of passing in the actual patron object. Having the `ProxyIndicatorComponent` do the data lookup feels a little surprising.

app/components/proxy_indicator_component.rb
```diff
@@ -2,10 +2,16 @@
...
 
-  def initialize(proxy:)
-    @proxy = proxy
+
+  def initialize(proxy_id:)
+    @proxy_id = proxy_id
+  end
+
+  def proxy
+    current_user.patron.proxy_user(proxy_id)
```
cbeer: If there's a reason to do this refactor, `current_user.patron` is definitely wrong for a proxy user viewing what other proxy users have done (we have to go through the sponsor)?

dnoneill: My understanding of this is it is only supposed to show up for the sponsor so they can see what the proxies are doing. It was not under the impression that proxies should be able to see what other proxies are doing https://github.com/sul-dlss/sul-requests/issues/3847

app/components/checkout_component.html.erb
dnoneill: I was on the fence. I ended up having the exact same code in the both components for the proxy users so it felt like it should go into the proxyindicatorcomponent but I move it back into the respective components.

app/components/proxy_indicator_component.rb
cbeer: We can check with Darcy, but that would be a significant change from the current behavior so it'd surprise me if that was the case.

dnoneill: Okay. I think the updated code should work for both.

cbeer: I confirmed we're not changing behavior, and this should show up on any borrowed-by-a-proxy thing.

dnoneill: If you are the proxy I am assuming it shouldn't show up?

cbeer: It should show up on any borrowed-by-a-proxy thing whether you're a sponsor or proxy.

dnoneill: Yeah I am saying if you are the proxy who borrowed it should, it show up? For example, in our tests Piper Proxy borrowed something. If I was logged in as Piper proxy would the badge show up? I would assume no. But maybe the data does the work for us?

PR #3863, opened 2026-06-22
disable prev/next buttons if they fall outside the range of min/max calendar values

dnoneill:
https://github.com/user-attachments/assets/74674014-b196-4dde-9427-b2885c3ba849

Screenshot 2026-06-23 at 4 41 22 PM

### DIFF
--- a/app/javascript/controllers/date_picker_controller.js
+++ b/app/javascript/controllers/date_picker_controller.js
@@ -19,7 +19,7 @@ import { Controller } from "@hotwired/stimulus"
 //   grid               element where the day buttons are rendered
 //   availabilityStatus optional element shown while a month's availability is being fetched
 export default class extends Controller {
-  static targets = ["input", "calendar", "button", "selectedValue", "monthLabel", "grid", "announce", "prevBtn", "nextBtn", "legend", "availabilityStatus"]
+  static targets = ["input", "calendar", "button", "selectedValue", "monthLabel", "grid", "announce", "prevBtn", "nextBtn", "legend", "availabilityStatus", "leadTimeMessage"]
   static values = {
     disabled: Array, marked: Array, min: String, max: String, openDays: Array, year: Number, month: Number, focused: String,
     availabilityUrl: String,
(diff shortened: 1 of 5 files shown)


app/javascript/controllers/date_picker_controller.js
```diff
@@ -268,6 +271,13 @@ export default class extends Controller {
     ].join("-")
   }
 
+  #isPrevNextDisabled() {
+    this.prevBtnTarget.disabled =  new Date(this.yearValue, this.monthValue, 1) < new Date()
```
taylor-steve:
```suggestion
    this.prevBtnTarget.disabled = new Date(this.yearValue, this.monthValue, 1) <= new Date(this.minValue)
```
I think practically this is equivalent, but I wonder if we should we respect minValue

app/javascript/controllers/date_picker_controller.js
```diff
@@ -268,6 +271,13 @@ export default class extends Controller {
...
+    const nextDisabled = new Date(this.yearValue, this.monthValue, 31) > new Date(this.maxValue)
```
taylor-steve:
```suggestion
    const nextDisabled = new Date(this.yearValue, this.monthValue + 1, 1) > new Date(this.maxValue)
```
Is this functionally the same, but without the day math?

taylor-steve (approved):

cbeer:
There's something funny about the initial load today -- I can't go backwards to June until I go forwards first. 

The styling of the disabled button also seems funny to me (the prev button is disabled in this picture, but looks more clickable):
Screenshot 2026-06-25 at 08 16 56

I wonder if we can use the same unavailable date style, or if we need to hide the button entirely... or a more dramatic cursor style?

dnoneill:
I agree on the button. I found it to be kind of off. It also seemed odd that it didn't match the calendar elements. I can see if I can get it to match.
I will look at the weirdness on the button

dnoneill:
Okay I figured out that yearValue and monthValue aren't set on initialization. I moved the function into renderCalendar because it is called by everything. I also changed the disabled button. I think it looks much better
Screenshot 2026-06-25 at 11 59 24 AM

PR #3869, opened 2026-06-22
Display the proxy badge for requests.

cbeer:
At least pending design revisions that move it into the white area (if, say, we ever group multi-volume requests together)

Screenshot 2026-06-22 at 12 09 44

Completes #3847


dnoneill:
text needs to be changed to requested by

taylor-steve (approved):

PR #3870, opened 2026-06-22
fix sort/filter menu for accessibility

dnoneill:
Issues:

- sort/filter had problems with the role="menu"

Screenshot 2026-06-22 at 2 39 07 PM


cbeer:
Could you split the menu changes from the button labels? I feel pretty good about the `role="menu"` business, but want to spend a little more time thinking about the button labels (.. is a super long title + box information really helping / is context sufficient (and/or are we missing a context-boundary element/role somewhere else?)

Issue #2854, opened 2026-02-03
Home for My Library Account


dbranchini:
### Preview

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/af5472ff-fddf-45e1-bd2d-44b097aad5e7" />

### Designs with specificaitons

- [Card view (with items in each category)](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=8557-59616&t=w07BMeGKkSNcdlcs-4)
- [Card view (no items in any category)](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=8749-114189&t=w07BMeGKkSNcdlcs-4)
- [List view](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=8713-59954&t=w07BMeGKkSNcdlcs-4)

### Browser views

- Card view
- List view

### Design notes
- @cbeer, @taylor-steve and I discussed several ideas and options. We decided we want to try both with real data. The card view is more or less parity with production now (but with added Aeon cards). The list view contains the same information as the card view plus the user sees details and can take action on priority fees/items/appointments/etc.
- For the list view: 
  - If there isn't any activity within a card category, then the card is not shown.
  - All action buttons should behave the same as those actions on their associated pages. Reading room appointments should go the following:
    - [Items](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5163-76909&t=w07BMeGKkSNcdlcs-4) (perhaps it should be renamed to "Manage items")
    - [Edit appointment](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5335-22637&t=w07BMeGKkSNcdlcs-4) 
    - [Cancel](https://www.figma.com/design/sqzYVhfRgn3nMArUAdrs0n/My-Account---Requests?node-id=5418-98824&t=w07BMeGKkSNcdlcs-4) 
  - For the open to discussion on what appears in each section. I'm initially thinking the following:
    - Borrowed items - anything overdue or due within 3 days
    - Pickup requests - anything ready
    - Digitization requests - anything delivered within 3 days (should be the same as we are doing on submitted requests)
    - Reading room - the next three (OR should it be time-bound, meaning anything within a week?)
    - Activities - the next three (OR same question as above?)
- For the card view:
[...]

PR #3871, opened 2026-06-22
Add initial home page card and list views

taylor-steve:
Part of #2854 
Screenshot 2026-06-22 at 1 03 26 PM
Screenshot 2026-06-23 at 9 54 34 AM

Trying to leave things hand-tight as we're exploring both layout options and functionality. The dashboard model was a convenience, if we pick a single direction it could likely be consolidated or removed.

There's also the world where we turbo-with-placeholder these cards, some are slow loading. Keeping it simple while we decide between the two.

The variant switcher is only for this testing period, it's not part of the design. 

TODOs:

- [x] Put more thought into the markup/responsiveness of the questions aside
- [x] Break up `AppointmentTimeRangeComponent` into smaller composable components. The list view in the design only has `fw-semibold` for the date, `AppointmentTimeRangeComponent` applies it for all.
- [ ] Revisit all the data sources/sorting. Some things are stubbed, some are Aeon only where they should be mixed.
- [ ] Revisit what Darcy wants to control list view card visibility. For development, in the "List" view I added a "preview" fallback for some cards where Darcy suggested a time window. If there's nothing in the time window, but there are "upcoming" appointments/requests, I favored displaying a minimal set of those. Darcy has some question marks in the issue regarding what's best, this was my initial thought, but we need to have a discussion with her.
- [ ] Add actions for renew (borrowed items) and cancel (pickup)
- [ ] Add "Items" action for reading room appointments
- [ ] Wire up the pay button for real, I copy and pasted some stuff 🤷 
- [x] Populate any remaining stubbed links
- [x] Are all those ordered lists ordered?
- [ ] Tests

### DIFF
--- a/app/models/home/dashboard.rb
+++ b/app/models/home/dashboard.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+module Home
+  # Home page presenter wrapping user data for both view variants.
+  class Dashboard
+    DigitizationStatus = Data.define(:count, :label)
+
+    attr_reader :user
+
+    def initialize(user)
+      @user = user
+    end
+
+    def checkouts
+      @checkouts ||= patron.checkouts.sort_by { |c| c.sort_key(:due_date) }
+    end
+
+    def actionable_borrowed_items
+      @actionable_borrowed_items ||= checkouts.select { |c| c.due_date && c.due_date <= 3.days.from_now }
+    end
+
+    def borrowed_items_preview
+      @borrowed_items_preview ||= preview(actionable_borrowed_items, checkouts)
+    end
+
+    def fines
+      @fines ||= patron.fines
+    end
+
+    def balance
+      @balance ||= fines.sum { |f| f.owed || 0 }
+    end
+
+    def pickup_requests
+      @pickup_requests ||= preview(ready_for_pickup, patron.requests)
+    end
+
+    def ready_for_pickup
+      @ready_for_pickup ||= patron.requests.select(&:ready_for_pickup?)
+    end
+
+    def digital_requests
+      @digital_requests ||= aeon.requests.digitization.submitted.newest_first
+    end
+
+    def recently_delivered_digital_requests
+      @recently_delivered_digital_requests ||= aeon.requests.digitization.recently_delivered.newest_first(&:delivered_date)
+    end
+
+    def digitization_preview_requests
+      preview(recently_delivered_digital_requests, digital_requests)
+    end
+
+    def digitization_status
+      @digitization_status ||= begin
+        count = recently_delivered_digital_requests.count
+        label = count.positive? ? 'delivered recently via email' : 'Recently requested'
+        DigitizationStatus.new(count:, label:)
+      end
+    end
+
+    def appointments
+      @appointments ||= aeon.appointments
+    end
+
+    def appointment_requests_count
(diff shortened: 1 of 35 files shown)


dnoneill:
I am getting two accessibility errors. One I think is a false positive but I am including it for reference.
1. No heading. We should probably add a visually hidden h1 or make the card headers h1
2. Links are not clearly identifiable. This is only happening on Interlibrary loan and purchase recommendation links which makes me think its a false positive because it isn't getting tripped on the other links in the sidebar.

app/models/home/dashboard.rb
```diff
@@ -0,0 +1,144 @@
...
+      @aeon ||= user.aeon
+    end
+
+    def preview(primary, fallback, limit: 3)
+      primary.presence || fallback.first(limit)
+    end
+
+    def next_upcoming_date(collection)
+      now = Time.zone.now
+      collection.select { |item| item.start_time && item.start_time >= now }
```
dnoneill: I honestly don't know if this is better but it seems like it wouldn't you have loop through everything because find stops when it finds a match and sort should re-order them in the correct order

```suggestion
      collection.sort.find { |item| item.start_time && item.start_time >= now }
```

taylor-steve: Yes, I was definitely doing too much work and doing it in the wrong places. I've started moving it into more appropriate locations.

app/views/home/_list_view.html.erb
```diff
@@ -0,0 +1,143 @@
...
+    count: dashboard.checkouts.count,
+    view_all_label: 'View all borrowed items',
+    view_all_path: checkouts_path) do |c| %>
+    <% c.with_items do %>
+      <ul class="borrowed-items list-unstyled mb-2">
+        <% dashboard.borrowed_items_preview&.each do |request| %>
+          <li id="<%= dom_id(request, :home) %>" class="border-bottom d-flex flex-column gap-1 py-1">
+            <% if request.overdue? %>
+              <span class="fw-semibold small text-cardinal"><i class="bi bi-exclamation-triangle text-cardinal me-1"></i>Overdue</span>
+              <span class="ms-3">Accruing <%= number_to_currency(request.overdue_fines_rate['quantity']) %>/<%= request.overdue_fines_rate['intervalId'] %> until returned
```
cbeer: I have an overdue item with a nil `overdue_fines_rate` (fine policy `3.00/21.00 recall overdue fine`... i guess it only accrues fines if there's a recall?)

app/views/home/_list_view.html.erb
```diff
@@ -0,0 +1,143 @@
...
+
+  <%= render Home::ListView::SummaryCardComponent.new(
+    title: 'Borrowed items',
+    icon: 'bi-book',
+    count: dashboard.checkouts.count,
+    view_all_label: 'View all borrowed items',
+    view_all_path: checkouts_path) do |c| %>
+    <% c.with_items do %>
+      <ul class="borrowed-items list-unstyled mb-2">
+        <% dashboard.borrowed_items_preview&.each do |request| %>
```
cbeer: is `request` copy-pasta, or should this be `checkout` or similar?

taylor-steve: oh yeah, big time copy-pasta that should indeed be `checkout`

PR #3880, opened 2026-06-23
Display submitted Aeon + FOLIO requests on a single page

cbeer:
TODO:
- [x] resort the list as things load.
- [x] resort subgroups as things load too..
- [x] handle filtering client-side
- [x] progress / spinner indicators
- [x] how do we handle sort options that vary by source?
- [x] decide if this is even a good idea

Future work presumably includes:
- [ ] figure out what to call the default sort (hold shelf expiration + appointment time)
- [ ] expand this to cancelled/completed requests (if we determine we can/want to show that FOLIO data)
- [x] Map ILLiad data better (including differentiating ILL vs scans)
- [ ] ???

### DIFF
--- a/app/models/illiad_requests.rb
+++ b/app/models/illiad_requests.rb
@@ -64,11 +64,11 @@ class IlliadRequests
     def sort_key(key)
       sort_key = case key
                  when :date
-                   [*date_sort_key, title, author, call_number]
+                   [placed_date.strftime('%FT%T'), title, author, call_number]
                  when :title
                    [title, author, call_number]
-                 when :date_modified
-                   [-1 * placed_date.to_i, title, author, call_number]
+                 when :default
+                   [date_sort_key, title, author, call_number]
                  end
       sort_key.join('---')
     end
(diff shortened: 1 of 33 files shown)


app/views/requests/index.html.erb
```diff
@@ -0,0 +1,94 @@
...
+        <div class="bg-white px-3 py-2 d-flex column-gap-4 row-gap-2 justify-content-between align-items-start flex-wrap flex-md-nowrap">
+          <div class="placeholder col-4"></div>
+          <div class="col-5"></div>
+          <div class="placeholder col-1"></div>
+        </div>
+      <% end %>
+    <% end %>
+  <% end %>
+
+  <%= turbo_frame_tag :aeon_requests, src: aeon_requests_path(async: true, kind: 'submitted', aria: { hidden: true }) do %>
```
taylor-steve: `aria` as a path param?

app/models/illiad_requests.rb
```diff
@@ -67,7 +67,7 @@ def sort_key(key)
                    [*date_sort_key, title, author, call_number]
                  when :title
                    [title, author, call_number]
-                 when :date_modified
+                 when :default
```
taylor-steve, line 70: Do we need to have Illiad requests emit in `strftime('%FT%T')` format for sorting as well? It's a little hard to trace but I think they might be in seconds currently.

taylor-steve (approved):

PR #3881, opened 2026-06-23
add beta banner to requests

dnoneill:
closes #3883 
Screenshot 2026-06-23 at 2 41 33 PM
Screenshot 2026-06-23 at 2 44 42 PM

### DIFF
--- a/app/views/shared/_feature_flag.html.erb
+++ b/app/views/shared/_feature_flag.html.erb
@@ -1,15 +1,19 @@
-<% if can?(:toggle, :feature_flags) && !Settings.features.requests_redesign %>
-  <div class="my-2 col-8 text-end">
-    <% if use_requests_redesign? %>
-      <% if controller_name == 'patron_requests' && action_name == 'new' && @patron_request&.ead_url %>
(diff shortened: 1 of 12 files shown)


app/views/shared/_feature_flag.html.erb
```diff
@@ -1,15 +1,19 @@
 <% if can?(:toggle, :feature_flags) && !Settings.features.requests_redesign %>
-  <div class="my-2 col-8 text-end">
-    <% if use_requests_redesign? %>
-      <% if controller_name == 'patron_requests' && action_name == 'new' && @patron_request&.ead_url %>
-        <%= link_to Settings.aeon_archives_url + "&Value=#{@patron_request.ead_url}", class: 'su-underline' do %>
-          <i class="bi bi-box-arrow-up-right"></i> Use Aeon web interface
+  <div class="p-2" style="background: rgb(var(--stanford-illuminating-light-rgb))">
```
cbeer: Can we define e.g. `.bg-stanford-illuminating-light` instead of doing this inline?

dnoneill: I wasn't sure that worth it because when we go live we are going to rip this out. But yeah I can add it

PR #3894, opened 2026-06-24
Make the whole home page card a click target.

cbeer:


dnoneill:
Making note that the only reason this isn't a default class is the fees/fines has two links (pay now and view fines) and this wouldn't work with that.

dnoneill (approved):

PR #3899, opened 2026-06-24
add interactions for add item modal

dnoneill:
closes #3886 

As I mentioned in the ticket I think the modal needs to be a separate ticket 

Other fixes on this modal
- [x]  Change the title from "Eligible items... " to "Saved for later (Only materials associated with [x] are eligible to add)"
- [x]     Change button "X Remove" to "Save for later"
- [ ]     Constrain the height on the modal with a scroll
- [x] Add line above list item

https://github.com/user-attachments/assets/5fb5e139-446c-43f3-999a-1fa397bd311a

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -990,3 +990,31 @@ summary .disclosure-icon::before {
 [data-home-view="list"] [data-home-view-variant="card"] {
   display: none;
 }
+
+.request.list-group-item {
+  --border-top: 1px solid var(--stanford-20-black);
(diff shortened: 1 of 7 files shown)


cbeer:
nice, but it looks like the border around the highlighted item snaps away (awkwardly, after the background color fades). Can we fade the border too?

dnoneill:
Yep. Will do.

dnoneill:
Okay I think fixed 

https://github.com/user-attachments/assets/cbe11056-f1f4-43d3-aedd-6ff6aa9abdc9

PR #3901, opened 2026-06-25
Use slots for home list view cards.

cbeer:


app/components/home/list_view/summary_card_component.rb
```diff
@@ -4,18 +4,28 @@ module Home
...
+
       renders_one :items
 
-      attr_reader :title, :icon, :count, :pill_noun, :view_all_label, :view_all_path, :status
+      attr_reader :title, :count, :pill_noun, :view_all_label, :view_all_path
 
-      def initialize(title:, icon:, count:, view_all_label:, view_all_path:, pill_noun: 'total', status: nil) # rubocop:disable Metrics/ParameterLists
+      def initialize(title:, icon:, count:, view_all_label:, view_all_path:, pill_noun: 'total') # rubocop:disable Metrics/ParameterLists
         @title = title
         @icon = icon
```
dnoneill: Should this get changed to icon_class or something along those lines. I think it might be a little weird to have both an icon slot and an icon variable. Especially since the variable really is just a class not the html markup

app/components/home/list_view/summary_card_component.rb
```diff
@@ -4,18 +4,28 @@ module Home
   module ListView
     # Generic count-with-teaser card for the list-view home page.
     class SummaryCardComponent < ViewComponent::Base
+      renders_one :icon, ->(icon:) { render_icon(icon) }
```
dnoneill, line 7:
```suggestion
      renders_one :icon, ->(&block) { render_icon(&block) }
```

dnoneill: Maybe I am confused about this. Wouldn't it be block? I think I am confused by how this slot works. Or would it be icon_class?

app/components/home/list_view/summary_card_component.rb
```diff
@@ -4,18 +4,28 @@ module Home
...
-        @icon = icon
+        @icon_class = icon_class
         @count = count
         @pill_noun = pill_noun
         @view_all_label = view_all_label
         @view_all_path = view_all_path
-        @status = status
+      end
+
+      def default_icon
```
dnoneill, line 23: cruft?

cbeer:
We're going to pause list view for the moment (#3915) and focus on the card layout.

PR #3914, opened 2026-06-25
add tab param for activities

dnoneill:


cbeer:
Yeah, I'm not sure the indirection adds much clarity. There's probably a way to slice it up that works better, but it feels like a pretty big lift for marginal improvement?

PR #3918, opened 2026-06-25
add save for later functionality to cancelled items

dnoneill:
closes #3900 

Different thing noticed here. Some things don't have volumes (see Stanford microform screenshots) but we still have background and space for it. Obviously separate ticket

Screenshot 2026-06-25 at 4 33 03 PM
Screenshot 2026-06-25 at 4 32 54 PM
Screenshot 2026-06-25 at 4 32 45 PM
Screenshot 2026-06-25 at 4 32 35 PM

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -723,6 +723,31 @@ summary .disclosure-icon::before {
   }
 }
 
+.cancelled.request.list-group-item {
+  grid-template-columns: 1fr;
+  grid-template-areas:
+    "id"
+    "detail"
+    "actions"
+    "footer";
+
+  @media (min-width: 576px) {
+    grid-template-columns: 1fr auto;
+    grid-template-areas:
+      "id          id"
+      "detail      actions"
+      "footer      footer";
+  }
+
+  @media (min-width: 992px) {
+    grid-template-columns: 1fr auto;
+    grid-template-areas:
+      "id      actions"
+      "detail       detail"
+      "footer       footer";
+  }
+}
+
 .list-group:has(.request.list-group-item .request-fulfillment) .request.list-group-item {
   grid-template-columns: 1fr;
   grid-template-areas:
@@ -741,18 +766,24 @@ summary .disclosure-icon::before {
       "footer      footer";
   }
 
-  @media (min-width: 992px) {
-    grid-template-columns: 1fr 19.5rem 3rem;
-    grid-template-areas:
-      "id     fulfillment actions"
-      "detail detail      detail"
-      "footer footer      footer";
+  &:not(.cancelled) {
+    @media (min-width: 992px) {
+      grid-template-columns: 1fr 19.5rem 3rem;
+      grid-template-areas:
+        "id     fulfillment actions"
+        "detail detail      detail"
+        "footer footer      footer";
+    }
   }
-}
 
-.list-group:has(.request.list-group-item .request-fulfillment):has(.btn.resubmit) .request.list-group-item {
-  @media (min-width: 992px) {
-    grid-template-columns: 1fr 19.5rem 5.5rem;
+  &.cancelled {
+    @media (min-width: 992px) {
+      grid-template-columns: 1fr 19.5rem 8rem;
+      grid-template-areas:
+        "id     fulfillment actions"
+        "detail detail      detail"
+        "footer footer      footer";
+    }
   }
 }
 
--- a/app/components/aeon/request_actions_component.html.erb
+++ b/app/components/aeon/request_actions_component.html.erb
@@ -1,59 +1,57 @@
-    <div class="d-flex text-nowrap align-items-baseline align-self-start justify-content-xl-end">
-      <% if request.cancelled? %>
-        <form action="<%= resubmit_aeon_request_path(transaction_number) %>" method="PATCH">
-          <button type="submit" class="btn btn-link btn-sm su-underline resubmit" aria-label="Resubmit request"><i class="bi bi-arrow-clockwise"></i> Resubmit</button>
-        </form>
+<div class="d-flex text-nowrap align-items-baseline align-self-start justify-content-xl-end">
+  <% if include_bulk_actions? %>
+    <span>
+      <input type="checkbox" class="form-check-input" id="delete-bulk-<%= request.id %>"
+        data-action="change->request-bulk-delete#updateDeleteButton request-bulk-delete#updateSelectAll"
+        data-request-bulk-delete-target="select" data-id="<%= request.id %>">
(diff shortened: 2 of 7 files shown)


app/components/aeon/request_actions_component.html.erb
```diff
@@ -1,33 +1,27 @@
...
+            <i class="bi bi-pencil-fill me-1 align-middle"></i><span class="visually-hidden">Edit request</span>
           <% end %>
         <% end %>
-        <% if helpers.can? :destroy, request %>
-          <button class="btn btn-link su-underline p-0" data-bs-toggle="modal" data-bs-target="#delete-<%= transaction_number %>">
-            <i class="bi bi-trash align-middle"></i><span class="visually-hidden">Delete <%= title %> request</span>
-          </button>
-        <% end %>
+      <% end %>
+      <% if !request.cancelled? && helpers.can?(:destroy, request) %>
```
cbeer: Shouldn't we just update the ability to say cancelled requests can't be deleted?

app/components/aeon/request_actions_component.html.erb
```diff
@@ -1,33 +1,27 @@
...
+        <span>
+          <input type="checkbox" class="form-check-input" id="delete-bulk-<%= request.id %>"
+            data-action="change->request-bulk-delete#updateDeleteButton request-bulk-delete#updateSelectAll"
+            data-request-bulk-delete-target="select" data-id="<%= request.id %>">
+          <%= label_tag "delete-bulk-#{request.id}", 'Select', class: 'ms-1 me-3 small form-check-label fw-normal' %>
+        </span>
+      <% end %>
+      <% if helpers.can? :edit, request %>
+        <%= link_to edit_aeon_request_path(request), class: 'btn btn-link btn-sm su-underline p-0 me-sm-2', data: { action: 'modal#open' } do %>
+          <% if (request.saved_for_later? || request.cancelled?) && request.digital? %>
```
cbeer: These conditions were barely readable before and the added clause doesn't help. I'm not sure what the `else` case is trying to be, but maybe this would be more readable if it came first?

app/components/aeon/request_actions_component.html.erb
```diff
@@ -1,33 +1,27 @@
     <div class="d-flex text-nowrap align-items-baseline align-self-start justify-content-xl-end">
-      <% if request.cancelled? %>
-        <form action="<%= resubmit_aeon_request_path(transaction_number) %>" method="PATCH">
-          <button type="submit" class="btn btn-link btn-sm su-underline resubmit" aria-label="Resubmit request"><i class="bi bi-arrow-clockwise"></i> Resubmit</button>
```
cbeer, line 4: There's a gnarly css selector targetting `.btn.resubmit` that can probably be changed or removed.

app/components/aeon/request_actions_component.html.erb
dnoneill: I did that but the button stills shows up when the user is a super-admin. The delete button doesn't do anything but just hits the aeon server and does nothing because you can't delete a cancelled request. I didn't think we should be showing our super admins a button that doesn't do anything. Maybe this a use case that isn't going to happen but I found it not great.

app/components/aeon/request_actions_component.html.erb
dnoneill: Okay. Yeah if the reuqest isn't saved_for_later? && cancelled?. That should be easy enough

app/assets/stylesheets/application.css
```diff
@@ -758,12 +758,6 @@ summary .disclosure-icon::before {
   }
 }
 
-.list-group:has(.request.list-group-item .request-fulfillment):has(.btn.resubmit) .request.list-group-item {
-  @media (min-width: 992px) {
-    grid-template-columns: 1fr 19.5rem 5.5rem;
-  }
-}
-
```
taylor-steve, line 756: I think you'll need to adapt this, not remove it.

Image

taylor-steve (commented):
Noting that cancelled requests can and do have old appointments attached to them in production, although it's a limited number.

app/assets/stylesheets/application.css
dnoneill: Done

dnoneill: Actually no something else needs to happen

dnoneill: Okay now it is fixed

dnoneill:
@taylor-steve do you want me to do anything about the attached appointment in production?

app/models/aeon/request.rb
```diff
@@ -132,6 +132,12 @@ def cancelled_by_staff?
       (digital? && photoduplication_queue&.cancelled_by_staff?) || transaction_queue&.cancelled_by_staff?
     end
 
+    def cancelled_by_user?
+      return false unless cancelled?
+
+      (digital? && photoduplication_queue&.cancelled_by_user?) || transaction_queue&.cancelled_by_user?
+    end
+
```
taylor-steve: I see this method got added and wired up in `RequestFulfillmentComponent` as a mode, but that mode seems to be unused. Leftovers or something yet to be done?

taylor-steve:
> @taylor-steve do you want me to do anything about the attached appointment in production?

No, let's ignore it for now, thanks.

app/models/aeon/request.rb
dnoneill: No on purpose. The grid doesn't work properly if there isn't a request-fufillment class and we get a weird mixed view in terms of the location of the actions. This allows us to put messaging in the future but also makes sure the grid looks right without it. If we don't include this we have to do some really weird css stuff otherwise the page looks like this:
Screenshot 2026-07-02 at 12 10 18 PM

taylor-steve: Ahhhh, that explains my confusion with the empty request fulfillment HTML. I think we should to do the weird CSS stuff. The intent of the `.list-group:has(.request.list-group-item .request-fulfillment):has(.btn.resubmit) .request.list-group-item` rule was:
* If a list group has any items that display fulfillment info and have an action (resubmit then, add appointment now) provide appropriate space for all those elements
* Otherwise, provide more room for the title

Many users won't have "Cancelled by staff" messaging, so they won't have any fulfillment info on the page, so we *want* more room for the title in that case.

See how it gets squashed? If there's no fulfillment info at all on the page, this wrapping looks out of place
Image

taylor-steve (approved):
We still have mixed positioning, so I don't really think the extra complication with the type specific styles are worthwhile

Image

dnoneill:
@taylor-steve I can fix that mixed positioning. That was a whim. I will switch it back.

PR #3920, opened 2026-06-25
fix style of js sort dropdowns to match figma

dnoneill:
Before:
Screenshot 2026-06-25 at 5 23 19 PM

After:
Screenshot 2026-06-25 at 5 22 42 PM

Figma:
Screenshot 2026-06-25 at 5 22 32 PM

### DIFF
--- a/app/views/checkouts/index.html.erb
+++ b/app/views/checkouts/index.html.erb
@@ -3,20 +3,13 @@
     <div class="d-flex align-items-center justify-content-between flex-wrap row-gap-2">
       <h1 class="mb-0">Borrowed (checked out) items</h1>
       <div class="d-flex gap-2 align-items-center">
-        <div class="dropdown" data-sortable-target="sortMenu">
-          <button class="btn btn-outline-primary btn-sm dropdown-toggle" role="button" id="dropdownMenuLink" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-            Sort by due date
-          </button>
-
-          <ul class="dropdown-menu" aria-labelledby="dropdownMenuLink">
-            <li><button class="dropdown-item active" data-action="click->sortable#sort" data-sortable-sort-param="dueDate" data-sortable-label-param="Sort by due date">due date</button></li>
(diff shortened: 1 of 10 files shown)


cbeer:
I question whether `SortableDropdownComponent` is where we want to slice up the component. It feels a little incomplete .. if we're convinced we need to extract a component, shouldn't we extract the whole thing and not just the menu? 

(Perhaps I'm also skeptical that we need to extract a component. A few lines of relatively static HTML didn't bother me?)

app/views/checkouts/index.html.erb
```diff
@@ -7,15 +7,7 @@
           <button class="btn btn-outline-primary btn-sm dropdown-toggle" role="button" id="dropdownMenuLink" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
             Sort (Due date)
```
cbeer: Presumably this label needs an update too?

app/components/sortable_dropdown_component.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Render sort dropdown
+class SortableDropdownComponent < ViewComponent::Base
+  attr_reader :field_label_pair
+
+  def initialize(field_label_pair: {})
```
cbeer: `field_label_pair`  doesn't feel very view-component-y to me. I wonder if the options want to be slots?

app/views/checkouts/index.html.erb
dnoneill: Yeah.

app/components/sortable_dropdown_component.rb
dnoneill: In regards to your question about static html is that is everywhere and we aren't catching it in terms updating the labels. Perhaps we will only have to update it once but it gets kind of annoying having to update if the things change in terms of presentation. 

We could get the entire dropdown in the component. I don't have a problem with that. It would be relatively simple.

In regards to slots are you envisioning something like this?
component.with_menu_item(label: label, search_field: search_field) and then the component populates? I like that idea. It makes sense to me.

dnoneill: What I am assuming you are proposing
Screenshot 2026-06-26 at 1 09 40 PM
Screenshot 2026-06-26 at 1 08 52 PM

PR #3932, opened 2026-06-29
add renewal interactions

dnoneill:
closes #3919 

Hide whitespace is your friend on this PR

https://github.com/user-attachments/assets/2407b47c-8737-46b6-966d-4050715b5342

https://github.com/user-attachments/assets/bba53af7-feb5-4056-aa2f-bfafe75f888b

### DIFF
--- a/app/controllers/checkouts_controller.rb
+++ b/app/controllers/checkouts_controller.rb
@@ -21,6 +21,7 @@ class CheckoutsController < ApplicationController
 
   def renew # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
     @response = FolioClient.new.renew_checkout(@checkout)
+    update_checkouts([@response.updated_checkout]) if @response.success?
 
     respond_to do |format|
       format.html do
@@ -38,9 +39,10 @@ class CheckoutsController < ApplicationController
   # Renew all eligible items for a patron
   #
   # POST /checkouts/renew_eligible
-  def renew_eligible
+  def renew_eligible # rubocop:disable Metrics/AbcSize
     eligible_renewals = @checkouts.select(&:renewable?)
     @responses = eligible_renewals.map { |checkout| FolioClient.new.renew_checkout(checkout) }
+    update_checkouts(@responses.select(&:success?).map(&:updated_checkout))
 
     respond_to do |format|
       format.html do
@@ -67,6 +69,11 @@ class CheckoutsController < ApplicationController
     @checkout = @checkouts.find { |checkout| checkout.item_id == params[:id] }
 
     raise ActiveRecord::RecordNotFound, 'Checkout not found' if @checkout.nil?
+  end
+
+  def update_checkouts(updated_checkouts)
+    updates = updated_checkouts.index_by(&:id)
+    @checkouts = @checkouts.map { |checkout| updates[checkout.id] || checkout }
   end
 
   def bulk_renewal_success_flash(responses)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -246,17 +246,16 @@ en:
     subject: Stanford Libraries PIN reset
   mylibrary:
     renew_item:
-      in_progress_html: <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span> Renewing this item
-      success_html: <span class="font-weight-bold">Success!</span> "%{title}" was renewed.
+      success_html: Successfully renewed 1 item.
       error_html: <span class="font-weight-bold">Sorry!</span> Something went wrong and "%{title}" was not renewed.
       deny_access: An unexpected error has occurred
     renew_all_items:
       in_progress_html:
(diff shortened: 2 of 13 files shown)


dnoneill:
Renew all button go updated so the spinner isn't misaligned.
Screenshot 2026-06-29 at 4 40 14 PM

app/controllers/checkouts_controller.rb
```diff
@@ -69,6 +71,13 @@ def load_checkout
     raise ActiveRecord::RecordNotFound, 'Checkout not found' if @checkout.nil?
   end
 
+  def update_checkouts(updated_checkout)
+    index = patron_or_group.all_checkouts.find_index { |checkout| checkout.id == updated_checkout.id }
+    return unless index >= 0
```
taylor-steve:
```suggestion
    return if index.nil?
```
`find_index` can return `nil`, which can't be compared to 0 (it no methods), so maybe something like this.

app/controllers/checkouts_controller.rb
```diff
@@ -69,6 +71,13 @@ def load_checkout
...
+
+    patron_or_group.all_checkouts[index] = updated_checkout
```
taylor-steve: Pointing my finger at mutation, as is my nature 😄 

If we're only mutating for the renew all partial, did you consider changing the partial to take a local? Then you could do something like:

```ruby
  def updated_checkouts_from(responses)
    updates = responses.select(&:success?).map(&:updated_checkout).index_by(&:id)
    @checkouts.map { |checkout| updates[checkout.id] || checkout }
  end

  # renew
  @checkouts = updated_checkouts_from([@response])

  # renew_eligible
  @checkouts = updated_checkouts_from(@responses)
```

Just a thought. Doesn't look like the mutation bites us now, but mutation on memoized data makes me think.

app/controllers/checkouts_controller.rb
```diff
@@ -38,9 +39,10 @@ def renew # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
   # Renew all eligible items for a patron
   #
   # POST /checkouts/renew_eligible
-  def renew_eligible
+  def renew_eligible # rubocop:disable Metrics/AbcSize
     eligible_renewals = @checkouts.select(&:renewable?)
     @responses = eligible_renewals.map { |checkout| FolioClient.new.renew_checkout(checkout) }
+    @responses.select(&:success?).map { |response| update_checkouts(response.updated_checkout) }
```
taylor-steve:
```suggestion
    @responses.select(&:success?).each { |response| update_checkouts(response.updated_checkout) }
```

If you keep the mutation, I think this reads more clearly as `each`. We're iterating only for the side effects here, correct?

app/controllers/checkouts_controller.rb
dnoneill: I did. Wasn't sure that was better but can do.

dnoneill: Okay I didn't have to use locals. For some reason I didn't think checkouts loaded every time but it does so I switched this to use the global checkouts variable

config/locales/en.yml
```diff
@@ -246,17 +246,16 @@ en:
...
-      success_html: <span class="font-weight-bold">Success!</span> "%{title}" was renewed.
+      success_html: Successfully renewed 1 item.
       error_html: <span class="font-weight-bold">Sorry!</span> Something went wrong and "%{title}" was not renewed.
       deny_access: An unexpected error has occurred
     renew_all_items:
       in_progress_html:
-        one: <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span> Renewing 1 eligible item
-        other: <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span> Renewing %{count} eligible items
+        one: <span class="spinner-border spinner-border-sm me-1" style="vertical-align:center;" role="status" aria-hidden="true"></span> Renewing 1 item
+        other: <span class="spinner-border spinner-border-sm me-1" style="vertical-align:center;" role="status" aria-hidden="true"></span> Renewing %{count} items
```
taylor-steve, line 255: Sorry I missed this before, did you mean `middle` rather than `center`?

taylor-steve (approved):

config/locales/en.yml
dnoneill: No I did not. Unfortunately vertical-align middle which does have a bootstrap class but it looks like below. It is super annoying which is why I just inlined the code which is also annoying but keeps it from being oddly aligned.

Screenshot 2026-07-02 at 12 05 43 PM

taylor-steve: I was just surprised because that's not a valid value: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/vertical-align

dnoneill: Yes it is: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/vertical-align#browser_compatibility. I guess it is only for Firefox tho which I didn't realize

PR #3940, opened 2026-06-30
update form single/multiple selected item layout

dnoneill:
closes #3913 
closes #3908 

Screenshot 2026-07-01 at 6 56 55 PM
Screenshot 2026-07-01 at 6 56 52 PM
Screenshot 2026-07-01 at 6 56 40 PM
Screenshot 2026-07-01 at 6 56 28 PM
Screenshot 2026-07-02 at 12 38 39 PM
Screenshot 2026-07-02 at 12 38 14 PM

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -434,13 +434,12 @@ summary .disclosure-icon::before {
   grid-template-columns: 1fr;
   grid-template-areas:
     "title"
-    "select"
-    "actions";
+    "select";
 
   @media (min-width: 768px) {
     grid-template-columns: 1fr auto;
     grid-template-areas:
-      "title  actions"
+      "title"
       "select select";
   }
 
@@ -448,12 +447,8 @@ summary .disclosure-icon::before {
     grid-template-columns: auto 1fr;
     grid-template-areas:
       "title   select"
-      "actions select";
+      "select";
     align-items: center;
-
-    .appt-select {
-      justify-content: flex-end;
-    }
   }
 }
 
@@ -462,7 +457,6 @@ summary .disclosure-icon::before {
   grid-template-columns: 1fr;
   grid-template-areas:
     "title"
-    "actions"
     "select";
 
   .appt-select {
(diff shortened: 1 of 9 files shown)


taylor-steve:
I wonder if it would make sense to do https://github.com/sul-dlss/sul-requests/issues/3908 at the same time, since we're actively getting feedback? This by itself significantly changes the layout of the multiple item case. I'm not sure I'm smart enough to review one at a time.

dnoneill:
Okay I can do. I was actually thinking I would pick that up and branch off this but I can put it in the same PR instead

taylor-steve:
> Okay I can do. I was actually thinking I would pick that up and branch off this but I can put it in the same PR instead

It'd be fine to do as stacked PRs, I'm saying it's likely we'd wait to review until we can look at the lump sum

dnoneill:
Added multiple select layout to this PR. should be ready

taylor-steve:
Could you please take another pass on responsiveness? Some things don't feel quite right.
Screenshot 2026-07-02 at 10 56 19 AM

dnoneill:
I am not seeing that. I tried the exact same one you did.

dnoneill:
nvm. Was at a weird breakpoint. Fixed

cbeer:
I'm looking at the multi-item/reading room display here:

- [ ] It looks like we have a mix of button/link styles (only Save for later has the underline dots):

Screenshot 2026-07-02 at 14 07 53

- [ ] I'm not sure why my appointment selection is wrapped, no matter how wide I make my browser. 

- [ ] I think Darcy's design has some vertical padding you're missing.

- [ ] Something's gone wrong with the view contents modal (it lost the backdrop, and reveals itself funny):
Screenshot 2026-07-02 at 14 11 56

cbeer:
And for multi-item digitization:

- [ ] Darcy's designs have some top padding we're missing:
Screenshot 2026-07-02 at 14 16 18

(and the same mix of link styles)

- [ ] Darcy also has the hide/show caret vertically centered (and bigger?)

- [ ] The success checkmark looks funky:

Screenshot 2026-07-02 at 14 17 46

PR #3955, opened 2026-07-02
add scheduling/saving for later interaction on appointment page

dnoneill:
part of #3885 
The interaction for flashing still needs work. I am having trouble untangling because of the way we are rendering on the page.

https://github.com/user-attachments/assets/5ac5a737-493d-4581-8a92-1d0beea69c2a


cbeer:
I was mostly ok with the renewal button using custom javascript because of the weird interaction with the renew all button, but I guess I don't see the save for later can't just be `turbo_submits_with` with a couple lines of CSS to remove the button style (e.g. #3957)

dnoneill:
solved by https://github.com/sul-dlss/sul-requests/pull/3957

PR #3959, opened 2026-07-03
Include status in single-item group_key to avoid empty group cards

taylor-steve:
I think the group for a reading room request with an appointments gets removed after the request is destroyed (presumably due to some appointment handling?), but digitization for non-multi-item selector requests are not currently.

Screenshot 2026-07-02 at 4 52 33 PM
Screenshot 2026-07-02 at 4 52 47 PM

### DIFF
--- a/app/models/aeon/request.rb
+++ b/app/models/aeon/request.rb
@@ -77,7 +77,7 @@ module Aeon
       if multi_item_selector?
         [status, title, digital?]
       else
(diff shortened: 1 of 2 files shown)


cbeer:
I'm not totally convinced this is the right fix (and we didn't screw up the update response somewhere), but the test is good 🤷‍♂️

PR #3961, opened 2026-07-06
Maintain grid style for new labelled Aeon edit links 

taylor-steve:
Targets https://github.com/sul-dlss/sul-requests/pull/3918

I think it is easier to maintain current grid behavior while adding these new actions and review any changes to the responsive styling separately.

### DIFF
--- a/app/components/aeon/request_actions_component.html.erb
+++ b/app/components/aeon/request_actions_component.html.erb
@@ -8,17 +8,20 @@
     </span>
   <% end %>
   <% if helpers.can? :edit, request %>
-    <%= link_to edit_aeon_request_path(request), class: 'btn btn-link btn-sm su-underline p-0 me-sm-2', data: { action: 'modal#open' } do %>
-      <% if !request.saved_for_later? && !request.cancelled? %>
-        <i class="bi bi-pencil-fill me-1 align-middle"></i><span class="visually-hidden">Edit request</span>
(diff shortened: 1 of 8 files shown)


app/components/aeon/request_actions_component.html.erb
```diff
@@ -8,17 +8,20 @@
...
+          <i class="bi bi-calendar-plus me-1 align-middle"></i> Add appointment
+        <% end %>
       <% else %>
-        <i class="bi bi-calendar-plus me-1 align-middle"></i> Add appointment
+        <i class="bi bi-pencil-fill me-1 align-middle"></i><span class="visually-hidden">Edit request</span>
       <% end %>
     <% end %>
   <% end %>
-  <% if !request.cancelled? && helpers.can?(:destroy, request) %>
+  <% if helpers.can?(:destroy, request) %>
```
cbeer, line 24: I was confused about this too, but without making a change in Ability, we end up showing super admins a destroy button for these requests  🤷

dnoneill (changes requested):
This breaks the thing I was trying to fix with the grid as it was which is there is mixed alignment with actions that we don't want unless something has changed

Image

taylor-steve:
I'd argue it maintains current behavior 😄

dnoneill:
Okay but my understanding is we don't want current behavior and the PR you are merging against does not maintain current behavior.

PR #3970, opened 2026-07-06
Exclude component previews from Style/Documentation cop

taylor-steve:


.rubocop.yml
```diff
@@ -29,6 +29,10 @@ Lint/MissingSuper:
   Exclude:
     - "app/components/**/*.rb"
 
+Style/Documentation:
+  Exclude:
+    - "spec/component_previews/**/*.rb"
```
cbeer: The rubocop default is:

```suggestion
    - "spec/**/*.rb"
```

(well, it includes test too, but we don't have one so 🤷 )

PR #3974, opened 2026-07-06
Add axe specs for checkouts and fees pages

taylor-steve:


app/components/folio/proxy_indicator_component.html.erb
```diff
@@ -1,5 +1,5 @@
 <div>
-  <%= render PillComponent.new(classes: %w[text-white bg-poppy-dark me-1]) do %>
+  <%= render PillComponent.new(classes: %w[text-white bg-poppy-darker me-1]) do %>
```
cbeer, line 2: I'm confused about this one. I thought Darcy specified this new, darker poppy color for accessibility (and `bg-poppy-darker` is darker than `bg-poppy-dark`?).

Measuring in Firefox, I see the pill contrast as 4.63 which is fine?

taylor-steve: I'm confused by your comment, it sounds like you're making a case for this change.

As far as I can tell, Figma specifies `#D1660F` for this pill, which I believe is `bg-poppy-darker`.

cbeer: Yep, I got my proxy pills confused. I was looking at the one in the navbar which is already darker.

PR #3980, opened 2026-07-07
Start updating selected item form styles

cbeer:
This PR is an alternate take on #3908 + #3913

- [x] updates the vertical rhythm to use 24px (1.5rem / gap-4) spacing
- [x] bumps the actions below the title (generally) 
- [x] updates the new request column layout to use  9 / 3 (instead of 8 / 4) to make more room for the form element at -lg and -xl 
- [x] adds action labels and make the buttons consistent

Changes from the design:
- [x] Darcy is reluctantly ok with keeping the actions outside the volume box area

Future TODOs:
- [ ] need a decision about action wrapping at -md / -lg
- [ ] handle hiding the additional notes field if all requests are save-for-later'ed (extracted to #3981)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -435,13 +435,6 @@ summary .disclosure-icon::before {
     "actions";
 
   @media (min-width: 768px) {
-    grid-template-columns: 1fr auto;
-    grid-template-areas:
-      "title  title"
-      "actions select";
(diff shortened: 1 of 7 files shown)


taylor-steve (approved):
I like the extra space for the title. The positioning of the actions seems more consistent versus main. I'm happy to try this.

PR #3981, opened 2026-07-07
Hide + show the additional information section when there's no submitted items.

cbeer:
Part of #3913, done with CSS instead of javascript toggling.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -462,6 +462,10 @@ summary .disclosure-icon::before {
   }
 }
 
+#reading:not(:has(.appointments-item)) #additional-information {
+  display: none;
+}
+
 .selected-items-container {
   .accordion-item:last-child .next-item {
     display: none;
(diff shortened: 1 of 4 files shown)


taylor-steve (approved):
Is this only for the multiple case? Single still shows.

Image

dnoneill:
Yeah I attempted to do this with css instead of JS. Then again the dom has changed so it might be possible now. But I am seeing the same thing as Steve where it doesn't work if you have one saved for later item.

app/assets/stylesheets/application.css
```diff
@@ -462,6 +462,11 @@ summary .disclosure-icon::before {
   }
 }
 
+#reading:not(:has(.appointments-item)) #additional-information {
+  background-color: red;
```
dnoneill: cruft.

app/jobs/submit_aeon_patron_request_job.rb
```diff
@@ -90,7 +90,7 @@ def common_aeon_data_from_patron_request(patron_request, volume_params, activity
       site: patron_request.aeon_site,
       special_request: if patron_request.request_type == 'scan'
                          volume_params['additional_information']
-                       else
+                       elsif volume_params['appointment_id'].present?
```
dnoneill, line 93: Is this purposeful to this PR it feels separate?

cbeer: Yes, very purposeful. If you don't do this, we still save the additional information onto the saved requests.

dnoneill (approved):

PR #3993, opened 2026-07-08
add flash interaction for appointments

dnoneill:
closes #3885 

https://github.com/user-attachments/assets/33fcc5ec-97b3-4fc8-aa50-ea0ff033165d

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -960,31 +960,58 @@ summary .disclosure-icon::before {
   min-height: 196px;
 }
 
-.request.list-group-item {
-  --border-top: 1px solid var(--stanford-20-black);
+.request {
+  --border-top: 0;
   --border-side-bottom: 0;
+  --flash-border-top-end: 0; /* default end-state for border-top after flash */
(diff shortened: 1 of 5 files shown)


taylor-steve:
Can we use Darcy's styles for both the flash placeholders, she's got rounded borders and different padding.
Screenshot 2026-07-08 at 5 08 19 PM

Is there any hope for slow loading guys like me? If I tap a couple at a time, it's chaos:

Screenshot 2026-07-08 at 5 12 22 PM

hard to capture a good image, but they overwrite each other here and the page is left in an inconsistent state versus reality.

Screenshot 2026-07-08 at 5 12 27 PM

dnoneill:
@taylor-steve The replacement issue is on main as well. I think it has to do with the anonymous turbo frame that got added but that isn't an issue specific to this branch. I will fix the rounding issue and the padding on the sidebar. I did look at the padding in the appointment block but it seems like the padding on the actual view is different from what we had and I honestly has having trouble untangling things because even the figma seems kind of inconsistent in terms of interaction.

taylor-steve (approved):

PR #4014, opened 2026-07-09
Anticipate the modal title

cbeer:
This reduces the number of things that jump around when the content actually loads 🤷‍♂️


dnoneill:
Is there a reason we shouldn't use a placeholder instead? I mean similar to what we do on the requests page where we scaffold out what we expect the header to look like

cbeer:
I think providing the title (which we basically know already) is a good affordance for the user (so they know they're loading the right thing).

taylor-steve (approved):

PR #4039, opened 2026-07-10
make sure focus stays on day after avaliability loads

dnoneill:
closes #3911 

From what I can tell this is happening because await on fetch removes focus

### DIFF
--- a/app/javascript/controllers/date_picker_controller.js
+++ b/app/javascript/controllers/date_picker_controller.js
@@ -196,18 +196,22 @@ export default class extends Controller {
 
     // Fallback: if focusedValue is outside this month or disabled, use the first enabled day
     let focusedBtn = this.gridTarget.querySelector("button[tabindex='0']")
+
     if (!focusedBtn || focusedBtn.disabled) {
       focusedBtn = this.gridTarget.querySelector("button:not(:disabled)")
       if (focusedBtn) {
         focusedBtn.tabIndex = 0
         this.focusedValue = focusedBtn.dataset.datePickerDateParam
       }
+    } else if(!focusedBtn.disabled) {
+      // make sure we keep focus on the button.
+      // This will go away during fetchMonthAvailability because focusedValue/focusedBtn is already set
+      focusedBtn?.focus({ focusVisible: true })
     }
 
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/date_picker_controller.js
```diff
@@ -207,7 +207,6 @@ export default class extends Controller {
     this.#isPrevNextDisabled()
     this.#fetchMonthAvailability(year, month)
     this.#fetchMonthAvailability(...this.#nextMonth(year, month))
-    this.#updateAvailabilityStatus()
```
dnoneill, line 209: I removed this because it is called in fetchMonthAvailability and I don't think it needs to be called if there is no availability check so this seems redundant

app/javascript/controllers/date_picker_controller.js
```diff
@@ -228,6 +227,7 @@ export default class extends Controller {
         this.disabledValue = [...new Set([...this.disabledValue, ...additions])]
         this.renderCalendar()
       }
+      this.gridTarget.querySelector("button[tabindex='0']")?.focus({ focusVisible: true })
```
taylor-steve: I believe `renderCalendar` is the culprit (which is called by `fetchMonthAvailability`, so you're correct that it's a side-effect).

I think we should push your change in there, because the element replacement `renderCalendar` does is the cause of the focus loss. Maybe something like this around here:
https://github.com/sul-dlss/sul-requests/blob/53e99a88833ed2e07bea343c9d4c6c930a8f048f/app/javascript/controllers/date_picker_controller.js#L188

```js
const gridHadFocus = this.gridTarget.contains(document.activeElement)
this.gridTarget.replaceChildren(table)
if (gridHadFocus) this.gridTarget.querySelector("button[tabindex='0']")?.focus({ focusVisible: true })
```

dnoneill: I checked that because that was my original thought but it was not getting additions. I forgot about the next month which was *forehead slap

dnoneill: Okay fixed

taylor-steve (approved):

PR #4049, opened 2026-07-13
Use re-submit as the button label for cancelled requests

cbeer:
Darcy wants to use the "resubmit" label still


app/components/aeon/request_actions_component.html.erb
```diff
@@ -16,7 +16,8 @@
         <% else %>
           <i class="bi bi-calendar-plus me-1 align-middle"></i> Add appointment
         <% end %>
-
+      <% elsif request.cancelled? %>
```
taylor-steve, line 19: There's a `request.cancelled?` in the case above this

taylor-steve (approved):

PR #4051, opened 2026-07-13
Derive day-only appointment time server-side from reading room hours

taylor-steve:
Avoids the refreshAvailability roundtrip with day only appointments.

Adds server-side appointment time validation, which required a bit of re-arranging of the appointment deconfliction code.

### DIFF
--- a/app/views/aeon_reading_rooms/available.html.erb
+++ b/app/views/aeon_reading_rooms/available.html.erb
@@ -1,28 +1,24 @@
 <%= form_with(model: Aeon::Appointment.new) do |f| %>
   <turbo-frame id="available_appointments">
-    <% unless @reading_room.day_only_appointments? %>
-      <div class="appointment-slots-container" style="--appointment-slots: <%= @available_appointments.count %>">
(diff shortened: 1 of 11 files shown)


app/views/aeon_reading_rooms/available.html.erb
```diff
@@ -22,10 +17,12 @@
     <% end %>
     <% if params[:date] && next_appt = @available_appointments.find { |x| x.start_time.to_date > @date } %>
       <div>There are no appointments on <%= l(@date, format: :short) %>. The next available appointment is on <%= l(next_appt.start_time, format: :date_only) %>.</div>
-      <label class="visually-hidden-focusable required-placeholder">
-        No avaiable appointments.
-        <%= f.text_field :start_time, class: 'visually-hidden-focusable', required: true %>
-      </label>
+      <% unless @reading_room.day_only_appointments? %>
```
dnoneill: We need this for day only appointments for if the race condition of availability doesn't load on time and the reading room is booked

dnoneill:
I think the idea is so we don't have to do a bunch of conditionals for day only. Also if we are going with this approach I think it is kind of odd to do this check for every day but then not have anything. I think we might want to revisit having all the dates greyed out if we are going with this approach and then not even loading the turbo-frame

PR #4059, opened 2026-07-13
Pull cancel out of the folio edit form.

cbeer:


app/components/folio/request_component.html.erb
```diff
@@ -67,6 +67,35 @@
...
+              <button class="btn btn-link su-underline p-0" data-bs-toggle="modal" data-bs-target="#<%= dom_id(request, :delete) %>">
+                <i class="bi bi-trash align-middle" aria-hidden></i><span class="visually-hidden">Cancel request</span>
+              </button>
+
+              <!-- Delete modal -->
+              <%= render ModalComponent.new(id: dom_id(request, :delete)) do |component| %>
+                <% component.with_title.with_content('Delete request?') %>
+                <% component.with_body do %>
+                  <div class="d-flex flex-column gap-3">
+                    <div class="text-lagunita-dark">Pickup</div>
```
dnoneill, line 79: Can't this also be scan?

cbeer: No. Scans are tracked in Illiad.

dnoneill: Also I had to do pretty similar code for ill. I wonder if there is a way to get this into one partial. But maybe it is not worth it.

taylor-steve (approved):

PR #4060, opened 2026-07-13
Disable expensive cancelled appointment check

taylor-steve:
This is doing an Aeon requests fetch, which is the more time consuming Aeon request, in places where it's unexpected to need to fetch requests (e.g., the "new appointment" page).

Caching might solve this? Alternatively we could remediate Aeon appointments and treat the Aeon appointment status as the source of truth.

### DIFF
--- a/app/models/aeon/appointment.rb
+++ b/app/models/aeon/appointment.rb
@@ -70,7 +70,9 @@ module Aeon
     end
 
     def cancelled?
-      appointment_status == 'Cancelled' || (requests.none? && !editable?)
+      appointment_status == 'Cancelled'
+      # TODO: Restore or delete this requests/editable check (e.g., after caching, remediation)
+      # appointment_status == 'Cancelled' || (requests.none? && !editable?)
     end
 
     def edit_policy
--- a/spec/models/aeon/appointment_spec.rb
+++ b/spec/models/aeon/appointment_spec.rb
@@ -85,6 +85,7 @@ RSpec.describe Aeon::Appointment do
     end
 
     it 'is cancelled if the appointment is not editable and has no requests assigned' do
+      pending 'an approach that determines emptiness without an expensive requests fetch'
       appointment = build(:aeon_appointment, requests: [], start_time: 1.week.ago)
 
       expect(appointment).to be_cancelled


app/models/aeon/appointment.rb
```diff
@@ -70,7 +70,9 @@ def sort_key
     end
 
     def cancelled?
-      appointment_status == 'Cancelled' || (requests.none? && !editable?)
+      appointment_status == 'Cancelled'
+      # TODO: Restore or delete this requests/editable check (e.g., after caching, remediation)
+      # appointment_status == 'Cancelled' || (requests.none? && !editable?)
```
dnoneill, line 75: Do we actually want to restore this? Shouldn't we have something that actually cancels the appointment in aeon not logic that marks it as cancelled but still keeps that seat? Also if we do restore it shouldn't the logic be reversed? Seems like it should check if editable? first i.e. !editable? && request.none?

taylor-steve: I think I addressed this in my PR comment, it's a possibility that we wouldn't want to.

The conditional order should be flipped, but it doesn't short circuit the issue we're experiencing.

dnoneill: Yeah but we have this as a TODO: which implies we do want to put it back and the actual line. I am wondering if we should keep this or at least put more context in the TODO because we might not remember what this is doing/the goal

taylor-steve: I always leave myself wiggle room: `TODO: Restore or delete` 🤷‍♂️

dnoneill (approved):

taylor-steve:
Okay I'm bringing this back. I don't see why the availability check or the create new appointment flow need all of a user's requests.

Alternatively we could stop rejecting here, but then we'd have to reject at each use site (maybe 4-6 of those?):
https://github.com/sul-dlss/sul-requests/blob/2908f2a7e87ab81702129d1580fb993ffc49819f/app/models/aeon/user.rb#L76

cbeer:
Is it really worth speeding up those two API calls when the rest of your experience is still slow? I guess I'd rather look at #3975 as a solution (or doing something special for those two routes) instead of losing this behavior.

PR #4062, opened 2026-07-13
Only load activities if we need to.

cbeer:
Alternative to #4060 ?

### DIFF
--- a/app/components/aeon/remove_from_activity_button_component.rb
+++ b/app/components/aeon/remove_from_activity_button_component.rb
@@ -14,7 +14,7 @@ module Aeon
     end
 
     def call
-      form_with(url: aeon_request_path(request), method: :delete) do
+      form_with(url: aeon_request_path(request, activity: true), method: :delete) do
         tag.button(type: :submit, class: 'btn btn-link su-underline') do
           tag.i(class: 'bi bi-trash align-middle me-1', aria: { hidden: true }) +
             tag.span("Remove #{request.title} from activity", class: 'visually-hidden')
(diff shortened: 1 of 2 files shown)


app/components/aeon/remove_from_activity_button_component.rb
```diff
@@ -14,7 +14,7 @@ def render?
     end
 
     def call
-      form_with(url: aeon_request_path(request), method: :delete) do
+      form_with(url: aeon_request_path(request, activity: true), method: :delete) do
```
dnoneill, line 17: can't we just put the kind on here?
```suggestion
      form_with(url: aeon_request_path(request, kind: 'activity'), method: :delete) do
```

cbeer: We could, but I thought it'd be confusing to overload the request index filtering with "no, i swear this request is on the activities page" hack 🤷‍♂️

taylor-steve:
#4060 is trying to get at things like this requests fetch on `aeon_appointments/new`:

```
[AeonClient] GET Users/sctaylor%40stanford.edu -> 200 (106ms)
[AeonClient] GET Users/sctaylor%40stanford.edu/appointments -> 200 (134ms)
[AeonClient] GET Users/sctaylor%40stanford.edu/requests -> 200 (6271ms)
[AeonClient] GET Activities -> 200 (125ms)
```

cbeer:
I guess I'm not sure what page you're on that doesn't need the Aeon request data anyway?

taylor-steve (approved):

PR #4086, opened 2026-07-14
make banner work when someone updates the duration/date

dnoneill:
Before:

https://github.com/user-attachments/assets/3a87e9f6-c19b-4ae7-ad70-31f4163b05e1

If we just put a click action on duration:

https://github.com/user-attachments/assets/d17d5579-e7f9-4ff1-96a5-2cebf241a767

This PR:

https://github.com/user-attachments/assets/e9338fe0-9336-493b-b778-9cdefe18ec18

### DIFF
--- a/app/javascript/controllers/appointment_controller.js
+++ b/app/javascript/controllers/appointment_controller.js
@@ -80,6 +80,7 @@ export default class extends Controller {
     const hasDate = !!formData.get('aeon_appointment[date]');
     const hasDuration = !!formData.get('aeon_appointment[duration]');
     this.availabilityTarget.classList.toggle('form-incomplete', !hasDate || !hasDuration);
+    this.updateBanner()
   }
 
   applyDurationFilter() {
--- a/app/views/aeon_appointments/_form.html.erb
+++ b/app/views/aeon_appointments/_form.html.erb
@@ -25,7 +25,7 @@
       <div class="d-flex align-items-center gap-3">
         <%= render DatePickerComponent.new(:date, form: f,
               schedule: Aeon::AppointmentSchedule.new(reading_room: @appointment.reading_room, appointment_id: @appointment.id, existing_appointments: existing_appointments),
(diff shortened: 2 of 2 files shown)


app/javascript/controllers/appointment_controller.js
```diff
@@ -76,6 +76,7 @@ export default class extends Controller {
     const hasDate = !!formData.get('aeon_appointment[date]');
     const hasDuration = !!formData.get('aeon_appointment[duration]');
     this.availabilityTarget.classList.toggle('form-incomplete', !hasDate || !hasDuration);
+    this.updateBanner()
```
taylor-steve, line 83: Would it be better to add `appointment#updateBanner` to the existing actions here: https://github.com/sul-dlss/sul-requests/blob/72441c92bdbbbd2dc1672b0a0752739f6c927c6a/app/components/aeon/appointment_duration_selector_component.html.erb#L7

I think adding `updateBanner` to `updateFormStatus` runs it twice in some circumstances. I guess you could argue the reverse, remove all the existing stimulus actions and drive it all off the form status.

dnoneill: There is a video in the text of the PR for what happens if we put just a click action on duration (i.e. what you are proposing). It does not update correctly when you change the day because the availability is loading so there is no start time, even though there is when it is finished loading.

taylor-steve: Fair enough. Any hope of avoiding:
* the double execution
* the dual conventions

dnoneill: I will see what I can do.

dnoneill: Can't figure out a way around dual conventions. We are going to need the click on the appointments, unless there is something clever I haven't thought of. The double execution should be gone now.

taylor-steve (approved):

PR #4104, opened 2026-07-15
fix scan not working with checked out items due to required fields being enabled

dnoneill:
The fulfillment type is getting enabled after it is disabled. This only happens on items that have the fulfillment_preference because the fulfillment_preference is getting enabled by   showHideItemGroups. We need to just not enable if pickup isn't checked/not enable if scan isn't checked.
closes #4102

### DIFF
--- a/app/javascript/controllers/patron_request_controller.js
+++ b/app/javascript/controllers/patron_request_controller.js
@@ -32,9 +32,15 @@ export default class extends Controller {
   }
 
   showHideItemGroups(event) {
+    const formData = new FormData(this.element);
+    const requestType = formData.get('patron_request[request_type]');
     const itemGroups = this.element.querySelectorAll('.selected-items-container');
 
     Array.from(itemGroups).forEach(itemGroup => {
+      const itemGroupType = itemGroup.closest('.accordion-item').dataset.patronrequestForrequesttype
+      // We don't want to enable required pickup fields when the user has selected scan and visa versa
+      if (itemGroupType != requestType) { return }
+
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/patron_request_controller.js
```diff
@@ -32,10 +32,15 @@ export default class extends Controller {
...
   showHideItemGroups(event) {
+    const formData = new FormData(this.element);
+    const requestType = formData.get('patron_request[request_type]');
     const itemGroups = this.element.querySelectorAll('.selected-items-container');
 
     Array.from(itemGroups).forEach(itemGroup => {
-      if (!itemGroup.querySelector('[data-content-id]')) {
+      const itemGroupType = itemGroup.closest('.accordion-item').dataset.patronrequestForrequesttype
+      if (itemGroupType != requestType) {
+        // Do not enable pickup fields if scan is selected and visa versa
```
taylor-steve: The no-op here makes the conditional read a bit awkwardly. Seems like an early return, `if (itemGroupType !== requestType) return`, might be nicer.

dnoneill: Yeah but it will break the loop won't it?

taylor-steve: I would expect it to exit the callback (a closure), not the loop.

dnoneill: Ah. I see forEach is not really for a loop in the traditional sense. Okay done.

taylor-steve (approved):

PR #4110, opened 2026-07-16
refactor illiad requests params into illiadclient

dnoneill:
This also fixes the fact that scan will show up as pickup item on main

### DIFF
--- a/app/controllers/ill_requests_controller.rb
+++ b/app/controllers/ill_requests_controller.rb
@@ -20,24 +20,11 @@ class IllRequestsController < ApplicationController
     authorize! :create, Illiad::Request
   end
 
-  def create # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
+  def create # rubocop:disable Metrics/AbcSize
     authorize! :create, Illiad::Request
 
-    illiad_create_params = create_params.except(:title, :author, :NotWantedAfter, :AcceptAlternateEdition, :Note).to_h.merge(
-      ProcessType: 'Borrowing',
-      WebRequestForm: 'LoanRequest',
-      RequestType: 'Loan',
-      LoanTitle: create_params[:title],
-      LoanAuthor: create_params[:author],
-      Username: current_patron.username,
-      UserInfo1: current_patron.blocked? ? 'Blocked' : nil,
-      UserInfo5: current_patron.barcode,
-      AcceptAlternateEdition: ActiveModel::Type::Boolean.new.cast(create_params[:AcceptAlternateEdition]) ? 'E Version Acceptable' : '',
-      NotWantedAfter: not_wanted_after_param.strftime('%Y-%m-%d')
-    ).compact_blank
-
-    request = illiad_client.create(illiad_create_params)
-    illiad_client.create_transaction_note(transaction_number: request.key, note: create_params[:Note]) if create_params[:Note].present?
+    request = illiad_client.create(IlliadClient::RequestData.with_defaults.with(**request_data_params).as_json)
+    illiad_client.create_transaction_note(transaction_number: request.key, note: create_params[:note]) if create_params[:note].present?
 
     redirect_to root_path, notice: 'Your request has been submitted to Interlibrary Loan.'
   end
@@ -53,6 +40,14 @@ class IllRequestsController < ApplicationController
   end
 
   private
+
+  def request_data_params
+    static_params = { patron: current_patron }
+    if create_params[:accept_alternate_edition].present?
+      static_params['accept_alternate_edition'] = ActiveModel::Type::Boolean.new.cast(create_params[:accept_alternate_edition])
+    end
+    create_params.except(:note).to_h.compact_blank.merge(static_params)
+  end
 
   def illiad_client
     @illiad_client ||= IlliadClient.new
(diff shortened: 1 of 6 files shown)


app/services/illiad_client.rb
```diff
@@ -53,6 +53,67 @@ def update_request_route(transaction_number:, status:)
...
+  RequestData = Data.define(:call_number, :accept_alternate_edition, :patron, :isbn, :publisher, :destination_library_code, :title, :author,
+                            :origin_library_code, :origin_location_code, :scan_authors, :scan_title, :scan_page_range, :needed_date,
+                            :published_location, :published_date, :edition, :oclcn, :item_url, :barcode, :volume, :request_type) do
+    def as_json # rubocop:disable Metrics/AbcSize,Metrics/MethodLength,Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+      {
+        ProcessType: 'Borrowing',
+        RequestType: scan? ? 'Article' : 'Loan',
+        LoanTitle: scan? ? UNSET : title,
+        LoanAuthor: scan? ? UNSET : author,
+        NotWantedAfter: (needed_date || 1.year.from_now).strftime('%Y-%m-%d'),
```
taylor-steve: I wonder if we should rework this one a bit. `needed_date` could be UNSET by default here. UNSET is truthy, which leads to a potentially strange behavior?

```
sul-requests(dev):002> (IlliadClient::UNSET || 1.year.from_now).strftime('%Y-%m-%d')
=> #<data>
```

spec/models/patron_request_spec.rb
```diff
@@ -505,6 +505,14 @@
       let(:scan_page_range) { '1-10' }
 
       it 'returns the correct ILLiad request params for scan request' do
+        expect(request.illiad_request_params(item).keys).to eq(
```
taylor-steve:
```suggestion
        expect(request.illiad_request_params(item).keys).to contain_exactly(
```

Since I'm leaving another comment I'll ask for this one. Might be nice to be permissive of different key ordering.

dnoneill: Done

app/services/illiad_client.rb
dnoneill: Agreed. I didn't think about that use case. I went ahead and reworked so it should work with unset dates

app/controllers/ill_requests_controller.rb
```diff
@@ -20,24 +20,10 @@ def new
...
-    illiad_create_params = create_params.except(:title, :author, :NotWantedAfter, :AcceptAlternateEdition, :Note).to_h.merge(
-      ProcessType: 'Borrowing',
-      WebRequestForm: 'LoanRequest',
-      RequestType: 'Loan',
-      LoanTitle: create_params[:title],
-      LoanAuthor: create_params[:author],
-      Username: current_patron.username,
-      UserInfo1: current_patron.blocked? ? 'Blocked' : nil,
-      UserInfo5: current_patron.barcode,
-      AcceptAlternateEdition: ActiveModel::Type::Boolean.new.cast(create_params[:AcceptAlternateEdition]) ? 'E Version Acceptable' : '',
```
dnoneill, line 35: This breaks the form. I don't believe we can do this. It needs to be a boolean

app/controllers/ill_requests_controller.rb
```diff
@@ -20,24 +20,10 @@ def new
...
-      NotWantedAfter: not_wanted_after_param.strftime('%Y-%m-%d')
-    ).compact_blank
-
-    request = illiad_client.create(illiad_create_params)
-    illiad_client.create_transaction_note(transaction_number: request.key, note: create_params[:Note]) if create_params[:Note].present?
```
taylor-steve, line 40: Do we need to preserve this? It's recent, from 5929ee6c

dnoneill: Yeah. It got missed in the rebase.

app/services/illiad_client.rb
```diff
@@ -60,6 +60,81 @@ def update_request_route(transaction_number:, status:)
...
+    def method_missing(*, **, &)
+      self # always return the unset object itself so it chains without error
+    end
+
+    def respond_to_missing?(*)
+      false
+    end
+  end.new
+
+  RequestData = Data.define(:accept_alternate_edition, :author, :barcode, :call_number,
```
cbeer, line 73: I question whether we want to do the key translation here or up another later. In AeonClient::RequestData, we kept these generally matching the actual API keys.

dnoneill: I think there is value in knowing what iteminfo2 is and not having to figure it out. I can't take a quick glance at the aeon forms and know what that actually is.

dnoneill: Also it looks like we alias them in our models. i.e. https://github.com/sul-dlss/sul-requests/blob/35d7fab0942bbc0133b39020d97d672317e9c22a/app/models/aeon/request.rb#L67. Something we aren't doing at all in Illiad. I can do that but I think the reason the model in illiad doesn't name library_code as item_info4 is it is confusing  https://github.com/sul-dlss/sul-requests/blob/35d7fab0942bbc0133b39020d97d672317e9c22a/app/models/illiad/request.rb#L145-L147

PR #4138, opened 2026-07-20
swap digits to be a reverse sort on reqeust dates

dnoneill:
part of #4121

### DIFF
--- a/app/models/aeon/request.rb
+++ b/app/models/aeon/request.rb
@@ -195,7 +195,7 @@ module Aeon
                  when :title
                    [title, default_sort_key]
                  when :date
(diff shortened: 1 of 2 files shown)


app/models/aeon/request.rb
```diff
@@ -205,6 +205,16 @@ def sort_key(key = nil) # rubocop:disable Metrics/AbcSize
       sort_key.join('---')
     end
 
+    def swap_digits(digit)
+      # Swaps each digit d with (9 - d)
+      # 0<->9, 1<->8, 2<->7, 3<->6, 4<->5
+      str = digit.to_i.to_s
```
cbeer, line 211: I don't know why you're going through an int here. Here's the approach I was trying to suggest: https://github.com/sul-dlss/sul-requests/pull/4139

PR #4141, opened 2026-07-20
Extract a service to figure out the eligible service points for a lis…

cbeer:
…t of folio items


app/models/patron_request.rb
```diff
@@ -222,23 +222,9 @@ def pickup_service_point
...
-    eligible_destinations = item_pickup_destinations.reject do |destination|
-      if patron.blank?
-        Settings.allowed_visitor_pickups.exclude?(destination.code)
-      else
-        destination.unpermitted_pickup_groups.include?(patron.patron_group_name)
-      end
-    end
-
-    eligible_destinations.map(&:code)
+    @pickup_destinations ||= FolioRequestServicePointOptionsService.new(selectable_items, patron:).possible_service_points.map(&:code)
```
taylor-steve, line 226: It seems like `possible_service_points` doesn't take into consideration the patron's restrictions (as far as I can tell, it doesn't use `patron_is_eligible_for_service_point?`), but the old code and spec did. Is that intentional?

cbeer: I pushed those checks down into `FolioRequestServicePointOptionsService#default_pickup_service_points` and  `FolioRequestServicePointOptionsService#additional_pickup_service_points`, but I guess I missed it for location-restricted items

taylor-steve (approved):

PR #4151, opened 2026-07-21
add edit modal to illiad requests

dnoneill:
Screenshot 2026-07-21 at 3 50 18 PM

Screenshot 2026-07-22 at 1 47 29 PM

### DIFF
--- a/app/controllers/ill_requests_controller.rb
+++ b/app/controllers/ill_requests_controller.rb
@@ -4,9 +4,13 @@
 class IllRequestsController < ApplicationController
   include FolioController
 
+  layout 'application_redesign'
+
   before_action :authenticate_user!
 
   before_action :load_requests
+  before_action :load_request, only: [:destroy, :edit, :update]
+  before_action :set_variant, only: [:edit]
 
   # Renders user requests from ILL sources
   #
@@ -18,6 +22,10 @@ class IllRequestsController < ApplicationController
 
   def new
     authorize! :create, Illiad::Request
+  end
+
+  def edit
+    authorize! :edit, @request
   end
 
   def create # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
@@ -46,8 +54,16 @@ class IllRequestsController < ApplicationController
     redirect_to root_path, notice: 'Your request has been submitted to Interlibrary Loan.'
   end
 
+  def update
+    authorize! :update, @request
+    IlbMailer.updated_ilb_notification(current_patron, @request, updated_fields).deliver_now unless updated_fields.empty?
+    respond_to do |format|
+      format.html { redirect_to unified_requests_path, notice: 'Request updated successfully' }
+      format.turbo_stream
+    end
+  end
+
   def destroy
-    load_request
     authorize! :destroy, @request
     @response = illiad_client.update_request_route(transaction_number: @request.id, status: Settings.illiad.cancelled_by_user)
     respond_to do |format|
@@ -57,6 +73,10 @@ class IllRequestsController < ApplicationController
   end
 
   private
+
+  def updated_fields
+    update_params.to_h.compact_blank.filter { |param, value| @request.public_send(param)&.to_s&.strip != value.strip }
+  end
 
   def illiad_client
     @illiad_client ||= IlliadClient.new
@@ -81,6 +101,11 @@ class IllRequestsController < ApplicationController
     Date.parse(create_params[:not_wanted_after])
   rescue StandardError
     nil
+  end
+
+  def update_params
+    params.require(:illiad_request).permit(:not_wanted_after, :item_info4, :photo_article_title,
+                                           :photo_article_author, :photo_journal_inclusive_pages)
   end
 
   def create_params
--- a/app/models/illiad/request.rb
+++ b/app/models/illiad/request.rb
@@ -41,14 +41,38 @@ module Illiad
       INACTIVE_REQUEST_STATUSES.include? @illiad_result['TransactionStatus']
     end
 
+    def request_type
+      scan? ? 'scan' : 'pickup'
+    end
+
+    def item_type
+      @illiad_result['RequestType']
+    end
+
     def scan?
       @illiad_result['PhotoJournalTitle'].present?
+    end
+
+    def photo_article_title
+      @illiad_result['PhotoArticleTitle']
+    end
+
+    def photo_journal_inclusive_pages
+      @illiad_result['PhotoJournalInclusivePages']
+    end
+
+    def photo_article_author
+      @illiad_result['PhotoArticleAuthor']
     end
 
     def key
       @illiad_result['TransactionNumber'].to_s
     end
     alias id key
+
+    def item_link
+      @illiad_result['CitedIn']
+    end
 
     def sort_key(key)
       sort_key = case key
@@ -75,7 +99,7 @@ module Illiad
     end
 
     def author
-      scan? ? @illiad_result['PhotoArticleAuthor'] : @illiad_result['LoanAuthor']
+      @illiad_result['LoanAuthor'] unless scan?
     end
 
     def placed_date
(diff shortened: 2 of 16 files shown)


config/routes.rb
```diff
@@ -61,7 +61,8 @@
   get 'requests/unified', to: 'requests#index', as: :unified_requests
   get 'requests/mediated', to: 'mediated_requests#index', as: :mediated_requests
   resources :folio_requests, except: [:new, :create], path: 'requests'
-  resources :ill_requests, only: [:index, :new, :create, :destroy]
+  resources :ill_requests, only: [:index, :edit, :new, :create, :destroy, :update]
```
cbeer:
```suggestion
  resources :ill_requests
```

Isn't that all of them?

app/views/ill_requests/_edit_pickup_form.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="form-group mb-3">
+  <div>
+    <%= label_tag :service_point_id, 'Pickup point', class: 'form-label fw-semibold' %>
+    <%= select_tag :service_point_id, request_location_options(@request, patron_or_group), class: 'form-control' %>
```
cbeer: Is there a reason we're using form helpers in `_edit_scan_form` but the tag helpers here? I'd think form helpers would be preferred?

app/controllers/ill_requests_controller.rb
```diff
@@ -74,6 +85,10 @@ def not_wanted_after_param
     1.year.from_now
   end
 
+  def update_params
+    params.permit(:illiad_request, :scan_title, :scan_authors, :scan_page_range, :service_point_id, :fill_by_date)
```
cbeer: You used to have this as `params.require(:illiad_request).permit(:scan_title, :scan_authors, :scan_page_range, :request_type)`, which seems more standard. Why the change?

app/models/illiad/request.rb
```diff
@@ -41,15 +41,39 @@ def inactive?
...
       @illiad_result['TransactionNumber'].to_s
     end
     alias id key
 
+    def item_link
+      @illiad_result['CitedIn']
+    end
+
+    def hrid
+      item_link.chomp('/').split('/').last
```
cbeer: Should we at least make sure the link is a searchworks link? Using the forms in ILLiad, users could have put anything here.

app/models/illiad/request.rb
```diff
@@ -75,7 +99,7 @@ def call_number
     end
 
     def author
-      scan? ? @illiad_result['PhotoArticleAuthor'] : @illiad_result['LoanAuthor']
+      scan? ? nil : @illiad_result['LoanAuthor']
```
cbeer:
```suggestion
      @illiad_result['LoanAuthor'] unless scan?
```

would be a little more idiomatic.

app/views/ill_requests/_edit_pickup_form.html.erb
dnoneill: It was because it seemed like there is some weirdness with form helpers with this but I will double check.

app/controllers/ill_requests_controller.rb
dnoneill: Because I added the pickup which doesn't use form helpers. If I can get it back in it will go back

app/models/illiad/request.rb
dnoneill: Do we display user's illiad new requests on their account? If so yeah I can do a check and they can't edit that.

app/views/ill_requests/_edit_pickup_form.html.erb
dnoneill: Okay no weirdness. Added in.

config/routes.rb
dnoneill: lol. Wasn't even thinking about that but yeah

app/models/illiad/request.rb
```diff
@@ -91,13 +117,13 @@ def expiration_date
...
-      scan? ? based_on_placed : (user_supplied_expiration_date || based_on_placed)
+      scan? ? based_on_placed : (fill_by_date || based_on_placed)
     end
 
-    def user_supplied_expiration_date
+    def fill_by_date
       return nil if @illiad_result['NotWantedAfter'].blank?
 
-      parse_expiration_date(@illiad_result['NotWantedAfter'])
+      parse_expiration_date(@illiad_result['NotWantedAfter']).to_date
```
cbeer: It looks like `parse_expiration_date` tries to gracefully handle when `NotWantedAfter` isn't really a date. Should we be doing that here too, instead of raising a nil error?

app/models/illiad/request.rb
```diff
@@ -41,15 +41,41 @@ def inactive?
...
     def key
       @illiad_result['TransactionNumber'].to_s
     end
     alias id key
 
+    def item_link
+      @illiad_result['CitedIn']
+    end
+
+    def instance_hrid
```
cbeer: We're only using `instance_hrid` to figure out the eligible service points? If it's an ILL request, it seems like we could fall back on the default list because the FOLIO-specific rules don't really apply.

dnoneill: Okay yeah I guess that makes sense since it is going through ILL it won't matter. Thanks for the information.

config/routes.rb
dnoneill: We forgot show. I am changing it back.

app/views/ill_requests/new.html.erb
```diff
@@ -133,17 +133,7 @@
...
-          <%= f.select(
-          :item_info4, options_for_select(default_service_points(current_user.patron).map { |sp| [sp.name, sp.code] }), { include_blank: false, required: false },
-          class: 'form-select'
-          ) %>
-        </div>
-        <div class="mb-4">
-          <%= f.label :not_wanted_after, 'Not wanted after date', class: 'form-label fw-semibold' %>
-          <%= f.date_field :not_wanted_after, class: 'form-control' %>
-        </div>
+        <%= render 'edit_pickup_form', f: %>
```
cbeer: If we're trying to use these in both places, maybe `pickup_form` + `scan_form` are better names?

config/routes.rb
```diff
@@ -61,7 +61,8 @@
   get 'requests/unified', to: 'requests#index', as: :unified_requests
   get 'requests/mediated', to: 'mediated_requests#index', as: :mediated_requests
   resources :folio_requests, except: [:new, :create], path: 'requests'
-  resources :ill_requests, only: [:index, :new, :create, :destroy]
+  resources :ill_requests, only: [:index, :new, :create, :destroy, :edit, :update]
```
cbeer, line 64:
```suggestion
  resources :ill_requests
```

Isn't that all of them?

dnoneill: Nope. add :show and that is all of them.

spec/features/ill_requests_spec.rb
```diff
@@ -12,23 +12,50 @@
...
+    end
+
+    it 'can edit the request' do
+      visit ill_requests_path
+      click_link 'Edit Pickup title'
+      expect(page).to have_text 'Edit request'
+      select 'Music Library', from: 'Pickup point'
+      click_button 'Save'
+      wait_for_email
+      expect(ActionMailer::Base.deliveries.count).to eq(1)
```
cbeer, line 45:
```suggestion
      expect do
        click_button 'Save'
      end.to change { ActionMailer::Base.deliveries.count }.by(1)
```

Waiting is always kinda shady. Can we just tell rspec what we're waiting for?

dnoneill: Nope. If we don't wait we get the following error
```expected `ActionMailer::Base.deliveries.count` to have changed by 1, but was changed by 0```
       
I don't love putting the wait_for_email in but I think it is important to test the email since that is the main thing we are doing. I guess we could test the mailer elsewhere but I think an end to end test is kind of nice.

cbeer: I don't know, it just strikes me as really weird this is the way to test that.

cbeer: Ah, I see the issue(s):

- the test wasn't operating with the redesign layout (so the modal link opened in a new tab)
- if we wait for some change on the page (in #4182, i just tacked on an invisible .updated element for example), the mailer test works without the delay. If we hate that, we could check for something ugly (`expect(page).to have_css('.modal:not(.show) turbo-frame[complete]', visible: :hidden)`), but 🤷

dnoneill: I tried that but I must not have been specific enough with an element. I don't like the idea of adding a random element for testing only. But this gives me a good place to start.

dnoneill: Okay I went with the modal header not appearing on the screen and it works at least locally

PR #4160, opened 2026-07-22
Add an association between PatronRequestItem and the API response.

cbeer:
Needs to wait until we're passing PatronRequestItem further down the stack (#4161) so we can create new `ApiResponse` data.


db/migrate/20260722175759_add_patron_request_item_id_to_api_response.rb
```diff
@@ -0,0 +1,24 @@
+class AddPatronRequestItemIdToApiResponse < ActiveRecord::Migration[8.1]
+  def change
+    add_reference :api_responses, :patron_request_item, null: true, foreign_key: true
+  end
+
+  reversible do |dir|
+    dir.up do
```
taylor-steve: Does the `reversible` block need to be within `change`?

cbeer: Oh, probably.

taylor-steve (approved):

PR #4163, opened 2026-07-22
remove title/intro text from login page

dnoneill:
closes #4159


app/views/patron_requests/login.html.erb
```diff
@@ -1,15 +1,5 @@
 <% content_for :page_title do %>Request: <%= @patron_request.item_title %><% end %>
 
-<h1 class="fw-semibold my-3 fs-1">
```
cbeer, line 3: Is this template shared with old-style requests? If so, we should only remove this if we're in the redesign

dnoneill: I thought it was for everything, like this could go live now. Let me check with Darcy

PR #4166, opened 2026-07-22
ILL form updates

cbeer:
... at least as a ~first~ second pass.

### DIFF
--- a/app/helpers/requests_location_helper.rb
+++ b/app/helpers/requests_location_helper.rb
@@ -22,4 +22,12 @@ module RequestsLocationHelper
                                                selected_service_point_id: request.service_point_id,
                                                patron:).possible_service_points
   end
+
+  # Generates a list of default service point options for a patron.
+  #
+  # @param patron [Folio::Patron] The patron object
+  # @return [Array<Folio::ServicePoint>]
+  def default_service_points(patron)
+    FolioRequestServicePointOptionsService.new([], patron:).default_pickup_service_points.sort_by(&:sort_key)
+  end
 end
--- a/app/javascript/controllers/ill_request_controller.js
+++ b/app/javascript/controllers/ill_request_controller.js
@@ -1,0 +1,52 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ['section'];
+  static values = { type: { type: String, default: 'none' }, subtype: { type: String, default: 'book' } }
+
+  connect() {
+    this.showHideSections();
+  }
+
+  updateType(event) {
+    this.typeValue = event.target.value;
+  }
+
+  typeValueChanged() {
+    if (this.typeValue == 'pickup') {
+      this.subtypeValue = 'book';
+    }
+
+    this.showHideSections();
+  }
+
+  updateSubtype(event) {
+    this.subtypeValue = event.target.value;
+    this.showHideSections();
+  }
+
+  showHideSections() {
+    Array.from(this.sectionTargets).forEach(section => {
+      if ((section.dataset.illRequestForRequestType == undefined || section.dataset.illRequestForRequestType == this.typeValue) && (section.dataset.illRequestForRequestSubtype == undefined || section.dataset.illRequestForRequestSubtype == this.subtypeValue)) {
+        const template = section.querySelector('template[data-request-type-template]');
+
+        if (template) {
+          const content = template.content.cloneNode(true);
+          section.appendChild(content);
+          template.remove();
+        }
+
+        section.classList.remove('d-none');
+      } else if (section.dataset.illRequestForRequestType || section.dataset.illRequestForRequestSubtype) {
+        section.classList.add('d-none');
+
+        if (!section.querySelector('template[data-request-type-template]')) {
+          const template = document.createElement('template');
+          template.dataset.requestTypeTemplate = true;
+          template.content.append(...section.childNodes);
+          section.appendChild(template);
+        }
+      }
+    });
+  }
+}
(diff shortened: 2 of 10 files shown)


app/helpers/requests_location_helper.rb
```diff
@@ -22,4 +22,12 @@ def eligible_service_points(request, patron)
                                                selected_service_point_id: request.service_point_id,
                                                patron:).possible_service_points
   end
+
+  # Generates a list of default service point options for a patron.
+  #
+  # @param patron [Folio::Patron] The patron object
+  # @return [Array<Folio::ServicePoint>]
+  def default_service_points(patron)
+    FolioRequestServicePointOptionsService.new([], patron:).default_pickup_service_points.sort_by(&:sort_key)
```
dnoneill, line 31: `Science Library (Li and Ma)` is showing up with this which I am not sure is a valid pickup location? I have never seen it in the app before

cbeer: I'm not seeing any problem with `sort_by`, and I'm not sure how `default_pickup_service_points` could be a single value. 

I'd expect science to be a valid pickup location, along with a bunch of other libraries. Here's the list for an ordinary page:
Screenshot 2026-07-23 at 09 05 46

dnoneill: Yeah I went and double checked and it was there. And the sort_by was the branch was old.

app/javascript/controllers/ill_request_controller.js
```diff
@@ -0,0 +1,55 @@
...
+
+        if (template) {
+          const content = template.content.cloneNode(true);
+          section.appendChild(content);
+          template.remove();
+        }
+
+        section.classList.remove('d-none');
+      } else if (section.dataset.illRequestForRequestType || section.dataset.illRequestForRequestSubtype) {
+        console.log('hiding section for not matching type');
```
dnoneill: cruft?

app/javascript/controllers/ill_request_controller.js
```diff
@@ -0,0 +1,55 @@
...
+  updateSubtype(event) {
+    this.subtypeValue = event.target.value;
+    this.showHideSections();
+  }
+
+  showHideSections() {
+    Array.from(this.sectionTargets).forEach(section => {
+      console.log(section, this.typeValue, section.dataset, this.subtypeValue);
+      if ((section.dataset.illRequestForRequestType == undefined || section.dataset.illRequestForRequestType == this.typeValue) && (section.dataset.illRequestForRequestSubtype == undefined || section.dataset.illRequestForRequestSubtype == this.subtypeValue)) {
+        console.log('showing section for matching type');
```
dnoneill: cruft?

app/views/ill_requests/_scan_or_pickup.html.erb
```diff
@@ -0,0 +1,22 @@
+
```
dnoneill: extra space?

app/views/ill_requests/new.html.erb
```diff
@@ -3,93 +3,211 @@
...
+      <% c.with_title.with_content('Scan or pickup request') %>
+    <% end %>
+
+    <%= render AccordionStepComponent.new(id: 'pickup-request', classes: ['d-none'], step_index: scan_or_pickup_step ||= step_enum.next, form_id: 'illiad-request-form', data: { ill_request_target: 'section', ill_request_for_request_type: "pickup" }, submit: true) do |c| %>
+      <% c.with_title.with_content('Pickup request') %>
+      <% c.with_body do %>
+        <div class="mb-3">
+          <%= f.label :ItemInfo4, 'Deliver to', class: 'form-label d-block fw-semibold redesign-required-label' %>
+          <%= f.select(
+          :ItemInfo4, options_for_select(default_service_points(current_user.patron).map { |sp| [sp.name, sp.id] }), { include_blank: false, required: false },
```
dnoneill:
```suggestion
          :ItemInfo4, options_for_select(default_service_points(current_user.patron).map { |sp| [sp.name, sp.code] }), { include_blank: false, required: false },
```

I don't think we want to send the id do we? It isn't really human readable

app/views/ill_requests/new.html.erb
```diff
@@ -3,93 +3,211 @@
...
+          class: 'form-select'
+          ) %>
+        </div>
+        <div class="mb-4">
+          <%= f.label :NotWantedAfter, 'Not wanted after date', class: 'form-label fw-semibold' %>
+          <%= f.date_field :NotWantedAfter, class: 'form-control' %>
+        </div>
+
+        <div class="mb-3" data-controller="char-counter" data-char-counter-default-class="text-body-tertiary" data-char-counter-warning-class="text-cardinal">
```
dnoneill, line 148: This is text-cardinal everywhere we implement it. I think we should just hardcode it. I can create that in a separate PR if you would prefer.

app/views/ill_requests/new.html.erb
cbeer: Don't we? That's what we're doing in e.g. `request_location_options`.

app/views/ill_requests/new.html.erb
cbeer: I guess I don't have any problem with the duplication, and appreciate the clarity (and not having css classes also hiding in the javascript)

app/views/ill_requests/new.html.erb
dnoneill: That is going into Folio. This is going into Iliad. Illiad can't do the translation

dnoneill (approved):

PR #4172, opened 2026-07-23
Update all(?) modal forms to wait for a response before closing.

cbeer:


app/components/aeon/appointment_cancel_modal_component.html.erb
```diff
@@ -2,6 +2,8 @@
   <% component.with_title.with_content('Cancel appointment?') %>
   <% component.with_body(classes: []) do %>
     <%= form_with url: aeon_appointment_path(appointment.id), method: 'delete' do |f| %>
+      <%= hidden_field_tag :modal, id %>
```
dnoneill, line 5: Does it make more sense to put this in the modal template and component so we don't have to repeat it so much?
https://github.com/sul-dlss/sul-requests/blob/7771bfba7131463833c1a6f5b6af145e63dc4a39/app/components/modal_component.html.erb#L13
https://github.com/sul-dlss/sul-requests/blob/7771bfba7131463833c1a6f5b6af145e63dc4a39/app/views/layouts/application_redesign.html.erb#L107


Also I think you missed this on the feedback form. Which kind of indicates it would be nice in one place

dnoneill: Ah. I guess it is hard to pass it in the form. Nevermind.

app/views/aeon_appointments/items.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= turbo_frame_tag params[:modal] do %>
   <%= form_with(url: update_multiple_aeon_requests_path) do |f| %>
```
dnoneill, line 2: I am wondering if it might be better to pass as a param? maybe not. But then we wouldn't have to do an if. It should be done for us with the url builder I would think

```suggestion
  <%= form_with(url: update_multiple_aeon_requests_path, modal: params[:modal]) do |f| %>
```

cbeer: I'm not sure it really matters either way. Is there a reason to change what's working?

dnoneill: Less code. not having to do the if statement/having a modal variable in the form. I know for sure that works because I did it here: https://github.com/sul-dlss/sul-requests/pull/4151/changes#diff-6973f0e6aa72be7e8213ea115ba69aacd07888f6489dfbfadd3e01dd47d5cdf0R8. So it is the difference between two things that work.

cbeer: I still don't see any reason to break from the pattern we're already doing if it's working.

dnoneill (approved):

PR #4176, opened 2026-07-23
Extract IlliadClient::RequestData

cbeer:
I think these are the parts of #4110 we agree on, without adding our own keys to the RequestData object and mirrors how AaeonClient::RequestData works.

### DIFF
--- a/app/controllers/ill_requests_controller.rb
+++ b/app/controllers/ill_requests_controller.rb
@@ -23,21 +23,27 @@ class IllRequestsController < ApplicationController
   def create # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
     authorize! :create, Illiad::Request
 
-    illiad_create_params = create_params.except(:title, :author, :NotWantedAfter, :AcceptAlternateEdition, :Note).to_h.merge(
-      ProcessType: 'Borrowing',
-      WebRequestForm: 'LoanRequest',
-      RequestType: 'Loan',
-      LoanTitle: create_params[:title],
-      LoanAuthor: create_params[:author],
-      Username: current_patron.username,
-      UserInfo1: current_patron.blocked? ? 'Blocked' : nil,
-      UserInfo5: current_patron.barcode,
-      AcceptAlternateEdition: ActiveModel::Type::Boolean.new.cast(create_params[:AcceptAlternateEdition]) ? 'E Version Acceptable' : '',
-      NotWantedAfter: not_wanted_after_param.strftime('%Y-%m-%d')
-    ).compact_blank
+    title_param = if create_params[:photo_article_title].present?
+                    { photo_journal_title: create_params[:title] }
+                  else
+                    { loan_title: create_params[:title] }
+                  end
+
+    illiad_create_params = IlliadClient::RequestData.with_defaults.with(
+      process_type: 'Borrowing',
+      web_request_form: 'LoanRequest',
+      request_type: 'Loan',
+      username: current_patron.username,
+      user_info1: current_patron.blocked? ? 'Blocked' : nil,
+      user_info5: current_patron.barcode,
+      accept_alternate_edition: create_params[:accept_alternate_edition] || IlliadClient::UNSET,
+      not_wanted_after: not_wanted_after_param || IlliadClient::UNSET,
+      **title_param,
+      **create_params.except(:title, :not_wanted_after, :accept_alternate_edition, :note).to_h.compact_blank
+    )
 
     request = illiad_client.create(illiad_create_params)
-    illiad_client.create_transaction_note(transaction_number: request.key, note: create_params[:Note]) if create_params[:Note].present?
+    illiad_client.create_transaction_note(transaction_number: request.key, note: create_params[:note]) if create_params[:note].present?
 
     redirect_to root_path, notice: 'Your request has been submitted to Interlibrary Loan.'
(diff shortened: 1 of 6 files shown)


app/services/illiad_client.rb
```diff
@@ -34,12 +34,88 @@ def initialize(url: Settings.sul_illiad, api_key: Settings.illiad_api_key)
...
+                            :loan_author, :loan_date, :loan_edition, :loan_place, :loan_publisher, :loan_title,
+                            :location, :not_wanted_after,
+                            :photo_article_author, :photo_article_title, :photo_journal_inclusive_pages,
+                            :photo_journal_volume,
+                            :photo_journal_issue, :photo_journal_month, :photo_journal_year, :photo_journal_title,
+                            :process_type, :reference_number, :request_type, :spec_ins,
+                            :username, :user_info1, :user_info2, :user_info3, :user_info4, :user_info5, :web_request_form) do
+    def as_json # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
+      {
+        AcceptAlternateEdition: ActiveModel::Type::Boolean.new.cast(accept_alternate_edition) ? 'E Version Acceptable' : '',
```
dnoneill: This breaks the form. It will not submit with this. Illiad requires a boolean

app/controllers/ill_requests_controller.rb
```diff
@@ -23,21 +23,25 @@ def new
...
+    title_param = if scan
+                    { photo_journal_title: create_params[:title] }
+                  else
+                    { loan_title: create_params[:title] }
+                  end
+
+    illiad_create_params = IlliadClient::RequestData.with_defaults.with_patron(current_patron).with(
+      process_type: scan ? 'DocDel' : 'Borrowing',
+      web_request_form: 'LoanRequest',
+      request_type: scan ? IlliadClient::UNSET : 'Loan',
```
dnoneill, line 36: Is there a reason we want to do this in the controller and not pass the request type to the RequestData? I am not quite sure why we should do this calculation everywhere instead of just doing it once in the RequestData function

app/services/illiad_client.rb
cbeer: That's really funny. This is the value it gets coming back out of ILLIad.

app/controllers/ill_requests_controller.rb
cbeer: Maybe I'm just not convinced yet that the new ILL forms and the folio-to-item ILL requests are actually the same, or, if they are  the same, we shouldn't make `ScanRequestData` and `PickupRequestData` instances.

app/services/illiad_client.rb
dnoneill: I just know when I try to submit I get this error from Illiad:
```@response_body={"Message"=>"The request is invalid.", "ModelState"=>{"model.AcceptAlternateEdition"=>["An error has occurred."]}}```

It seems to want a boolean

File: app/controllers/ill_requests_controller.rb (hunk 2 above)
Comment (dnoneill) [reply app/controllers/ill_requests_controller.rb:36]: I like the idea of having a ScanRequestData and a PickupRequestData and only difference are the with_defaults are different. That seems really nice to me.

Something along the lines of 
```
LoanRequestData = RequestData.with_defaults.with(process_type: 'Borrowing')
```

Comment (cbeer) [reply app/controllers/ill_requests_controller.rb:36]: Right. I just can't tell if we're there yet, but this is probably a stepping stone to get there or to rename some of these keys to be meaningful to the application.

Comment (dnoneill) [reply app/controllers/ill_requests_controller.rb:36]: I guess that depends on if we are just missing things between the two requests or they are purposeful. Okay I am happy with this as is. I will be doing some aliasing on edit but that is fine. Or maybe not. Will see when I get there.

File: app/models/patron_request.rb [hunk 3]
```diff
@@ -519,46 +519,41 @@ def scan_earliest
   # directed to ILLiad for fulfillment.
   # @return [Hash] the parameters to send to ILLiad for the request
   def illiad_request_params(item)
-    default_values = {
-      ProcessType: 'Borrowing',
```
dnoneill, line 523: Was this correct for scan? I do wonder since we have something different when creating a new request. Just noting. Not sure there is anything to do there.

app/views/ill_requests/new.html.erb
```diff
@@ -134,20 +134,20 @@
       <% c.with_title.with_content('Pickup request') %>
       <% c.with_body do %>
         <div class="mb-3">
-          <%= f.label :ItemInfo4, 'Deliver to', class: 'form-label d-block fw-semibold redesign-required-label' %>
+          <%= f.label :item_info4, 'Deliver to', class: 'form-label d-block fw-semibold redesign-required-label' %>
           <%= f.select(
           :ItemInfo4, options_for_select(default_service_points(current_user.patron).map { |sp| [sp.name, sp.code] }), { include_blank: false, required: false },
```
dnoneill:
```suggestion
          :item_info4, options_for_select(default_service_points(current_user.patron).map { |sp| [sp.name, sp.code] }), { include_blank: false, required: false },
```

app/models/patron_request.rb
cbeer: I don't see these in the existing scan forms.

dnoneill (approved):

PR #4178, opened 2026-07-23
no items messages

dnoneill:
closes #4158

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -687,11 +687,15 @@ summary .disclosure-icon::before {
   --bs-border-width: 0;
 }
 
-.no-requests {
+.no-requests, .no-appointments {
   display: none;
 }
 
 main:not(:has(#requests li, turbo-frame[busy])) .no-requests {
+  display: block;
+}
+
+.appointments-index:not(:has(.appointment)) .no-appointments {
   display: block;
 }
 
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/application.css
```diff
@@ -687,6 +687,14 @@ summary .disclosure-icon::before {
   --bs-border-width: 0;
 }
 
+.no-requests {
+  display: none;
+}
+
+ul#requests:not(:has(li)) ~ turbo-frame:not([busy]) ~ .no-requests {
```
taylor-steve: I understand this is the `Steve` problem but I see the no requests message for quite a while when loading the unified requests:

Image

Maybe:
```
.page-section:not(:has(#requests li, turbo-frame[busy])) .no-requests {
```

or
```
.page-section:not(:has(.card.request, .card.request-group, turbo-frame[busy])) .no-requests {
```

Maybe `.page-section` isn't the right class, but those are some ideas that seemed to work for me.

dnoneill: I am confused. How is that happening? Are you saying you have a time when your turbo-frame isn't loading? Can we huddle on this?

dnoneill: I mean that works so I will go with that. It is super weird it showing up when turbo-frames are busy

taylor-steve: I believe the sibling combinator is an "any" not "all", so in my case there were:
* no li elements, so the first condition matches
* then one of the turbo frames removes busy, so the second condition matches as well.

taylor-steve (approved):

PR #4199, opened 2026-07-24
guard for pickup_service_point being nil

dnoneill:
closes #4186

### DIFF
--- a/app/models/patron_request.rb
+++ b/app/models/patron_request.rb
@@ -642,7 +642,7 @@ class PatronRequest < ApplicationRecord
 
   # Validate that the chosen service point is a valid pickup location for the items
   def pickup_service_point_is_valid
-    return if pickup_service_point.code.in? pickup_destinations
+    return if pickup_service_point&.code.in? pickup_destinations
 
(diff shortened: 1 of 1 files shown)


app/models/patron_request.rb
```diff
@@ -642,7 +642,7 @@ def folio_pickup?
 
   # Validate that the chosen service point is a valid pickup location for the items
   def pickup_service_point_is_valid
-    return if pickup_service_point.code.in? pickup_destinations
```
cbeer, line 645: Is this what we want to do? I'm not sure how that honeybadger report got into this situation, but it feels like either the pickup service point shouldn't be nil, or it's appropriate to fail validation?

dnoneill: I couldn't figure out how honeybadger got it but when the page loads it can be nil. Also we are guarding in in the model so it makes sense to do this.

https://github.com/sul-dlss/sul-requests/blob/main/app/models/patron_request.rb#L166

dnoneill: Huh. I looks like it submitted an empty value. Hmmm. Very odd.

PR #4202, opened 2026-07-24
get rid of required for submit, set required for appointment_id, requested_pages form

dnoneill:
closes #4195

Screenshot 2026-07-27 at 12 57 19 PM

### DIFF
--- a/app/components/aeon/appointment_form_item_component.html.erb
+++ b/app/components/aeon/appointment_form_item_component.html.erb
@@ -21,8 +21,8 @@
     <div class="appt-select d-flex flex-wrap align-items-center column-gap-2 row-gap-1">
       <div class="fs-6 mb-0 w-100 form-label fw-semibold redesign-required-label possibly-visually-hidden-appointments-label">Appointment</div>
       <div class="dropdown custom-select flex-grow-1 flex-md-grow-0" data-controller="appointment-select" data-action="change->appointment-select#updateSelected input->appointment-select#updateSelected appointment-select:change@document->appointment-select#updateItemCounts">
-        <%= f.hidden_field :appointment_id, class: 'visually-hidden', data: { 'appointment-select-target': :input, 'required-for-submit': true } %>
         <%= button_tag 'Select appointment', class: 'py-1 btn border dropdown-toggle text-start d-flex align-items-center justify-content-between', type: 'button', data: { bs_toggle: 'dropdown', appointment_select_target: 'button' }, aria: { expanded: false }, disabled: selectable_appointments.blank? %>
+        <%= f.text_field :appointment_id, class: 'visually-hidden-focusable required-placeholder', data: { 'appointment-select-target': 'input' }, required: true %>
         <menu class="dropdown-menu appt-dropdown w-100">
           <li></li>
           <%= render Aeon::AppointmentOptionComponent.with_collection(selectable_appointments, data_action: 'appointment-select#select') %>
--- a/app/javascript/controllers/appointment_select_controller.js
+++ b/app/javascript/controllers/appointment_select_controller.js
@@ -18,13 +18,12 @@ export default class extends Controller {
   }
 
   updateItemCounts() {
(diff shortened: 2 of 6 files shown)


app/javascript/controllers/appointment_select_controller.js
```diff
@@ -34,6 +33,7 @@ export default class extends Controller {
     this.element.querySelector('.selected')?.classList?.remove('selected');
     element.classList.add('selected');
 
+    this.inputTarget.setAttribute('value', element.dataset.value)
```
dnoneill: You need both setAttribute and value=. I wish I knew why but formCount doesn't work without setAttribute. It is super annoying. If you don't set .value = then something else doesn't work which I can't remember. It is all not nice.

dnoneill: Okay. I figured out this problem. Unfortunately you need to update both attribute and value in a text field because updating the value updates the live value but it doesn't update the attribute in the dom the same way a hidden field updating the value updates the dom. Unfortunately in order to get things to update properly we need to do this twice. Other options include switching back to a hidden element and adding js to validate. Or updating the query to filter instead of querySelect based on value. i.e. 
`this.element.closest('#reading-accordion').querySelectorAll("input").filter(elem => elem.value == option.dataset.appointmentId)` which I think might be the better option.

```
const hidden = document.createElement('input');
hidden.type = 'hidden';
hidden.value = 'abc';
console.log(hidden.getAttribute('value')); // 'abc' — attribute updated!

const text = document.createElement('input');
text.type = 'text';
text.value = 'abc';
console.log(text.getAttribute('value')); // null — attribute untouched
```

app/javascript/controllers/appointment_select_controller.js
```diff
@@ -18,12 +18,11 @@ export default class extends Controller {
   }
 
   updateItemCounts() {
-    if (!this.element.closest('#reading-accordion')) return;
+    if (!this.element.closest('#reading')) return;
```
dnoneill, line 21: The reading id also covers the modal option. So now the modal updates with counts

Image

app/components/aeon/appointment_form_item_component.html.erb
```diff
@@ -21,8 +21,8 @@
     <div class="appt-select d-flex flex-wrap align-items-center column-gap-2 row-gap-1">
       <div class="fs-6 mb-0 w-100 form-label fw-semibold redesign-required-label possibly-visually-hidden-appointments-label">Appointment</div>
       <div class="dropdown custom-select flex-grow-1 flex-md-grow-0" data-controller="appointment-select" data-action="change->appointment-select#updateSelected input->appointment-select#updateSelected appointment-select:change@document->appointment-select#updateItemCounts">
-        <%= f.hidden_field :appointment_id, class: 'visually-hidden', data: { 'appointment-select-target': :input, 'required-for-submit': true } %>
         <%= button_tag 'Select appointment', class: 'py-1 btn border dropdown-toggle text-start d-flex align-items-center justify-content-between', type: 'button', data: { bs_toggle: 'dropdown', appointment_select_target: 'button' }, aria: { expanded: false }, disabled: selectable_appointments.blank? %>
+        <%= f.text_field :appointment_id, class: 'visually-hidden-focusable required-placeholder', data: { 'appointment-select-target': 'input' }, required: true %>
```
taylor-steve, line 25: I haven't gone through all of the PR yet, but should this be focusable?

Image

dnoneill: According to what we are already doing in this application yes.
Screenshot 2026-07-27 at 3 07 09 PM

taylor-steve: I think persisting behavior is different than an intentional choice 😄

dnoneill: I can double check with Chris but I believe it was intentional choice. I think the alerts aren't accessible if the inputs aren't focusable.

taylor-steve (approved):

PR #4214, opened 2026-07-27
Use the delete modal for deleting activities.

cbeer:
Fixes #4212


app/components/aeon/appointment_request_group_item_component.html.erb
```diff
@@ -4,7 +4,9 @@
 
       <turbo-frame>
         <% if request.activity? %>
-          <%= render Aeon::RemoveFromActivityButtonComponent.new(request:) %>
+          <div class="px-2 py-1">
```
cbeer, line 7: There's something really weird about how activity requests get row spacing that I'm not going to try to tackle in this PR..

taylor-steve (approved):

PR #4219, opened 2026-07-27
update styling on reading room appointments list

dnoneill:
closes #4209 

Screenshot 2026-07-27 at 1 53 07 PM


app/components/aeon/appointment_request_group_item_component.html.erb
```diff
@@ -10,8 +10,4 @@
         <% end %>
       </turbo-frame>
     </div>
-
-    <div class="rgi-detail pb-1">
```
dnoneill, line 16: I couldn't figure out why this was here. It seems to be un-used and not in spec.

taylor-steve (approved):

PR #4230, opened 2026-07-28
Add the FOLIO location for EADs with barcodes

cbeer:
Kinda hacky, but presumably nicer for staff processing?

### DIFF
--- a/app/jobs/submit_aeon_patron_request_job.rb
+++ b/app/jobs/submit_aeon_patron_request_job.rb
@@ -91,6 +91,7 @@ class SubmitAeonPatronRequestJob < ApplicationJob
       item_number: item.barcode,
       item_info4: patron_request.ead_doc.conditions_governing_access,
       item_volume: item.title,
+      location: origin_location_code_from_folio(item),
       web_request_form: 'multiple'
     )
   end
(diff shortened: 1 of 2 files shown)


app/jobs/submit_aeon_patron_request_job.rb
```diff
@@ -91,6 +91,7 @@ def as_aeon_create_ead_request_data(patron_request, item, activity_id)
       item_number: item.barcode,
       item_info4: patron_request.ead_doc.conditions_governing_access,
       item_volume: item.title,
+      location: origin_location_code_from_folio(item),
```
dnoneill, line 94: shouldn't this go in `common_aeon_data_from_patron_request`?

cbeer: I don't think so? If it's a FOLIO item, we have a less convoluted path to get the data.

dnoneill: I guess I am wondering why we don't have the function do that check first? But I guess since we know what the is Folio and not we might as well have them as separate functions.

dnoneill (approved):

PR #4236, opened 2026-07-28
Update checkout statuses

cbeer:
Fixes #4233

### DIFF
--- a/app/components/folio/checkout_component.rb
+++ b/app/components/folio/checkout_component.rb
@@ -18,18 +18,25 @@ module Folio
     end
 
     def non_renewable_reason
-      return 'Assumed lost' if lost?
-      return 'Renew' if recalled? || renewal_blocked_by_hold?
-      return 'Claim review is in process' if claimed_returned?
-      return 'Renew in person' if reserve_item?
-      return 'Too soon to renew' if too_soon_to_renew?
+      return 'Too soon to renew' if too_soon_to_renew? && header_message.blank?
 
       'Renew'
     end
 
-    def header_message
-      return 'Please return as soon as possible. Item cannot be renewed.' if recalled? || renewal_blocked_by_hold?
-      return accruing_message if checkout.overdue?
+    def header_message # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+      @header_message ||= 'This item was requested by another user. Please return as soon as possible.' if recalled? || renewal_blocked_by_hold? # rubocop:disable Layout/LineLength
+      @header_message ||= accruing_message if checkout.accruing?
+      @header_message ||= 'You reported this item as returned. Library staff are reviewing its status.' if checkout.claimed_returned?
+      @header_message ||= 'This overdue item is assumed lost.' if checkout.lost?
+      @header_message ||= 'This is a reserve item. Renew in person.' if reserve_item?
+      @header_message ||= 'Renewals are not allowed.' if checkout.item_category_non_renewable?
+      @header_message ||= 'Online renewals are not allowed. Renew in person.' unless checkout.unseen_renewals_allowed?
+      @header_message ||= 'No online renewals remain. Renew in person.' if unseen_renewals_remaining.zero?
+
+      @header_message
+    end
+
+    def specific_return_location_message
       return unless reserve_item? && location&.library
 
       "NOTE: This item must be returned to the #{location.library.primary_service_points.first.name}"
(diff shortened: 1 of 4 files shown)


app/components/folio/checkout_component.rb
```diff
@@ -18,18 +18,26 @@ def initialize(checkout:, patron:, renewal_view: false)
...
       return 'Too soon to renew' if too_soon_to_renew?
 
       'Renew'
     end
 
-    def header_message
-      return 'Please return as soon as possible. Item cannot be renewed.' if recalled? || renewal_blocked_by_hold?
-      return accruing_message if checkout.overdue?
+    def header_message # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+      @header_message ||= 'This item was requested by another user. Please return as soon as possible.' if recalled? || renewal_blocked_by_hold? # rubocop:disable Layout/LineLength
```
dnoneill, line 27: The ticket says this should be in a pill. Not the header message.

dnoneill (dismissed):

app/components/folio/checkout_component.rb
cbeer: I'm not sure what you're suggesting. 

Here's the Figma:
Screenshot 2026-07-28 at 14 00 45

and as implemented:
Screenshot 2026-07-28 at 14 01 33

dnoneill (approved):
The spacing/font-weight looks off around the pills but I am fine leaving that for a different PR/ticket if we want since the original issue is around messaging.

Image
Image

app/components/folio/checkout_component.rb
dnoneill: This is no longer needed I believe. It can be static

app/components/folio/checkout_component.rb
```diff
@@ -18,18 +18,26 @@ def initialize(checkout:, patron:, renewal_view: false)
...
+      @header_message ||= accruing_message if checkout.accruing?
```
dnoneill, line 28:
```suggestion
      @header_message ||= accruing_message
```

https://github.com/sul-dlss/sul-requests/blob/main/app/components/folio/checkout_component.rb#L38-L42 already has a return nil unless checkout.accruing?

cbeer: i thought this was clearer, and it was nice to have parallel construction here 🤷

app/components/folio/checkout_component.rb
```diff
@@ -18,18 +18,26 @@ def initialize(checkout:, patron:, renewal_view: false)
...
+      @header_message ||= 'You reported this item as returned. Library staff are reviewing its status.' if checkout.claimed_returned?
+      @header_message ||= 'This overdue item is assumed lost.' if checkout.lost?
```
dnoneill, line 30: I believe all of these are already delegated. We might not want to have two different ways of doing this.
https://github.com/sul-dlss/sul-requests/blob/main/app/components/folio/checkout_component.rb#L10-L11

app/components/folio/checkout_component.rb
```diff
@@ -18,18 +18,26 @@ def initialize(checkout:, patron:, renewal_view: false)
     end
 
     def non_renewable_reason
-      return 'Assumed lost' if lost?
-      return 'Renew' if recalled? || renewal_blocked_by_hold?
-      return 'Claim review is in process' if claimed_returned?
-      return 'Renew in person' if reserve_item?
+      return 'Renew' if header_message.present?
```
dnoneill:
```suggestion
      return 'Too soon to renew' if too_soon_to_renew? && header_message.blank?
```

app/components/folio/checkout_component.rb
cbeer: I don't find the delegation very clear, and we were inconsistent about it already. I'd rather get rid of delegations we don't need, but I don't think we should tie that up with this implementation.

app/components/folio/checkout_component.rb
```diff
@@ -18,18 +18,26 @@ def initialize(checkout:, patron:, renewal_view: false)
     end
 
     def non_renewable_reason
-      return 'Assumed lost' if lost?
-      return 'Renew' if recalled? || renewal_blocked_by_hold?
-      return 'Claim review is in process' if claimed_returned?
-      return 'Renew in person' if reserve_item?
+      return 'Too soon to renew' if too_soon_to_renew? && header_message.blank?
       return 'Too soon to renew' if too_soon_to_renew?
```
dnoneill: This line needs to get deleted as well
```suggestion
      
```

app/components/folio/checkout_component.rb
dnoneill: Okay. I agree. I am fine getting rid of the delegations I just don't like the inconsistent mix.

PR #4239, opened 2026-07-28
put spacing on list item and not button

dnoneill:
closes #4235 
Screenshot 2026-07-28 at 4 34 27 PM


app/components/aeon/appointment_request_group_item_component.html.erb
cbeer:
```suggestion
    <div class="rgi-id d-flex align-items-center flex-row justify-content-between border-top py-1">
<%= render Aeon::RequestItemIdentifierComponent.new(request:) %>

<turbo-frame>
<% if request.activity? %>
<div class="px-2">
```

Otherwise, I think we're doubling up on padding here.

PR #4223, opened 2026-07-27
group folio requests by catkey

dnoneill:
closes #4210 

image

PR #4255, opened 2026-07-29
fix flashes for removing folio requests

dnoneill:
see https://github.com/sul-dlss/sul-requests/pull/4223#discussion_r3676222618

### DIFF
--- a/app/controllers/folio_requests_controller.rb
+++ b/app/controllers/folio_requests_controller.rb
@@ -8,6 +8,7 @@ class FolioRequestsController < ApplicationController
 
   before_action :load_requests
   before_action :load_request, except: [:index]
+  before_action :empty_flashes, only: [:destroy, :update]
   before_action :set_variant, only: [:edit]
 
   rescue_from RequestException, with: :deny_access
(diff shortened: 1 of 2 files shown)


app/controllers/folio_requests_controller.rb
```diff
@@ -8,6 +8,7 @@ class FolioRequestsController < ApplicationController
 
   before_action :load_requests
   before_action :load_request, except: [:index]
+  before_action :empty_flashes, only: [:destroy, :update]
```
cbeer, line 11:
```suggestion
```

I don't understand what this is about in general, but especially for `update`. I don't think we need it.

app/controllers/folio_requests_controller.rb
```diff
@@ -96,10 +95,10 @@ def handle_change_pickup_expiration(ils_client: FolioClient.new)
   def response_flash_message(response:, translation_key:)
     case response.status
     when 204
-      flash[:success].push(t("mylibrary.request.#{translation_key}.success_html", title: params['title']))
+      flash[:success].push(t("mylibrary.request.#{translation_key}.success_html", title: @request&.title))
```
cbeer, line 98:
```suggestion
      flash[:success] = t("mylibrary.request.#{translation_key}.success_html", title: @request&.title)
```

Am I missing a loop somewhere? I don't see why `push` is needed.

dnoneill: Push is needed for `update` https://github.com/sul-dlss/sul-requests/blob/main/spec/features/requests_spec.rb#L55 and I didn't see the point of writing two functions that do the same thing just one doesn't write to a list.

app/controllers/folio_requests_controller.rb
dnoneill: We do. If we don't set `flash[:success] = []` and `flash[:error] = []` and then try to push we get errors: ``` NoMethodError:
       undefined method 'push' for nil```

PR #4256, opened 2026-07-29
Update the full date format

cbeer:
Per Darcy, all the date added/modified should use the 3-letter abbreviation.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -97,7 +97,7 @@ en:
   time:
     formats:
       short: "%Y-%m-%d %l.%M%P"
-      full: "%B %-d, %Y %l:%M %P"
+      full: "%b %-d, %Y %l:%M %P"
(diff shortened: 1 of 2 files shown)


app/components/aeon/request_item_detail_component.rb
```diff
@@ -7,9 +7,8 @@ class RequestItemDetailComponent < ViewComponent::Base
 
     delegate :ead?, :requested_pages, :volume, :format, to: :request
 
-    def initialize(request:, variant: :full)
+    def initialize(request:)
       @request = request
-      @variant = variant
```
dnoneill, line 12: I think this might be the wrong spot for date? But if it isn't we should be updating 
https://github.com/sul-dlss/sul-requests/blob/30e620b78a2f048a7b695228aaf19e2560b9b86c/app/components/aeon/request_item_detail_component.html.erb#L1-L8

taylor-steve (approved):

PR #4257, opened 2026-07-29
Update the ILL needed after date.

cbeer:
... for consistency with the FOLIO requests.

### DIFF
--- a/app/views/ill_requests/new.html.erb
+++ b/app/views/ill_requests/new.html.erb
@@ -194,7 +194,7 @@
             </div>
 
             <div class="mb-4">
-              <%= f.label :not_wanted_after, 'Not wanted after date', class: 'form-label fw-semibold' %>
+              <%= f.label :not_wanted_after, 'Not needed after', class: 'form-label fw-semibold' %>
               <%= f.date_field :not_wanted_after, class: 'form-control' %>
             </div>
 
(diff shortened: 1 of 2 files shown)


dnoneill (approved):

app/views/ill_requests/new.html.erb
```diff
@@ -194,8 +194,8 @@
             </div>
 
             <div class="mb-4">
-              <%= f.label :not_wanted_after, 'Not wanted after date', class: 'form-label fw-semibold' %>
-              <%= f.date_field :not_wanted_after, class: 'form-control' %>
+              <%= f.label :not_wanted_after, 'Not needed after', class: 'form-label fw-semibold' %>
+              <%= f.date_field :not_wanted_after, value: 2.months.from_now.strftime('%Y-%m-%d'), class: 'form-control' %>
```
dnoneill, line 198: We used a default of 1.year from now. Is this supposed to be different for a new iLL request?
https://github.com/sul-dlss/sul-requests/blob/30e620b78a2f048a7b695228aaf19e2560b9b86c/app/views/patron_requests/_not_needed_after_date.html.erb#L10

cbeer: Yes, per Hilary.

PR #4266, opened 2026-07-29
Remove unused group form param

cbeer:


app/components/folio/checkout_component.html.erb
```diff
@@ -17,7 +17,6 @@
       <%= form_tag renew_checkout_path(checkout.item_id), method: :post do %>
         <%= hidden_field_tag :item_id, checkout.item_id %>
         <%= hidden_field_tag :title, checkout.title %>
-        <%= hidden_field_tag :group, params[:group] if params[:group] %>
```
dnoneill, line 20: I don't think the param exists but do we need this? Is this for patron_proxy group?

cbeer: No? We're not doing anything with it.

dnoneill (approved):

PR #4272, opened 2026-07-29
Only use the first word as the fine status.

cbeer:
FOLIO has some pretty wordy fine labels:

```
"Outstanding",
            "Paid partially",
            "Paid fully",
            "Waived partially",
            "Waived fully",
            "Transferred partially",
            "Transferred fully",
            "Refunded partially",
            "Refunded fully",
            "Credited fully",
            "Credited partially",
            "Cancelled item returned",
            "Cancelled item renewed",
            "Cancelled item declared lost",
            "Cancelled as error",
            "Suspended claim returned"
```

The first word is probably enough detail, per Darcy.

### DIFF
--- a/app/components/folio/fine_status_pill_component.rb
+++ b/app/components/folio/fine_status_pill_component.rb
@@ -3,11 +3,17 @@
 module Folio
   # Render a pill with correct value/color combination based on fee status
   class FineStatusPillComponent < ViewComponent::Base
-    attr_reader :status
+    attr_reader :fine
 
-    def initialize(status:)
-      @status = status
+    delegate :status, to: :fine
+
+    def initialize(fine:)
+      @fine = fine
       super()
+    end
+
+    def render?
+      fine.closed?
     end
 
     def status_classes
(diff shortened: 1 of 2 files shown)


app/components/folio/fine_status_pill_component.rb
```diff
@@ -3,28 +3,32 @@
...
         %w[fine-status bg-green text-green]
       else
         %w[fine-status bg-stanford-20-black text-black]
       end
     end
 
-    def fine_status
-      status.gsub('fully', '').strip.upcase
+    def status
+      fine.status.split(' ', 2).first
```
dnoneill, line 27: Shouldn't this be?
```suggestion
      fine.status.split(' ', 2).first.upcase
```

cbeer: We're not up-casing any of our other pills.

dnoneill: The figma has them upcased
Screenshot 2026-07-29 at 5 43 40 PM

cbeer: Sure, I'll make that change for now.

dnoneill (approved):

PR #4282, opened 2026-07-30
Extract System::DeleteWithConfirmationComponent

cbeer:
If we're looking to consolidate the delete button stuff, here's the kind of extraction I'd expect (shared layout, but let the rendering code populate the differences).

### DIFF
--- a/app/components/system/delete_with_confirmation_component.html.erb
+++ b/app/components/system/delete_with_confirmation_component.html.erb
@@ -1,0 +1,26 @@
+<button class="btn btn-link su-underline p-0" data-bs-toggle="modal" data-bs-target="#<%= id %>">
+  <i class="bi bi-trash align-middle" aria-hidden></i><span class="visually-hidden"><%= button_label %></span>
+</button>
+
+<!-- Delete modal -->
+<%= render ModalComponent.new(id: id) do |component| %>
+  <% component.with_title.with_content('Delete request?') %>
+
+  <% component.with_body do %>
+    <div class="d-flex flex-column gap-3">
+      <%= body %>
+
+      <%= render AlertComponent.new(type: :info, classes: ['border-0', 'shadow-none', 'mb-0'], with_icon: false) do %>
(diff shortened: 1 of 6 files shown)


app/components/system/delete_with_confirmation_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module System
```
dnoneill, line 3: This is only for requests. Since I just created RequestRgiFooter I am wondering if we want a requests component module, move the footer and this into there instead of the slightly less clear system.

app/components/system/delete_with_confirmation_component.html.erb
```diff
@@ -0,0 +1,26 @@
+<button class="btn btn-link su-underline p-0" data-bs-toggle="modal" data-bs-target="#<%= id %>">
+  <i class="bi bi-trash align-middle" aria-hidden></i><span class="visually-hidden"><%= button_label %></span>
+</button>
+
+<!-- Delete modal -->
+<%= render ModalComponent.new(id: id) do |component| %>
+  <% component.with_title.with_content('Delete request?') %>
+
+  <% component.with_body do %>
+    <div class="d-flex flex-column gap-3">
```
dnoneill, line 10: I am wondering if we should make request_type a variable and put in here. The html styling is the same across delete confirmations and if style tweaks happen we wouldn't have to do it in three places

cbeer: Maybe? It didn't feel like a hard requirement for the initial extraction (and raised some questions for me about the filter value and I didn't want to hold this up to figure that out 🤷 )

app/components/system/delete_with_confirmation_component.rb
cbeer: Maybe, but I don't have strong feelings. I think the only request-specific thing is the hard-coded modal title but otherwise it's pretty generic.

dnoneill (approved):

PR #4284, opened 2026-07-30
move js sort and filter dropdowns into component

dnoneill:
I want to switch the dropdowns in activity to js since activities are so slow. I think this will make things much easier.

### DIFF
--- a/app/components/system/sortable_dropdown_component.html.erb
+++ b/app/components/system/sortable_dropdown_component.html.erb
@@ -1,0 +1,10 @@
+<div class="dropdown" data-sortable-target="sortMenu">
+  <button class="btn btn-outline-primary btn-sm dropdown-toggle" id="dropdownMenuLink" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    Sort by <%= active_label %>
+  </button>
+  <ul class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+    <% list_items.each do |list_item| %>
+      <%= list_item %>
+    <% end %>
+  </ul>
+</div>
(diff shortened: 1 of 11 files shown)


app/components/system/sortable_dropdown_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="dropdown" data-sortable-target="sortMenu">
+  <button class="btn btn-outline-primary btn-sm dropdown-toggle" id="dropdownMenuLink" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    Sort by <%= button_label %>
```
cbeer: Won't this  render `Sort be Sort by` by default (although I think the javascript rewrites it?

dnoneill: yes that would be the default if the user doesn't pass anything to it. Do you have a preference for something else?

app/components/system/sortable_dropdown_component.rb
```diff
@@ -0,0 +1,30 @@
...
+                   }) do
+          label
+        end
+      end
+    }
+
+    def classes(label)
+      return 'dropdown-item' unless button_label == label
+
+      'dropdown-item active'
```
cbeer:
```suggestion
      return 'dropdown-item active' if button_label == label

      'dropdown-item active'
```

app/components/system/sortable_dropdown_component.html.erb
dnoneill: Ah. Yeah I changed that to an empty value. I also changed the variable to active_label to be more descriptive.

PR #4285, opened 2026-07-30
Activity sort

dnoneill:
see https://github.com/sul-dlss/sul-requests/pull/4284 for first commit on this PR


app/views/aeon_activities/_activities_list.html.erb
```diff
@@ -0,0 +1,20 @@
+<div data-controller="sortable">
+  <div class="d-flex flex-wrap justify-content-end gap-2 my-2">
+    <%= render System::SortableFilterDropdownComponent.new(button_label: 'All') do |component| %>
```
cbeer: As you've done it, won't the two `SortableFilterDropdownComponent` instances render on the same page with the same DOM id?

dnoneill: One is SortableFilterDropdownComponent and one is SortableDropdownComponent. So one is for filters and one is for sorting. I can rename them so they are less confusing but I thought it would nice is they were next to each other.

cbeer: Sorry, I mean the sort dropdown in the current tab is going to have the same id as the sort dropdown in the past tab.

dnoneill: Ah. Gotcha. I don't see an id at all on filters. Isn't the controller the one that would control the elements and shouldn't it only work on the dropdowns it encompasses?

dnoneill: Ah but it isn't accessible to have two of the same ids on the page. Fixing now.

PR #4304, opened 2026-07-31
add record header to edit modal

dnoneill:
closes #4263 

Screenshot 2026-07-31 at 3 01 44 PM

Screenshot 2026-07-31 at 2 37 52 PM

### DIFF
--- a/app/components/record_header_component.html.erb
+++ b/app/components/record_header_component.html.erb
@@ -1,25 +1,28 @@
 <%= tag.div(class: "card border-0 #{classes}") do %>
-  <div class="card-body">
-    <%= tag.public_send title_tag, record.title || 'Not available', class: @title_classes %>
-    <div class="d-flex column-gap-md-3 flex-md-row flex-column flex-wrap">
-      <span>
-        <%= render FormatIndicatorComponent.new(record: record) %><% if record.date.present? %> (<%= record.date %>)<% end %>
-      </span>
-      <% if call_number.present? %>
-        <span>Call number: <%= call_number %></span>
-      <% end %>
-      <span>
-        <%= render System::RequestLinkComponent.new(url: record.item_url) %>
-      </span>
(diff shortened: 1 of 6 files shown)


app/components/record_header_component.html.erb
```diff
@@ -1,25 +1,32 @@
...
+      <%= tag.public_send title_tag, record.title || 'Not available', class: @title_classes %>
+      <div class="d-flex column-gap-md-3 flex-md-row flex-column flex-wrap">
+        <span>
+          <%= render FormatIndicatorComponent.new(record: record) %><% if record.date.present? %> (<%= record.date %>)<% end %>
+        </span>
+        <% if call_number.present? %>
+          <span>Call number: <%= call_number %></span>
+        <% end %>
+        <span>
+          <%= render Aeon::RequestLinkComponent.new(url: record.item_url) %>
```
cbeer: Is Aeon::RequestLinkComponent` still the right name if we're using it for folio requests too?

dnoneill: No but I don't think it has been true for a while since we started using this component for patron_request which can be aeon or folio. I didn't think this PR was the right place to rename but I can add it to this PR:
https://github.com/sul-dlss/sul-requests/blob/main/app/views/patron_requests/new.html%2Baeonredesign.erb#L15
https://github.com/sul-dlss/sul-requests/blob/main/app/views/patron_requests/_confirmation_screen.html.erb#L5
https://github.com/sul-dlss/sul-requests/blob/main/app/views/patron_requests/show.html%2Baeon.erb#L11
https://github.com/sul-dlss/sul-requests/blob/main/app/views/patron_requests/new.html.erb#L9

dnoneill: see https://github.com/sul-dlss/sul-requests/pull/4312

PR #4340, opened 2026-08-05
don't memoize appointments

dnoneill:
closes #4339

### DIFF
--- a/app/models/aeon/user.rb
+++ b/app/models/aeon/user.rb
@@ -66,16 +66,14 @@ module Aeon
     end
 
     def appointments
-      @appointments ||= begin
-        appointments = self.class.aeon_client.appointments_for(username:)
+      appointments = self.class.aeon_client.appointments_for(username:)
 
-        # augment appointments with their requests
-        appointments.each do |appointment|
-          appointment.user = self
-        end
-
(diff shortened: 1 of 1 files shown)


app/models/aeon/user.rb
```diff
@@ -66,16 +66,14 @@ def activities
     end
 
     def appointments
-      @appointments ||= begin
-        appointments = self.class.aeon_client.appointments_for(username:)
+      appointments = self.class.aeon_client.appointments_for(username:)
```
cbeer, line 69: Not memoizing appointments seems really bad.  From the screenshot, I'm not sure what problem you're solving or why memoization has anything to do with it. Maybe a test demonstrating the problem would help?

dnoneill: There is a video attached. What it shows if you create a new appointment the memoization keeps the appointments from updating and it allows you to create a new appointment on a day you already booked in Field. So if I booked an appointment for August 10, then save it up. It updates and then go to create another appointment, I can still book August 10th again.

cbeer: I'm not seeing what you're seeing, and I don't understand how memoization has anything to do with it. Again, maybe a test would help demonstrate the problem.

